PR #50, opened 2026-06-05
chore: remove stray internal tracking ID from code comments

CiroGamboa:
## What

Two DeepSeek V3.2-Exp comments carried an internal tracking ID:

- `tolokaforge/core/data/model_presets.yaml`
- `tests/integration/llm/registry.py`

Both comments are reworded to drop the ID; the descriptive text stays.

## Why

Internal tracking IDs don't belong in public source — they're meaningless to OSS readers and go stale. This is a docs/comment-only change; no behavior change.

### DIFF
--- a/tests/integration/llm/registry.py
+++ b/tests/integration/llm/registry.py
@@ -868,7 +868,7 @@ _ALL: list[MC] = [
         ),
     ),
     # -----------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


CiroGamboa:
Folded into #48 (same internal-reference cleanup); closing this separate PR.

PR #61, opened 2026-06-16
refactor(adapters): make the runner adapter-agnostic (plugin-first)

### LINKED ISSUE #60: Make the adapter system plugin-first (stop branching on adapter identity)
## Summary

The engine still branches on a specific adapter's **identity** in several places, so
adding an adapter with any non-trivial behavior requires editing the engine rather
than just shipping a plugin. Make the adapter system fully plugin-first: an adapter
should be addable as an entry-point package that implements the contract and declares
its needs **as data/capabilities**, with **zero engine edits**.

## Where the engine hard-codes adapter knowledge today

- `runner/models.py`: `TaskDescription.adapter_type` is a **closed enum**, so a
  registered plugin can't even serialize/round-trip a `TaskDescription`.
- `runner/service.py`: `if adapter_type == TERMINAL_BENCH` gates **trial setup**
  (compose start), **grading** (test-execution), and **teardown** (compose stop);
  plus a hardcoded `available_adapters` list.
- `core/orchestrator.py`: `isinstance(self.adapter, NativeAdapter)` gates env-state sync.

## Proposed approach (interface + default + discovery)

Replace each adapter-identity branch with a capability/declarative seam whose default
preserves current behaviour:

- **Open `AdapterType`** → free string from the registry (built-in names kept as
  constants); report the registry's adapters dynamically.
- **Tool lifecycle as a capability** → base tool wrapper gains `has_lifecycle` +
  no-op `start(ctx)/stop()`; the compose-backed tool opts in. The runner drives
  lifecycle off the capability, not the adapter type.
- **Declarative grading method** → `GradingConfig.grading_method` (e.g.
  `test_execution`); the runner dispatches on config.
- **`BaseAdapter.provides_env_state()` capability** replaces the `isinstance` check.

## Acceptance

- No adapter-identity branches remain in engine code.
- A non-built-in adapter type round-trips through `TaskDescription` and resolves via
  the registry with no engine edit.
- No behaviour change: the golden grading regression tests pass unchanged; the
  built-in `native` and `terminal_bench` adapters work end-to-end (incl. the Docker
  compose-lifecycle + test-execution path).

CiroGamboa:
Makes the **runner** adapter-agnostic: it manages tool lifecycle and selects grading
from **declarative data/capabilities** in the `TaskDescription`, never from the
adapter name. Combined with an open `adapter_type`, a new adapter can ship as an
entry-point package and round-trip a `TaskDescription` with **zero engine edits**.
No behaviour change.

Closes #60

> **Re-scoped after review.** An earlier revision also de-identity'd the
> orchestrator's env-state branch via a `provides_env_state()` adapter hook. That was
> the wrong layer: environment-state ownership belongs to the runner (the sandbox)
> and the per-trial runtime, and the orchestrator's env-state syncing is boundary-less
> logic that a later orchestrator refactor will rework. That change has been
> **dropped and deferred**; this PR is scoped to the runner.

## Changes

- **Open `adapter_type`** — `TaskDescription.adapter_type` is now an open `str` from
  the registry (built-in names kept as constants). Consistent with the existing
  `core` model + run-config `type`. A host-side `ensure_registered_adapter()` guard
  validates the name where the registry is authoritative; the **runner stays
  permissive** (no model validator) so it never falsely rejects an adapter it doesn't
  have installed.
- **Tool lifecycle is a capability** — base `ToolWrapper` gains `has_lifecycle=False`
  + no-op `start(ctx)/stop()`; `DockerComposeExecToolWrapper` opts in and owns its
  per-trial naming/artifacts. `RegisterTrial`/`ResetTrial` drive lifecycle off the
  capability — removes the `== TERMINAL_BENCH` setup/teardown branches.
- **Declarative grading method** — `GradingConfig.grading_method` (e.g.
  `test_execution`); `GradeTrial` dispatches on config
  (`_grade_terminal_bench` → `_grade_via_test_execution`) — removes the grading
  identity branch.
- `terminal_bench` adapter declares `grading_method="test_execution"`; snapshot
  updated for the new field. (See "Why terminal-bench changes too" below.)

## Why terminal-bench changes too

The old contract was implicit: the runner inferred "use test-execution grading"
from the adapter's *name* (`adapter_type == AdapterType.TERMINAL_BENCH`). The whole
point of this PR is to stop reading the name and start reading explicit data. So:

- The runner stopped saying *"I'll guess what you want based on who you are."*
- It now says *"tell me explicitly via `grading_method` on the grading config."*

If the runner stops inferring, someone has to declare. The only candidate is the
adapter that produces the `TaskDescription` — and for test-execution grading
today, that's terminal-bench. **One adapter, one line** (plus its snapshot picking
up the new field). The two changes are coupled by the contract: the callee changed
the field it reads, the (sole) caller must write that field. They had to ship
together — splitting them would leave a window where terminal-bench falls through
to default grading and produces wrong verdicts.

## Design note: why `adapter_type` is an open string

The set of adapters is **open and discovered at runtime** (entry-point plugins), so
a closed enum can't represent it — it would block third-party adapters from even
round-tripping a `TaskDescription` and force an engine edit per adapter. The
standard pattern for that is a **string key validated against a registry** (the
same shape as setuptools entry points / pytest plugins).

More importantly: after this PR, `adapter_type` drives **no behaviour** — the
runner selects behaviour from explicit, well-typed capabilities (`grading_method:
Literal[…]`, `has_lifecycle: bool`, the tools themselves), not from the adapter
name. So `adapter_type` is now **provenance metadata** (which adapter produced
this task), and a plain string is the appropriate representation for a provenance
tag — a richer/closed type would signal it's load-bearing when the design goal is
the opposite. Validation lives at the host boundary
(`ensure_registered_adapter`, where the registry is authoritative) to catch typos
early, while the runner stays permissive.

## Contract for new adapter authors

What an adapter needs to know to use the new contract correctly:

- **Default works.** A new adapter that just produces a `TaskDescription` and
  doesn't set `grading_method` gets standard weighted grading
  (state checks / transcript rules / LLM judge per `combine_method` + `weights`).
  This is what most adapters want; nothing extra to do.
- **Want test-execution grading?** Set
  `grading=GradingConfig(grading_method="test_execution", ...)` **and** include an
  exec-capable lifecycle tool (today: `DockerComposeExecToolWrapper`) in
  `TaskDescription.agent_tools`. The runner runs `test.sh` inside the env container
  and scores by reading the reward float from `/logs/verifier/reward.txt`.
- **Tool lifecycle.** If a tool owns per-trial resources (e.g. a compose stack),
  set `has_lifecycle = True` on the wrapper and override `start(ctx)` / `stop()`.
  The runner drives lifecycle generically — no engine code knows your tool by
  name.

## Error handling for "new adapter does X wrong"

| Case | Caught by | Message |
|---|---|---|
| Unknown / typo'd `adapter_type` | `ensure_registered_adapter` (host) | `"Unknown adapter type: 'foo'. Available adapters: […]"` |
| Invalid `grading_method` string (typo) | Pydantic `Literal[…]` at model validation | `ValidationError` with allowed values |
| `grading_method="test_execution"` but no exec tool | Runner returns `GradeTrialResponse(success=False, …)` | `"test-execution grading was requested (grading_method='test_execution') but no exec-capable env tool was found in this trial. Include an exec-capable lifecycle tool (e.g. DockerComposeExecToolWrapper) in TaskDescription.agent_tools so the runner can execute the test suite inside the trial environment."` |
| Lifecycle tool `start()` raises | Runner returns `RegisterTrialResponse(success=False, …)` | `"Tool lifecycle start failed: <e>"` |
| Lifecycle tool `stop()` raises | Runner logs `WARNING` and continues teardown | `"ResetTrial: Failed to stop tool lifecycle: <e>"` |
| Adapter omits `grading_method` entirely | (intended) Falls through to default weighted grading | n/a — silent because `None` is the documented "I want default grading" signal |

The one case there's no automatic detection for is "I *meant* to set
`grading_method` but forgot." That can't be inferred without re-introducing magic
("you have a compose-exec tool, so you must want test-execution") — exactly the
identity coupling this PR removes. The defence there is documentation (the field
docstring) and a strict `Literal` type so typos can't sneak past.

## Known limitations of the lifecycle contract (tracked follow-up: #63)

The tool lifecycle contract introduced here (`has_lifecycle` + `start(ctx)` /
`stop()`) is intentionally **minimum viable** — sufficient for the one lifecycle
tool that ships today (a Docker Compose exec wrapper for `terminal_bench`) but
shallow on purpose. Known gaps for future custom lifecycle tools:

- `ToolLifecycleContext` is sparse (only `trial_id` + `artifacts_dir`); tools
  needing env-service endpoints, resource limits, secrets, etc. don't yet
  receive them.
- `start(ctx)` is synchronous; no timeout / retry / async companion.
- Only `start` / `stop` hooks exist (no `suspend` / `snapshot` /
  `reset-without-teardown`).
- No ordering / dependency / readiness signalling between multiple lifecycle
  tools in the same trial.

The shape is forward-compatible (`ToolLifecycleContext` is a dataclass; dispatch
is capability-driven, not adapter-identity-driven), so the contract will grow
additively when a real second user surfaces a need. Documented in
`docs/ADAPTER_INTERFACE.md` and tracked in #63 — not addressed in this PR.

## Explicitly deferred (not this PR)

- **Orchestrator env-state ownership** (`isinstance(self.adapter, NativeAdapter)`):
  left untouched — removing it properly means moving env-state ownership to the
  runner, which belongs to a later orchestrator refactor.
- **CLI `convert` `if name == "tau"`** param mapping and the tool `InvocationStyle`
  dispatch (a tool-mechanism concern, not adapter identity).

## Verification

- Full offline suite **1817 passed**, including the golden grading regression gate
  (verdicts unchanged).
- Scoped grep gate: **no `adapter_type == AdapterType` branches in `tolokaforge/runner/`**
  (enforced by a regression test that scans the runner's behavioural modules).
- Unit tests pin the seams: open-type round-trip, declarative `grading_method`
  dispatch, lifecycle capability (start/stop on lifecycle tools, no-op otherwise),
  host-side adapter_type guard (accept / reject / fresh-plugin), and the
  adapter-author error paths (typo'd grading method, missing exec tool).
- Docker end-to-end: a `terminal_bench` task runs the full pipeline (resolve →
  compose-up via the generic lifecycle → agent loop → `test_execution` grading →
  teardown). Behaviour matches `main` (A/B verified, identical verdict).

### DIFF
--- a/docs/ADAPTER_INTERFACE.md
+++ b/docs/ADAPTER_INTERFACE.md
@@ -50,6 +50,73 @@ Each adapter must subclass `BaseAdapter` and implement:
 7. Reset: cleanly reset state between trials.
 8. *(Optional)* Conversion: emit native format bundle via `convert_to_native()`.
 
+## Runtime Capabilities (declarative, opt-in)
+
+The runner selects per-trial behaviour from **data on the `TaskDescription`**,
+never from the adapter's identity. Two capabilities matter to adapter authors:
+
+### `GradingConfig.grading_method`
+
+A declarative selector on the grading config that tells the runner *how* to grade:
+
+- **`None`** *(default)* — standard grading: combines state checks / transcript
+  rules / LLM judge per `weights` + `pass_threshold`. Most adapters want this and
+  need do nothing.
+- **`"test_execution"`** — the runner runs a reference test suite inside the
+  trial's env container via an exec-capable lifecycle tool (today:
+  `DockerComposeExecToolWrapper`) and scores by reading a reward float written
+  by the suite. Requires such a tool in `TaskDescription.agent_tools`; otherwise
+  the runner returns a clear error at `GradeTrial` time. Used by the
+  `terminal_bench` adapter as a worked example.
+- `"hash"` / `"transcript"` / `"llm"` are reserved names for future
+  single-method dispatch and currently behave as part of the default path.
+
+Typos in this value are caught at validation by the `Literal[...]` field type;
+an "unknown grading method" cannot reach the runner silently.
+
+### `ToolWrapper.has_lifecycle`
+
+Tools may own per-trial resources (e.g. a compose stack started for the trial).
+The base `ToolWrapper` declares `has_lifecycle = False` and ships no-op
+`start(ctx)` / `stop()`. A tool that needs lifecycle management overrides:
+
+```python
+class MyTool(ToolWrapper):
+    has_lifecycle = True
+
+    def start(self, ctx: ToolLifecycleContext) -> None:
+        ...  # provision per-trial resources; ctx carries trial_id + artifacts_dir
+
+    def stop(self) -> None:
+        ...  # tear down what start() created
+```
+
+The runner calls `start` / `stop` generically on every tool whose
+`has_lifecycle` is set — it doesn't need to know your tool by name.
+
+#### Current limitations of the lifecycle contract
+
+The lifecycle contract is deliberately minimal today — sufficient for the one
+lifecycle tool that ships built-in, but shallow on purpose. Known gaps for
+custom lifecycle tools:
+
+- `ToolLifecycleContext` carries only `trial_id` and `artifacts_dir`. Tools
+  needing env-service endpoints, resource limits, secrets, or task metadata
+  must source them from elsewhere.
+- `start(ctx)` is synchronous, with no timeout or retry policy.
+- Only `start` / `stop` hooks exist (no `suspend` / `snapshot` /
+  `reset-without-teardown`).
+- No ordering or dependency declaration between multiple lifecycle tools in
+  the same trial — they're started in dict-iteration order.
+- No explicit readiness signal; `start()` returning is treated as "ready".
+
+The shape is forward-compatible: `ToolLifecycleContext` is a dataclass and the
+runner dispatches off the capability (not adapter identity), so the contract
+will grow additively when a real second user surfaces a need. Until then,
+custom lifecycle tools must either fit within the current shape or vendor the
+missing pieces internally. See the open tracking issue for the planned
+direction.
+
 ## Constructor Contract
 
 Adapters receive a `params: dict[str, Any]` in their constructor. Common params:
--- a/docs/TASK_DESCRIPTION_SCHEMA.md
+++ b/docs/TASK_DESCRIPTION_SCHEMA.md
@@ -38,7 +38,15 @@ from pydantic import BaseModel, Field
 # =============================================================================
 
 class AdapterType(str, Enum):
-    """Source adapter that produced this description."""
+    """Canonical constants for well-known built-in adapter names.
+
+    NOTE: ``TaskDescription.adapter_type`` is an open ``str`` from the adapter
+    registry, not a closed enum. Entry-point/third-party adapters round-trip
+    with their own names (e.g. an adapter installed via
+    ``pip install some-adapter-pkg``) without any engine edit. This enum
+    provides typed constants for the built-ins; the ``adapter_type`` field
+    accepts any name present in the registry at runtime.
+    """
     NATIVE = "native"
     TAU = "tau"
     TLK_MCP_CORE = "tlk_mcp_core"
@@ -275,7 +283,7 @@ class TaskDescription(BaseModel):
     name: str
     category: str                                 # Domain: e.g. "airline", "retail"
     description: str                              # Task description / user goal
-    adapter_type: AdapterType
+    adapter_type: str                             # Open string from the adapter registry; use AdapterType.* constants for the built-ins
     schema_version: str = "1.0.0"
     
     # --- System Prompt ---
--- a/external_adapters/tolokaforge-adapter-terminal-bench/src/tolokaforge_adapter_terminal_bench/adapter.py
+++ b/external_adapters/tolokaforge-adapter-terminal-bench/src/tolokaforge_adapter_terminal_bench/adapter.py
@@ -236,6 +236,9 @@ class TerminalBenchAdapter(BaseAdapter):
                 combine_method="weighted",
                 weights={"custom_checks": 1.0},
                 pass_threshold=0.5,
+                # Score by running the reference test suite in the env container.
+                # The runner dispatches on this method, not on the adapter name.
+                grading_method="test_execution",
             ),
             tool_artifacts=artifacts,
             metadata={
--- a/tests/canonical/snapshots/tbench_echo_hello/task_description.json
+++ b/tests/canonical/snapshots/tbench_echo_hello/task_description.json
@@ -49,6 +49,7 @@
   "generated_at": null,
   "grading": {
     "combine_method": "weighted",
+    "grading_method": "test_execution",
     "llm_judge": null,
     "pass_threshold": 0.5,
     "state_checks": null,
--- a/tests/unit/test_plugin_first_adapters.py
+++ b/tests/unit/test_plugin_first_adapters.py
@@ -1,0 +1,164 @@
+"""Plugin-first adapter system: the *runner* must not branch on adapter identity.
+
+These pin the capability/declarative seams that replace the old runner-side
+``adapter_type == AdapterType.TERMINAL_BENCH`` branches (tool lifecycle + grading
+method), the open ``adapter_type`` string, and the host-side registry guard — so a
+new adapter can be added as an entry-point package with zero engine edits.
+"""
+
+import pytest
+
+from tolokaforge.adapters import (
+    available_adapters,
+    ensure_registered_adapter,
+    register_adapter,
+)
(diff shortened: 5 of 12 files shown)


CiroGamboa:
### Grading validation (terminal_bench, Docker) — behavior-preserving

Confirmed the `terminal_bench` grading path is unchanged by this PR, including an A/B against `main`:

- **Deterministic grading is correct.** Running the grading commands against the unsolved baseline (compose on host) reads the expected partial-credit reward: `13/30 → reward=0.433333`. So the lifecycle + `test_execution` path produces correct values.
- **A/B vs `main` (same near-baseline state, low `max_turns`):** both `main` and this branch return the **identical** verdict `score: 0.0, binary_pass: False`. The only diff is the reasons label (`terminal-bench reward:` → `test-execution reward:`), which is the intended rename. → no behaviour change.
- **Pre-existing bug surfaced (not introduced here):** in the orchestrated runner-container flow, the grade reads `0.0` even though `test.sh` passes 13/30 — reproduced on `main`. Root cause appears to be the host-socket **log-mount path** where `reward.txt` isn't read back in the runner-driven flow (the host-compose smoke reads it fine). Tracking separately; out of scope for this refactor.

Net: this PR keeps `terminal_bench` grading identical; the `0.0` seen in agent runs is the existing orchestrated-grading issue, independent of the plugin-first changes.

CiroGamboa:
Re-scoped after review: dropped the orchestrator `provides_env_state()` change (environment-state ownership belongs to the runner and is part of a later orchestrator refactor, not an adapter hook). This PR is now scoped to making the **runner** adapter-agnostic + an open `adapter_type` with a host-side registry guard. Force-pushed; full offline suite (1814) + golden gate green; behaviour-preserving (A/B vs main identical).

CiroGamboa:
Follow-up tightenings added (per review): (1) first-party engine code now references the `AdapterType` constants instead of raw adapter-name string literals (registry seed/comparison, orchestrator default, native log), with the enum documented as the canonical home for built-in names; (2) a regression test (`TestRunnerStaysAdapterAgnostic`) that fails if an `adapter_type ==` / `== AdapterType` / adapter `isinstance` branch is reintroduced into the runner's behavioural modules. Full offline suite green (1815).

CiroGamboa:
Improved adapter-author error handling and added explanatory sections per review:
- Strengthened the docstring on `GradingConfig.grading_method` to spell out each value's effect, what `None` means, and what `test_execution` requires the adapter to ship.
- Made the "no exec tool" runtime error actionable: now names the field the adapter set and points to the agent_tools list to fix it.
- Added two adapter-author error-path tests: typo'd `grading_method` is caught at validation; the missing-exec-tool error keeps its actionable phrasing.
- PR body now has a **Why terminal-bench changes too** section explaining the coupling, a **Contract for new adapter authors** section (what to do for default / for test-execution / for custom lifecycle), and an **Error handling** table walking every failure mode.

Suite: 1817 offline tests, golden gate green. Force-pushed (`75b2b72`).

CiroGamboa:
Public-doc updates added where they were genuinely stale; everything else left untouched per "only if really necessary":

- **`docs/TASK_DESCRIPTION_SCHEMA.md`**: corrected the schema for `adapter_type` — was shown as a closed `AdapterType` enum field, now correctly documented as an open `str` from the registry with `AdapterType` as canonical constants. Otherwise unchanged.
- **`docs/ADAPTER_INTERFACE.md`**: added a short **Runtime Capabilities (declarative, opt-in)** section to the adapter author contract, documenting the two new opt-in extension points (`GradingConfig.grading_method` and `ToolWrapper.has_lifecycle`) with code examples — so new adapter authors discover them in the right place.

Not touched (not stale): README, `ADAPTER_ARCHITECTURE.md`, `ADAPTERS.md`, `GRADING.md`, per-adapter docs.

CiroGamboa:
Per review: documented the lifecycle contract's current limitations and tracked the evolution.
- **`docs/ADAPTER_INTERFACE.md`**: added a "Current limitations of the lifecycle contract" subsection under `ToolWrapper.has_lifecycle`, listing the specific gaps (sparse context, sync `start`, no timeout/ordering, only start/stop hooks) and noting the shape is forward-compatible.
- **PR body**: new "Known limitations of the lifecycle contract" section linking to the tracker.
- **Tracker**: GitHub #63 (with the gap list, triggers, and proposed direction).

Force-pushed (`b746d4a`). Suite still 1817 green.

CiroGamboa:
Quick update on the "pre-existing terminal-bench 0.0" note that was in this PR's body: investigated today and **could not reproduce**. Ran the orchestrated billing flow 6 times on clean main (including a truly cold image rebuild) — every run produced the correct baseline reward `0.4333`. Yesterday's `0.0` appears to have been a transient environmental issue that has since resolved.

The note has been removed from the PR body and the tracking issue is closed (#62 / TECHDEL-367 cancelled). This PR's behaviour-preserving claim now has no asterisk.

azorej (approved):

PR #66, opened 2026-06-16
fix(grading): unknown jsonpath operators fail loud + deterministic reasons

### LINKED ISSUE #58: Grading: jsonpath assertions with unrecognized operators silently pass
## Summary

In the grading engine, a state-check jsonpath assertion that uses an **unrecognized
operator** is silently treated as **satisfied** as long as the path exists — the
intended comparison is never applied. For a grading harness this is a correctness
footgun: a mis-specified assertion produces a falsely-passing verdict rather than an
error.

## Details

`StateChecker.check_jsonpaths` only recognizes these operator keys: `equals`,
`equals_ci`, `contains`, `contains_ci` (`tolokaforge/core/grading/state_checks.py`).
If none of those keys is present, control falls through to:

```python
else:
    # No operator specified, just check path exists
    found_match = True
    satisfied += 1
```

So an assertion like `{path: "$.counter", op: "gte", expected: 5}` — where `op` /
`expected` are not keys the checker reads — counts as satisfied whenever `$.counter`
exists, regardless of its value. There is no validation that rejects unknown keys.

## Real occurrence

`tests/data/tasks/calc_custom_checks/grading.yaml` ships exactly this shape:

```yaml
jsonpaths:
  - path: "$.counter"
    op: "gte"
    expected: 5
    required: true
    reason: "Counter should be at least 5"
```

This assertion does **not** check `counter >= 5`; it passes as long as `$.counter`
is present.

## Impact

- Task authors can write assertions that look strict but are no-ops.
- Verdicts can be falsely positive.

## Proposed fix

Make the checker fail loud instead of silent: if an assertion has neither a
recognized operator nor an explicit "exists-only" marker, raise a config/validation
error (or, at minimum, treat the assertion as failed and add a clear reason).
Optionally support the `op`/`expected` shape if it's intended to be valid.

Note: fixing this **changes verdicts** for any task currently relying on the
silent-pass behavior, so it should land behind the grading golden-coverage tests so
the verdict changes are visible and intentional.

### LINKED ISSUE #59: Grading: transcript reason strings embed set reprs (nondeterministic order)
## Summary

Some grading `reasons` strings interpolate Python `set` representations, whose
element order is not guaranteed. This makes the human-readable `reasons` text
nondeterministic across runs/processes (the numeric scores are unaffected).

## Details

In `tolokaforge/core/grading/transcript.py`, tool-expectation reasons are built from
set values:

```python
tools_used = {log.get("tool") for log in tool_log}
...
missing = set(required_tools) - tools_used
if missing:
    reasons.append(f"Missing required tools: {missing}")
...
violations = tools_used & set(disallowed_tools)
if violations:
    reasons.append(f"Used disallowed tools: {violations}")
```

`{missing}` / `{violations}` render as set reprs (e.g. `{'a', 'b'}`), so the order
can differ between runs. Scores are order-independent and unaffected.

## Impact

- `reasons` text isn't reproducible, which complicates snapshotting/diffing grade
  output and any downstream parsing.

## Proposed fix

Render these as sorted, stable strings, e.g.
`", ".join(sorted(missing))`, so the reason text is deterministic.

CiroGamboa:
Two grading-correctness fixes that the golden coverage gate now lets us land
deliberately. No engine architecture change; the goldens flag exactly the two
verdict diffs we intend, and the diffs themselves are the proof of the fix.

Closes #58
Closes #59

## 1. Silent jsonpath operator pass (#58)

`StateChecker.check_jsonpaths` only recognizes `equals` / `equals_ci` /
`contains` / `contains_ci`. Any assertion with another shape — most commonly
the `op: <name>, expected: <value>` pattern that several tasks use — fell
through the no-operator branch and **silently satisfied** as long as the
JSONPath existed. A strict-looking assertion was a no-op.

**Fix.** Unknown-operator assertions now FAIL with an actionable reason that
names the offending keys and lists the supported operators:

```
Assertion at $.counter has no recognized operator
(got keys: ['expected', 'op']); supported operators are
'equals', 'equals_ci', 'contains', 'contains_ci' (Counter should be 5)
```

**Bug laid bare in the snapshots.** The existing canon test
`test_minimal_calculation_state_fail` used `op: "eq"` (also unrecognized) and
its snapshot pinned `score: 1.0` — *identical to the pass snapshot*. The
"fail" assertion had been a no-op the whole time. Fixed test code +
regenerated snapshot now show the real failure with the actionable reason.

Affected data: `tests/data/tasks/calc_custom_checks/grading.yaml` switched
from `op: gte / expected: 5` → `equals: 5` (the deeper "counter >= 5"
semantic is correctly covered by `counter_reached_target()` in `checks.py`).

## 2. Set repr in transcript reasons (#59)

`transcript.py` interpolated Python `set` values into reason text via
`f"...: {missing}"`, so the ordering drifted run-to-run with Python's set
repr. Now rendered as sorted, comma-joined strings:

```
Before:  Missing required tools: {'write_file', 'db_query'}   # unstable order
After:   Missing required tools: db_query, write_file         # alphabetical
```

Scores are unchanged (only reason text). Snapshot regenerated for the new shape.

## Snapshot diffs (intentional)

Only two snapshot files change, and each diff IS the fix:

- `grading_state_calc/fail_result.json` — `"reasons": "", "score": 1.0` →
  `"reasons": "Path $.counter: expected 5, got 0 (Counter should be 5)", "score": 0.0`
- `grading_transcript_calc/fail_result.json` — `"Missing required tools: {'write_file'}"` →
  `"Missing required tools: write_file"`

Pass snapshots and all other goldens are unchanged.

## Tests

Two new test classes pin the new contracts:

- `TestUnknownOperatorFailsLoud` (5 tests) — `op: gte` / typo (`eqaul`) /
  no-operator-at-all / mixed valid+unknown partial credit / regression pin
  for the exact pre-fix silent-pass scenario.
- `TestToolExpectationsReasonsAreDeterministic` (3 tests) — missing reason
  is sorted, violations reason is sorted, repeated calls produce byte-identical
  text.

Suite: **1843 passed, 1 skipped** (was 1817 on main, +26 net new tests).
Identical across two consecutive runs (verified deterministic).

### DIFF
--- a/tests/canonical/snapshots/grading_paths/jsonpath_unknown_operator_fails_loud.json
+++ b/tests/canonical/snapshots/grading_paths/jsonpath_unknown_operator_fails_loud.json
@@ -1,0 +1,10 @@
+{
+  "binary_pass": false,
+  "components": {
+    "custom_checks": null,
+    "llm_judge": null,
+    "state_checks": 0.0,
+    "transcript_rules": null
+  },
+  "score": 0.0
+}
--- a/tests/canonical/snapshots/grading_state_calc/fail_result.json
+++ b/tests/canonical/snapshots/grading_state_calc/fail_result.json
@@ -1,4 +1,4 @@
 {
-  "reasons": "",
-  "score": 1.0
+  "reasons": "Path $.counter: expected 5, got 0 (Counter should be 5)",
+  "score": 0.0
 }
--- a/tests/canonical/snapshots/grading_transcript_calc/fail_result.json
+++ b/tests/canonical/snapshots/grading_transcript_calc/fail_result.json
@@ -1,4 +1,4 @@
 {
-  "reasons": "Missing required tools: {'write_file'}",
+  "reasons": "Missing required tools: write_file",
   "score": 0.875
 }
--- a/tests/canonical/test_grading_canon.py
+++ b/tests/canonical/test_grading_canon.py
@@ -19,10 +19,8 @@ class TestStateCheckerCanon:
         jsonpath_assertions = [
             {
                 "path": "$.counter",
-                "op": "eq",
-                "expected": 5,
-                "required": True,
-                "reason": "Counter should be 5",
+                "equals": 5,
+                "description": "Counter should be 5",
             },
         ]
 
@@ -43,10 +41,8 @@ class TestStateCheckerCanon:
         jsonpath_assertions = [
             {
                 "path": "$.counter",
-                "op": "eq",
-                "expected": 5,
-                "required": True,
-                "reason": "Counter should be 5",
+                "equals": 5,
+                "description": "Counter should be 5",
             },
         ]
 
--- a/tests/canonical/test_grading_paths_canon.py
+++ b/tests/canonical/test_grading_paths_canon.py
@@ -178,11 +178,39 @@ def _scenario_fail_hash_mismatch():
     return config, traj, final_env_state
 
 
+def _scenario_jsonpath_unknown_operator_fails_loud():
+    """End-to-end exercise of the fail-loud branch through GradingEngine: an
+    assertion with `op: gte / expected: 5` (legacy shape, no recognized
+    operator) used to silently satisfy as long as the path existed. Now the
+    engine returns state_checks=0.0 and binary_pass=False even though the
+    JSONPath exists and would have matched the wished-for `gte` semantic.
+    """
+    config = {
+        "combine": {"method": "weighted", "weights": {"state_checks": 1.0}, "pass_threshold": 0.5},
+        "state_checks": {
+            "jsonpaths": [
+                {
+                    "path": "$.db.counter",
+                    "op": "gte",  # unrecognized — used to silently pass
+                    "expected": 5,
+                    "description": "Counter should be at least 5",
+                }
+            ],
+        },
+    }
+    traj = _trajectory([Message(role=MessageRole.USER, content="go")])
+    # counter=7 would have satisfied `gte 5` if it were a real operator; it
+    # used to silently pass for path-existence. Must now fail.
+    final_env_state = {"db": {"counter": 7}}
+    return config, traj, final_env_state
+
+
 _SCENARIOS = {
     "jsonpath_partial": _scenario_jsonpath_partial,
     "transcript_actions_and_info": _scenario_transcript_actions_and_info,
     "weighted_combine": _scenario_weighted_combine,
     "fail_hash_mismatch": _scenario_fail_hash_mismatch,
+    "jsonpath_unknown_operator_fails_loud": _scenario_jsonpath_unknown_operator_fails_loud,
 }
 
 
--- a/tests/data/tasks/calc_custom_checks/grading.yaml
+++ b/tests/data/tasks/calc_custom_checks/grading.yaml
@@ -6,12 +6,10 @@ combine:
   pass_threshold: 0.8
 
 state_checks:
-  jsonpaths:
-    - path: "$.counter"
-      op: "gte"
-      expected: 5
-      required: true
-      reason: "Counter should be at least 5"
+  # No jsonpaths — counter_reached_target() in checks.py owns the "counter >= 5"
+  # assertion (the engine has no gte operator; an `equals: 5` here would tighten
+  # the original "at least 5" semantic).
+  jsonpaths: []
 
 custom_checks:
   enabled: true
--- a/tests/unit/grading/test_state_checks.py
+++ b/tests/unit/grading/test_state_checks.py
@@ -317,3 +317,103 @@ class TestCombinedGrading:
             hash_weight=0.5,
         )
         assert score == pytest.approx(0.5)
+
+
+@pytest.mark.unit
+class TestUnknownOperatorFailsLoud:
+    """Pins the contract: an assertion with no recognized operator must FAIL
+    (with an actionable reason), not silently satisfy on path-existence.
+
+    Recognized operators are ``equals``, ``equals_ci``, ``contains``,
+    ``contains_ci``. Anything else (``op: gte``, typo like ``eqaul``, no operator
+    at all) used to be a silent no-op — this class regression-pins the
+    loud-failure behaviour.
+    """
+
+    @pytest.fixture
+    def checker(self):
+        return StateChecker()
+
+    def test_unknown_operator_keys_fail_with_actionable_reason(self, checker):
+        """An assertion with ``op``/``expected`` (unsupported by the engine)
+        previously satisfied silently as long as the path existed. Now it fails."""
+        state = {"counter": 0}
+        assertions = [
+            {
+                "path": "$.counter",
+                "op": "gte",
+                "expected": 5,
+                "description": "Counter should be at least 5",
+            }
+        ]
+        score, reasons = checker.check_jsonpaths(state, assertions)
+        assert score == 0.0
+        assert len(reasons) == 1
+        r = reasons[0]
+        assert "no recognized operator" in r
+        assert "'equals'" in r and "'contains'" in r
+        assert "op" in r and "expected" in r  # offending keys are surfaced
+
+    def test_no_operator_at_all_fails(self, checker):
+        """An assertion that only specifies ``path`` (no operator) used to
+        silently satisfy; it now fails loudly."""
+        state = {"counter": 0}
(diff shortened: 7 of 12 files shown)


balazs-toloka (dismissed):

CiroGamboa:
Applied the review feedback (force-pushed as `d28b4f2`):

**High-priority change requested — done.**
- `tests/data/tasks/calc_custom_checks/grading.yaml`: dropped the `equals: 5` assertion entirely (`jsonpaths: []`). The engine has no `gte` operator, and `counter_reached_target()` in `checks.py` already owns the "counter ≥ 5" semantic. Keeping an `equals: 5` would have tightened the original "at least 5" intent to "exactly 5". Empty jsonpaths returns 1.0 — effective grading semantics preserved.

**Minor nits — done.**
- `state_checks.py`: error message no longer ends with a stray `()` when description is omitted. Pinned by a new test (`test_actionable_reason_has_no_empty_parens_when_no_description`).
- Added `test_path_glob_with_unknown_operator_fails` covering the `path_glob` + unknown-operator code path.

**Suite:** 1845 passed (was 1843 → +2 new tests), still deterministic.

Not done in this PR (acknowledged as optional in the review):
- A CHANGELOG migration note — could be a small follow-up PR.
- Grep across internal/private adapter repos for the `op:`/`expected:` pattern — needs a cross-repo pass which is out of scope here.

CiroGamboa:
## Thorough testing — results

Per request to test through really well, ran three orthogonal validations.

### 1. New E2E scenario through `GradingEngine.grade_trajectory` (added to this PR)

`tests/canonical/test_grading_paths_canon.py` now has a `jsonpath_unknown_operator_fails_loud` scenario that drives the **full engine pipeline** end-to-end with a config carrying `op: gte / expected: 5`. State has `counter: 7` — which **would have satisfied** a real `gte 5` semantic — yet the verdict is correctly `state_checks=0.0, score=0.0, binary_pass=False`. The fail-loud branch is proven through the engine, not just at the `check_jsonpaths` unit level. New canonical golden pinned.

### 2. Real-example A/B (Anthropic, max_turns=10): main vs this PR

| Example | main | PR | Verdict |
|---|---|---|---|
| `coding_public_example_01` (uses `contains`, `contains_ci`) | score=0.7833 binary_pass=False | score=0.7833 binary_pass=False | ✅ **identical** |
| `tool_use_public_example_01` (uses `equals`, `contains_ci`) | score=0.3500 binary_pass=False | score=0.5667 binary_pass=False | ⚠️ different — explained below |

The tool_use difference is **agent nondeterminism, not a grading regression**:
- Both reasons strings include the same `Files: …` prefix → same code path (`evaluate_jsonpath_file_checks`).
- The differing assertion: PR has `PASS: Plan includes rollback precaution`; main has `FAIL: No files match /env/fs/agent-visible/submissions/*`. In the PR run the agent created a submission file with the rollback wording; in the main run it didn't.
- Trajectory message counts differ (PR=23, main=20) — different conversation paths from the LLM at temp=0.
- Both pipelines correctly grade the state they were given; nothing in this PR changes the file-checks code path.

### 3. Full offline suite

**1846 passed, 1 skipped** (was 1843 → +3 net new tests: the new E2E scenario + the 2 lint-fix tests added in the last review round). Deterministic across runs.

### Overall

- The fix path is **proven E2E through `GradingEngine`** — not just unit-tested.
- The one real example using the same code paths as my fix shows **identical verdicts** on main and PR.
- The one differing run is fully explained by agent nondeterminism in the state the agent produced — the grading layer behaved identically on both branches.

Force-pushed as `eef27ce`.

CiroGamboa:
## Additional E2E verification — internal task A/B

Per request, validated the PR against an internal benchmark task with rich `state_checks.jsonpaths` (real `equals` operators on `$.db.X` paths + hash-based grading via `golden_actions`).

| Branch | score | binary_pass | components |
|---|---|---|---|
| `main`         | 0.3000 | False | state_checks=0.0, transcript_rules=1.0 |
| `fix/grading-correctness` | 0.3000 | False | state_checks=0.0, transcript_rules=1.0 |

**Identical verdict — no regression.** Same score, same components, same reasons (byte-identical).

**Cross-repo migration check.** Grepped all internal task `grading.yaml` files (27 total) for the now-loud `op:`/`expected:` shape and for the `tool_expectations` legacy field this PR's set-repr fix targets:
- **0 tasks use the broken `op:` shape** — no migration needed.
- **0 tasks use `tool_expectations`** — the set-repr fix doesn't change any internal task's grading output either.

So both fixes are purely additive (catch *future* footguns); they touch no currently-shipped internal task behavior.

**Separate observation surfaced (not this PR's concern).** The internal task uses 7 carefully-written `path:`-style jsonpath assertions that are silently SKIPPED by the file-checks evaluator (which only handles `path_glob:`), with reasons like `Files: SKIP: No path_glob — First order is assigned ID O-001`. Same on both branches → pre-existing, unrelated to this PR. Worth filing as a follow-up so those assertions actually evaluate.

CiroGamboa:
## Folded in: sibling silent-skip in runner-side jsonpath evaluator

Per review: the internal-task A/B surfaced the **same family of bug** (silent skip in grading) in a **different code path** — `evaluate_jsonpath_file_checks` (runner-side, used during Docker grading), which only handles `path_glob:` assertions and silently skipped `path:`-style ones with `SKIP: No path_glob`. Folded in a small loud-fail change so the PR is comprehensive within the same scope.

### What changed (added to this PR)

`tolokaforge/runner/grading.py`: when an assertion has no `path_glob`, the runner-side evaluator now emits an **actionable FAIL** instead of a silent skip — and detects `path:`-style assertions specifically, pointing at the host-side evaluator.

### Demonstration on the same internal task

| | Before | After |
|---|---|---|
| Reason for first assertion | `SKIP: No path_glob — First order is assigned ID O-001` | `FAIL: assertion uses 'path' (env-state JSONPath) but this runner-side evaluator only supports 'path_glob' (file glob); path-style assertions are graded host-side — First order is assigned ID O-001` |
| score / binary_pass / components | 0.3000 / False / state=0.0 | 0.3000 / False / state=0.0 |

**No verdict regression** — purely diagnostic. A task author hitting this previously saw a mysterious 0.0 and confusing `SKIP` reasons; now they get an exact pointer to what went wrong.

### What's NOT in this PR (deliberately separated)

The deeper architectural fix — actually routing `path:`-style assertions to the host-side evaluator so they're evaluated rather than just loud-failed — is significantly larger (cross-host/runner boundary, changes the gRPC contract). Leaving as a separate follow-up. The loud-fail makes the symptom visible until that lands.

### Tests added

`TestRunnerSideAssertionsWithoutPathGlobFailLoud` (2 tests) in `tests/unit/test_runner_jsonpath_grading.py` — pins the new loud-fail wording for both shapes (`path:` misrouted and totally-missing target).

### Final suite

**1848 passed, 1 skipped** (was 1846 → +2 new tests). Force-pushed as `e96c594`.

CiroGamboa:
## A/B re-verified properly (replacing my earlier extrapolation)

You caught me being loose — I'd claimed "no verdict regression" for the runner-side change without actually re-running on main with the new code. Did it properly this time: three-way comparison (PR ×2 for determinism + main).

| run | score | binary_pass | components | reasons head |
|---|---|---|---|---|
| **PR #1** | 0.3000 | False | state=0.0, transcript=1.0 | `Files: **FAIL**: assertion uses 'path'... graded host-side — First order is assigned ID O-001;` |
| **PR #2** | 0.3000 | False | state=0.0, transcript=1.0 | (byte-identical to PR #1) |
| **MAIN** | 0.3000 | False | state=0.0, transcript=1.0 | `Files: **SKIP**: No path_glob — First order is assigned ID O-001;` |

- **Determinism (PR×2)**: score, components, **and reasons text byte-identical**.
- **PR vs main**: score / components / binary_pass **identical**. Reasons **differ as designed** — loud-fail vs silent-skip. The change is purely diagnostic.

No regression in the verdict the engine returns; only the human-readable diagnostic improves. Force-push remains `e96c594`.

CiroGamboa:
Follow-up filed for the architectural fix (out of scope here): https://github.com/Toloka/tolokaforge/issues/67 — make `path:`-style jsonpath assertions actually evaluate in Docker mode, rather than just loud-fail. Three design options sketched in the issue (bring JSONPath engine to the runner, share an evaluator module, or route back to host).

balazs-toloka (approved):

PR #68, opened 2026-06-17
feat(llm): add OpenRouter provider routing to ModelConfig

balazs-toloka:
## What
Adds two optional fields to `ModelConfig`: **`provider_order`** (`list[str] | None`) and **`allow_fallbacks`** (`bool`, default `True`). When `provider_order` is set on an `openrouter` model, `_build_kwargs` attaches `extra_body = {"provider": {"order": [...], "allow_fallbacks": ...}}` to the litellm call, so OpenRouter pins the request to the listed upstream provider(s) instead of routing to (and falling back to) a rate-limited one.

## Why
Several leaderboard models hit OpenRouter's shared-pool rate limit (HTTP 429, `is_byok:false`) because the provider OpenRouter routes them to is saturated. BYOK was ruled out (we use OpenRouter precisely to avoid self-managed provider keys/billing). The agreed alternative is provider routing: pin to a provider that still has capacity.

Example: `nvidia/nemotron-3-ultra-550b-a55b` is served by DeepInfra (saturated for us) and Together; setting `provider_order: ["Together"]`, `allow_fallbacks: false` avoids the 429. (Single-provider models like `mistralai/mistral-medium-3-5` have no alternative provider, so this does not help them.)

## Scope
- `tolokaforge/core/models.py`: 2 optional `ModelConfig` fields (default `None` / `True` => no behavior change).
- `tolokaforge/core/llm/client.py`: pass-through in `_build_kwargs`, OpenRouter block only.
- `tests/unit/test_llm_providers.py`: `TestProviderRouting` (3 tests: routing is set, no-routing back-compat, native provider ignores it).

## Notes
- Backward-compatible: with no field set, no provider routing is sent and behavior is unchanged.
- This adds the capability only. The actual per-model `provider_order` values go into the task configs after we test which providers have spare capacity.
- Local: 24/24 in `test_llm_providers.py` pass, plus 310 config/model unit tests green. PR CI (`ci.yml`) runs unit + canonical (no live integration), so no 429 risk in CI.

### DIFF
--- a/docs/ADD_NEW_MODEL.md
+++ b/docs/ADD_NEW_MODEL.md
@@ -74,6 +74,13 @@ Available policy slots (see
 | `cache_policy` | `none` | `anthropic_ephemeral` |
 | `params` | — | arbitrary dict — see `GenerationParams` |
 
+> **Rate-limited model?** If OpenRouter's default provider 429s the eval
+> (`is_byok:false`, "rate-limited upstream"), pin the model to a provider that
+> has capacity with an `openrouter:` block in the model config (OpenRouter-only,
+> validated): `openrouter: {provider_order: ["Together"], allow_fallbacks: false}`.
+> Slugs are case-sensitive OpenRouter provider names. This is a per-model routing
+> knob, separate from the preset policy above.
+
 ## 3. Add a ModelCertificate
 
 Append to `ALL_MODELS` in
--- a/tests/unit/test_llm_providers.py
+++ b/tests/unit/test_llm_providers.py
@@ -371,3 +371,78 @@ class TestReasoningParameter:
         """ModelConfig default reasoning must have ``mode='off'`` (opt-in)."""
         config = ModelConfig(provider="openrouter", name="test/model")
         assert config.reasoning.mode == "off"
+
+
+class TestProviderRouting:
+    """``ModelConfig.provider_order`` pins OpenRouter requests to specific upstream providers."""
+
+    def _capture_kwargs(
+        self,
+        provider: str,
+        name: str,
+        provider_order: list[str] | None = None,
+        allow_fallbacks: bool = True,
+    ) -> dict:
+        import tolokaforge.core.llm.client as mc_module
+
+        captured: dict = {}
+
+        def fake_completion(**kwargs):
+            captured.update(kwargs)
+            mock_resp = MagicMock()
+            mock_resp.choices = [MagicMock()]
+            mock_resp.choices[0].message.content = "ok"
+            mock_resp.choices[0].message.tool_calls = None
+            mock_resp.choices[0].message.reasoning_content = None
+            mock_resp.choices[0].message.thinking_blocks = None
+            mock_resp.usage = MagicMock()
+            mock_resp.usage.prompt_tokens = 10
+            mock_resp.usage.completion_tokens = 5
+            return mock_resp
+
+        config = ModelConfig(
+            provider=provider,
+            name=name,
+            provider_order=provider_order,
+            allow_fallbacks=allow_fallbacks,
+        )
+        client = LLMClient(config)
+
+        original = mc_module.completion
+        mc_module.completion = fake_completion
+        try:
+            client.generate(
+                system="test",
+                messages=[Message(role=MessageRole.USER, content="hello")],
+            )
+        finally:
+            mc_module.completion = original
+        return captured
+
+    def test_provider_order_sets_extra_body_provider(self):
+        """provider_order pins the request to the listed upstream provider(s)."""
+        kwargs = self._capture_kwargs(
+            "openrouter",
+            "nvidia/nemotron-3-ultra-550b-a55b",
+            provider_order=["Together"],
+            allow_fallbacks=False,
+        )
+        assert kwargs.get("extra_body", {}).get("provider") == {
+            "order": ["Together"],
+            "allow_fallbacks": False,
+        }
+
+    def test_no_provider_order_means_no_routing(self):
+        """Without provider_order, no provider routing is sent (back-compat)."""
+        kwargs = self._capture_kwargs("openrouter", "nvidia/nemotron-3-ultra-550b-a55b")
+        assert "provider" not in kwargs.get("extra_body", {})
+
+    def test_provider_order_ignored_for_native_provider(self):
+        """provider_order is OpenRouter-only; native providers ignore it."""
+        kwargs = self._capture_kwargs(
+            "anthropic",
+            "claude-opus-4.6",
+            provider_order=["Together"],
+            allow_fallbacks=False,
+        )
+        assert "provider" not in kwargs.get("extra_body", {})
--- a/tests/unit/test_llm_providers.py
+++ b/tests/unit/test_llm_providers.py
@@ -11,7 +11,7 @@ import pytest
 
 from tolokaforge.core.llm import GenerationResult, LLMClient, UserSimulator
 from tolokaforge.core.llm.usage import Usage
-from tolokaforge.core.models import Message, MessageRole, ModelConfig, ToolCall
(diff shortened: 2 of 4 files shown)


bberkes-toloka:
### Local testing

Run on Python 3.12 (engine venv):
- `tests/unit/test_llm_providers.py` — **24/24 pass** (3 new `TestProviderRouting` + 21 existing).
- config/model unit tests (`pytest tests/unit/ -k "config or model"`) — **310 pass, 0 fail** (no regression from the new `ModelConfig` fields).
- `ast` syntax check on all 3 changed files — clean.

### Methodology (the 3 new tests)

`TestProviderRouting` monkey-patches the module-level `completion` in `tolokaforge.core.llm.client` to capture the kwargs `generate()` would hand to litellm, then asserts on `extra_body["provider"]` (mirrors the existing `TestReasoningParameter` pattern). Three cases:

1. **Routing set** — `provider_order=["Together"]`, `allow_fallbacks=False` on an `openrouter` model => `extra_body["provider"] == {"order": ["Together"], "allow_fallbacks": False}`.
2. **Back-compat** — no `provider_order` => no `provider` key in `extra_body` (behavior unchanged).
3. **Native provider ignored** — `provider="anthropic"` with `provider_order` set => no `provider` routing (the pass-through lives only in the OpenRouter branch of `_build_kwargs`).

PR CI (`ci.yml`) re-runs these as part of unit + canonical (no live integration), so there is no 429 exposure in CI.

CiroGamboa:
Thanks for getting this out fast — the 429s are a real blocker and the change is correctly scoped and back-compat. One design question I think is worth pausing on before we merge, because the shape we pick here will be the template for the next provider-specific knob.

## The concern: provider vocabulary on a cross-provider type

`provider_order` and `allow_fallbacks` are **OpenRouter's exact API shape**, but they're being added to the top level of `ModelConfig` — the generic config every model carries (Anthropic native, OpenAI native, Bedrock, anything future). A few things follow from that:

1. **The schema leaks the OpenRouter mental model upward.** Every Anthropic/OpenAI/Bedrock model config in the repo now advertises `provider_order` / `allow_fallbacks` fields that mean nothing for it. When the next provider needs its own routing knob (Bedrock cross-region inference profiles, Vertex regional pinning, a future gateway's routing rules), the pattern this PR sets says "add another flat field to `ModelConfig`." That accumulates fast.

2. **Misuse is silently ignored.** `test_provider_order_ignored_for_native_provider` actually *codifies* that setting `provider_order: ["Together"]` on an `anthropic` model does nothing — no warning, no error. This runs against two principles we've been trying to hold in the engine:
   - **Honesty over convenience / no silent fallbacks.** A misconfigured field should surface, not vanish. (Same spirit as the loud-fail work in #66.)
   - **Plugin-first — interface + default.** Things that vary by provider should sit behind a narrow per-provider seam, not be hard-wired into the shared type.

3. **`allow_fallbacks: bool = True` at the top level is structurally separated from the field it qualifies.** It only has meaning when `provider_order` is set; defaulting it on every model config is schema noise.

The implementation in `_build_kwargs` is fine — it's correctly gated by `if self.provider.startswith("openrouter")` — so the runtime behavior is right. The question is purely about where the *fields* live in the config model.

## Suggestion

Nest the OpenRouter-specific knobs under a provider-specific block. This makes the coupling explicit, keeps `ModelConfig` clean, and gives the next provider its own block instead of forcing another flat field:

```python
class OpenRouterConfig(BaseModel):
    provider_order: list[str] | None = None
    allow_fallbacks: bool = True

class ModelConfig(BaseModel):
    ...
    openrouter: OpenRouterConfig | None = None
```

Then a `model_validator` raises loudly if `openrouter:` is set on a non-`openrouter` provider, and the test flips from "silently ignored" to "raises a config error pointing at the right field." YAML stays clean too — only OpenRouter models carry an `openrouter:` block.

## Smaller notes

- `kwargs["extra_body"].setdefault("provider", provider_pref)` quietly preserves any pre-existing `provider` value in `extra_body`. Today nothing else sets it; if that changes, the collision will be invisible. Prefer a direct assignment, or assert no prior value, so a future caller doesn't get a confusing silent override.
- `provider_order` values are case-sensitive OpenRouter slugs (`"Together"`, `"DeepInfra"`). Worth a docstring example so the next person setting them doesn't have to dig through OpenRouter docs to find the casing — and so typos fail at config time rather than producing a 404 from OpenRouter at runtime.
- Worth a one-liner in `docs/ADD_NEW_MODEL.md` (or `LLM_LAYER.md`) so the next person registering a model knows this lever exists and which provider it applies to.
- No CHANGELOG entry — small, but this is a user-visible config addition.

## Bottom line

Not against the capability — it's the right answer to the 429 problem and BYOK was rightly ruled out. The thing I'd like to see changed before merge is the nested-block shape above (plus the loud validation that comes with it), so we set a clean template for the next provider knob that lands and don't carry silently-ignored config fields into the schema.

bberkes-toloka:
Thanks for the thorough review. All points addressed in `1c5111c`:

- **Nested block.** `provider_order` / `allow_fallbacks` now live in a dedicated `OpenRouterConfig`, surfaced as `ModelConfig.openrouter: OpenRouterConfig | None`. The shared type no longer advertises OpenRouter-specific fields to every model, and the next provider knob gets its own block instead of another flat field.
- **Loud validation.** A `model_validator(mode="after")` rejects an `openrouter:` block on a non-`openrouter` provider with a clear error, instead of silently ignoring it.
- **Test flipped.** `test_openrouter_block_rejected_for_native_provider` now asserts a `ValidationError` (config-time failure), replacing the old "silently ignored" test.
- **No silent override.** `_build_kwargs` assigns `extra_body["provider"]` directly (dropped the `setdefault`), so a future caller that sets `provider` is not silently overridden.
- **Slug casing.** `OpenRouterConfig`'s docstring documents the case-sensitive slugs with examples (`["Together"]`, `["DeepInfra", "Nebius"]`).
- **Docs.** `ADD_NEW_MODEL.md` §2 now notes the lever (OpenRouter-only + validated, with the casing).
- **CHANGELOG.** The repo's CHANGELOG is commitizen-generated from Conventional Commits at release (`update_changelog_on_bump = true`, `changelog_incremental = true`), so the `feat(llm): ...` commit yields the entry at the next `cz bump`. A hand-written entry would be redundant with that, but happy to pre-stage one if you'd prefer.

Verification: `test_llm_providers.py` 24/24, config/model unit tests 310/0, `ruff check` clean.

CiroGamboa:
Looks good — every point landed cleanly. Skimming the new diff:

- `OpenRouterConfig` is well-shaped; the docstring with the case-sensitive slug examples and the link to the OpenRouter routing docs is exactly the kind of breadcrumb that saves the next person an hour.
- `_reject_openrouter_on_other_providers` is the right form (after-validator, clear error pointing at the offending `provider` value), and `test_openrouter_block_rejected_for_native_provider` flipping to assert `ValidationError` is the principled outcome.
- `kwargs.setdefault("extra_body", {})["provider"] = {...}` is fine — the `setdefault` is only on the container; the `provider` key is direct assignment, so a future caller that pre-populates `provider` will be visibly overwritten rather than silently lose its value.
- `ADD_NEW_MODEL.md` callout is well-placed; the "OpenRouter-only + validated" framing is exactly what a future reader needs.
- CHANGELOG via commitizen — verified `pyproject.toml` has `update_changelog_on_bump = true` + `changelog_incremental = true`, so the `feat(llm): …` commit will produce the entry at the next `cz bump`. Agreed, no hand-written entry needed.

One tiny nit, non-blocking: the class docstring on `TestProviderRouting` still says ``"``ModelConfig.provider_order`` pins OpenRouter requests…"`` — that path no longer exists, it's now `ModelConfig.openrouter.provider_order` (or just `OpenRouterConfig.provider_order`). Worth updating so the docstring doesn't lie about the field path.

Otherwise this is good to merge from my side.

CiroGamboa (approved):
LGTM

PR #69, opened 2026-06-17
feat(presets): operator-overridable preset overlay file

### LINKED ISSUE #64: Decouple model policy/codec layer from engine release cadence
The model-specific policy/codec layer (`tolokaforge/core/llm/`) plus `model_presets.yaml` shipped inside the wheel make every new failure pattern surfaced during evaluation require an engine release. This couples policy iteration to engine cadence.

## Today

- Policy classes are registered in `_RESPONSE_POLICIES`, `_REASONING_CODECS`, `_CONTENT_POLICIES`, etc. inside `tolokaforge/core/llm/presets.py`.
- `tolokaforge/core/data/model_presets.yaml` ships inside the wheel.

So the typical loop is: run integration tests → release with presets → run a larger-sample eval → if a custom failure pattern fixable by policy surfaces, release again. Two engine releases per new model when custom patterns appear.

## Proposed directions (cheapest first)

1. **Runtime-overridable preset file** (`--presets-file` / config path). Adding a new model that reuses existing policy classes becomes a config change with no release.
2. **Policies/codecs as plugins** via `importlib.metadata.entry_points`, mirroring the adapter plugin pattern. Custom policies ship as independent packages.
3. **Optional sidecar Python module** loaded from the run config — escape hatch for one-off experiments.

## Why it matters

Same shape as the open-`AdapterType`-enum and `tool_factory` decoupling items already on the board: the engine release shouldn't be the unit of iteration for things that move at a different cadence. Useful input for the architecture proposal that is currently being prepared.

CiroGamboa:
Closes #64

## What this changes for model-adders

**Today.** Registering a new model — even one that reuses existing
policy classes — means editing `tolokaforge/core/data/model_presets.yaml`
(and/or `pricing.json`, `tests/integration/llm/registry.py`) inside the
wheel, cutting a release, and rolling that release out before you can
run smoke eval. If smoke surfaces a model-specific failure pattern, the
fix is to re-route the model to a different existing policy → another
release. Two releases per model is common.

**After this PR.** For models that reuse existing policy classes
(`StandardResponse`, `AnthropicContent`, `AnthropicEphemeralCache`,
`AnthropicReasoningCodec`, …), the registration is a YAML file you hand
in at run time. Zero release.

```yaml
# my_overlay.yaml — same schema as the bundled model_presets.yaml
presets:
  my_provider_new_model:
    match: ["my-provider/new-model*"]
    content_policy: anthropic
    reasoning_codec: anthropic
    cache_policy: anthropic_ephemeral
```

```bash
tolokaforge run --config run.yaml --presets-file my_overlay.yaml
```

Two ways to point at the overlay, in precedence order:

1. **`--presets-file <path>` (CLI flag)** — for one-off overlays:
   smoke-eval iterations, ablations, anything you don't want to commit
   alongside the run config. Available on `run`, `prepare`, `worker`,
   and `config validate`.
2. **`engine.presets_file` in the run config** — when an overlay is
   part of *what this benchmark is*, commit it next to the rest of the
   run definition:
   ```yaml
   engine:
     presets_file: ./my_overlay.yaml
   ```

**What still needs an engine release.** A brand-new policy *class*
(novel response post-processing, new reasoning codec, new content
format). The overlay validator rejects unknown policy names at startup
with a clear error — by design. The boundary is *data + declaration →
externalizable; new classes → engine code*. This matches the
architecture proposal's framing for the model registry seam.

## How to verify your overlay before running

`tolokaforge config validate` now validates the overlay alongside the
run config:

```bash
tolokaforge config validate --config run.yaml --presets-file my_overlay.yaml
# ✓ Preset overlay OK: my_overlay.yaml
# (or, if you have a typo:)
# ✗ Preset overlay 'my_overlay.yaml' at presets.my_model: unknown
#   response_policy 'standerd'. Available: ['array_dict_map',
#   'json_coerce', 'minimax_m3_tags', 'standard', 'unwrap_input'].
```

The error always names (a) the overlay file path, (b) the offending key,
and (c) the available registry keys. Same loud-fail discipline as the
existing `capabilities:` override surface in run configs.

## Workflow before/after

| Step | Before | After (data-only change) |
|---|---|---|
| Add a preset | Edit `model_presets.yaml` in tree | Add to overlay YAML, no PR needed |
| Pricing | Edit `pricing.json` in tree | Use LiteLLM's `LITELLM_MODEL_COST_MAP_URL` (separate channel, already supported by LiteLLM — not changed by this PR) |
| Capability cert | Edit `tests/integration/llm/registry.py` | Same as before (this PR doesn't touch the certificate suite — integration tests still gate the engine release path) |
| Smoke eval | Wait for release rollout | Run with `--presets-file` immediately |
| Re-route after smoke finding | Another release | Edit overlay, re-run |
| Brand-new policy class | Release required | Release still required (correct boundary) |

For repeat registrations of the same set of models across multiple
runs, the run-config field (`engine.presets_file`) keeps the overlay
declarative — no flag threading.

## Distributed-worker note

`tolokaforge prepare --presets-file overlay.yaml` writes the resolved
overlay path into `engine_run_state.json` alongside the run queue.
`tolokaforge worker` invocations against the same `--run-dir` then
inherit the overlay automatically. Worker-side `--presets-file` still
wins if you set it (useful for A/B-style ablations across worker
fleets).

## Design decisions

**Mirrors the adapter plugin pattern's externalization (PR #61).** The
adapter side opened `TaskDescription.adapter_type` to a registry-sourced
string and added `ensure_registered_adapter()` as a host-boundary
validator. This PR does the same shape one layer up for the LLM preset
layer: opens the preset registry to operator data, validates at the
host boundary at load time, keeps every downstream caller untouched.
Loud-fail messages name the file path the same way
`ensure_registered_adapter` names the registry.

**Bit-identical behaviour when no overlay is configured.** The loader's
no-overlay path returns the bundled YAML verbatim. Golden grading
regression suite (PR #57) and the canonical-suite preset routing tests
all pass unchanged. The full offline suite went from 1817 (PR #61
baseline) to 1911 here — net +94 from new overlay tests; zero
regressions.

**Validator-versus-registry sync, pinned by a unit test.** Adding a new
policy class to the engine (in any of the six registries) must extend
`_validate_overlay` so it knows the new name is allowed. A unit test
scans `_validate_overlay`'s source for every registry constant by name;
a future contributor adding a registry without touching the validator
gets a test failure rather than a silent overlay-validation gap.

**Forward-compatible with the entry-point plugin path.** The data shape
the overlay produces is the same dict the bundled YAML produces, so a
future `tolokaforge.model_presets` entry-point group (the natural Phase
2 if operators want shippable preset bundles) can populate the same
structure without a parallel mechanism. Documented in ADR 0002
*Considered Options*.

**Pricing not externalized in this PR.** LiteLLM already supports an
external model cost map via `LITELLM_MODEL_COST_MAP_URL`. Duplicating it
inside the engine would create a second source of truth. Noted in
ADR 0002 *Trade-offs*.

**ADR scaffolding bundled here.** The `docs/architecture/adr/`
directory and its three scaffolding files (template, README, ADR 0001)
are copied verbatim from the `docs/architecture-overview` branch
(PR #11) so this PR cannot conflict with #11 regardless of merge order.
If #11 lands first, rebasing this PR is `git rm` on those four files.

## Test plan

- [x] **Unit tests (36):** loader merge, precedence (overlay-first
      iteration order), default/provider deep-merge, cache invalidation,
      every loud-fail path (missing file, malformed YAML, non-mapping
      top-level, unknown top-level keys, unknown policy names in
      `presets/default/providers`, unknown `params` keys), validator-
      versus-registry sync, lazy file-read.
- [x] **Integration tests (15):** real subprocess proving
      `engine_run_state.json` is the actual cross-process carrier
      between `prepare` and `worker`; real `tolokaforge config validate`
      CLI accepts good overlays and rejects bad ones with file-path-
      bearing errors; in-process orchestrator integration through
      `_build_resolved_block` (the helper that stamps the resolved
      preset name into per-trial `task.yaml`).
- [x] **Live eval (1, gated on API key + Docker):** real
      `tolokaforge run` against Anthropic via the public
      `examples/native/tool_use/dataset` task — 1 trial, 9 turns,
      ~$0.05. Asserts the per-trial `task.yaml` records the overlay's
      preset name in `model_config.agent.resolved.effective_preset` and
      that all four overridden policy slots landed in the fingerprint.
- [x] **Full offline regression:** `uv run pytest tests/unit
      tests/canonical -q` → **1895 passed, 1 skipped** (skip is
      pre-existing). Verdicts on the golden grading regression suite
      (PR #57) bit-identical to `main`.
- [x] **CLI smoke (real invocations):**
  - `tolokaforge config validate --config run.yaml --presets-file
    good.yaml` → exit 0, `✓ Preset overlay OK`.
  - Same with `bad.yaml` (typo'd policy name) → exit 1, error names
    file path + offending key + available alternatives.
  - Same with a missing file → exit 1, error names the path.

## Docs

- `docs/CONFIG.md` § **Preset overlay file** — the field, precedence,
  schema, and distributed-worker propagation.
- `docs/ADD_NEW_MODEL.md` § **Adding a model without an engine release** —
  5-line walkthrough that the next model-adder can follow directly.
- `docs/LLM_LAYER.md` — one-paragraph pointer at the preset loader.
- `docs/architecture/adr/0002-external-model-registry.md` — Proposed.

### DIFF
--- a/docs/ADD_NEW_MODEL.md
+++ b/docs/ADD_NEW_MODEL.md
@@ -81,6 +81,47 @@ Available policy slots (see
 > Slugs are case-sensitive OpenRouter provider names. This is a per-model routing
 > knob, separate from the preset policy above.
 
+### Adding a model without an engine release — preset overlay
+
+If the model reuses existing policy classes (which is the common case), you
+don't need to cut a new wheel. Put the same preset entry in a separate YAML
+file and point the engine at it:
+
+```yaml
+# my_overlay.yaml — same schema as model_presets.yaml
+presets:
+  my_provider_new_model:
+    match: ["my-provider/new-model*"]
+    content_policy: anthropic
+    reasoning_codec: anthropic
+    cache_policy: anthropic_ephemeral
+```
+
+Then run:
+
+```bash
+tolokaforge run --config run.yaml --presets-file my_overlay.yaml
+```
+
+Or set the overlay declaratively in `run.yaml` under `engine.presets_file`,
+or via `$TOLOKAFORGE_PRESETS_FILE`. Precedence is CLI > env > config field.
+See [`docs/CONFIG.md`](CONFIG.md#preset-overlay-file-no-engine-release-required)
+and [ADR 0002 — External model registry](architecture/adr/0002-external-model-registry.md).
+
+A few rules the overlay enforces (loud-fail at engine startup, naming the
+overlay path and the offending key):
+
+- Policy-name strings must resolve to existing classes shipped in the engine.
+  Adding a brand-new policy class still requires an engine release.
+- Overlay presets are prepended to the iteration order, so overlapping
+  `match:` globs let you shadow a bundled preset.
+- Same-named overlay presets replace the bundled entry (logged at INFO so the
+  swap is visible).
+
+For distributed runs, the overlay path passed to `tolokaforge prepare` is
+persisted into the run queue's state file. Subsequent `tolokaforge worker`
+invocations pick it up automatically without re-specifying `--presets-file`.
+
 ## 3. Add a ModelCertificate
 
 Append to `ALL_MODELS` in
--- a/docs/ADD_NEW_MODEL.md
+++ b/docs/ADD_NEW_MODEL.md
@@ -103,9 +103,9 @@ Then run:
 tolokaforge run --config run.yaml --presets-file my_overlay.yaml
 ```
 
-Or set the overlay declaratively in `run.yaml` under `engine.presets_file`,
-or via `$TOLOKAFORGE_PRESETS_FILE`. Precedence is CLI > env > config field.
-See [`docs/CONFIG.md`](CONFIG.md#preset-overlay-file-no-engine-release-required)
+Or set the overlay declaratively in `run.yaml` under `engine.presets_file`.
+Precedence is CLI > config field. See
+[`docs/CONFIG.md`](CONFIG.md#preset-overlay-file-no-engine-release-required)
 and [ADR 0002 — External model registry](architecture/adr/0002-external-model-registry.md).
 
 A few rules the overlay enforces (loud-fail at engine startup, naming the
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -196,6 +196,46 @@ presets:
     cache_policy: none             # none | anthropic_ephemeral (Anthropic only)
 ```
 
+### Preset overlay file (no engine release required)
+
+For models that reuse existing policy classes, you don't need to release the
+engine to add or adjust a preset. Point TolokaForge at a second YAML file that
+gets merged onto the bundled `model_presets.yaml` at startup. See
+[ADR 0002 — External model registry](architecture/adr/0002-external-model-registry.md)
+for the rationale and [`docs/ADD_NEW_MODEL.md`](ADD_NEW_MODEL.md) for a
+walkthrough.
+
+Set the overlay path three ways; precedence is **CLI flag > env var > config field**:
+
+1. CLI flag: `--presets-file overlay.yaml` on `run`, `prepare`, `worker`, and
+   `config validate`.
+2. Env var: `TOLOKAFORGE_PRESETS_FILE=/path/to/overlay.yaml`.
+3. Run-config field:
+
+   ```yaml
+   engine:
+     presets_file: ./overlay.yaml   # relative to the working directory
+   ```
+
+The overlay file uses the same schema as `model_presets.yaml`. Overlay
+presets are prepended to the iteration order so first-match-wins lets you
+shadow a bundled preset. Same-named overlay presets *replace* the bundled
+entry (logged at INFO so the replacement is visible). Policy-name strings
+(`schema_sanitizer`, `prompt_policy`, etc.) must resolve to a class already
+shipped in the engine — overlays can compose existing policies but cannot
+introduce new classes. Unknown names raise `ValueError` at startup naming
+both the overlay file and the offending key.
+
+#### Distributed-worker propagation
+
+`tolokaforge prepare --presets-file overlay.yaml ...` persists the resolved
+overlay path into `engine_run_state.json` alongside the run queue. Worker
+subprocesses launched later from the same `--run-dir` pick it up
+automatically — you don't have to thread the flag through every
+`tolokaforge worker` invocation. A worker-side `--presets-file` flag (or
+`$TOLOKAFORGE_PRESETS_FILE`) still wins over the persisted value when both
+are set.
+
 ## Task Specification (`task.yaml`)
 
 ```yaml
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -205,12 +205,13 @@ gets merged onto the bundled `model_presets.yaml` at startup. See
 for the rationale and [`docs/ADD_NEW_MODEL.md`](ADD_NEW_MODEL.md) for a
 walkthrough.
 
-Set the overlay path three ways; precedence is **CLI flag > env var > config field**:
+Set the overlay path two ways; precedence is **CLI flag > config field**:
 
 1. CLI flag: `--presets-file overlay.yaml` on `run`, `prepare`, `worker`, and
-   `config validate`.
-2. Env var: `TOLOKAFORGE_PRESETS_FILE=/path/to/overlay.yaml`.
-3. Run-config field:
+   `config validate`. Use this for one-off overlays — smoke-eval iterations,
+   ablations, anything you don't want to commit alongside the run config.
+2. Run-config field — commit the overlay path next to the run definition
+   when an overlay is part of *what this benchmark is*:
 
    ```yaml
    engine:
@@ -232,9 +233,8 @@ both the overlay file and the offending key.
 overlay path into `engine_run_state.json` alongside the run queue. Worker
 subprocesses launched later from the same `--run-dir` pick it up
 automatically — you don't have to thread the flag through every
-`tolokaforge worker` invocation. A worker-side `--presets-file` flag (or
-`$TOLOKAFORGE_PRESETS_FILE`) still wins over the persisted value when both
-are set.
+`tolokaforge worker` invocation. A worker-side `--presets-file` flag still
+wins over the persisted value when both are set.
 
 ## Task Specification (`task.yaml`)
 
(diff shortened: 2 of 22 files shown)


bberkes-toloka:
Took a closer pass on the overlay loader/validator. The merge and resolution path itself checks out (first-match-wins shadowing via the prepended presets, `cli > queue-state > config` precedence, loud-fail validation, bit-identical no-overlay path). Four items I'd flag before merge, two High and two Medium. A few lower-priority design nits omitted.

## High

### 1. The validator/registry sync test does not deliver the guarantee it claims

`test_preset_overlay_validation.py::TestValidatorRegistrySync::test_validator_references_every_registry` asserts a hardcoded 6-name list appears in `inspect.getsource(_validate_overlay)`. That catches *removal* of an existing check, not *addition* of a new registry: introduce a `_NEW_POLICIES` registry + policy class and forget to extend `_validate_overlay`, and the test stays green. That is exactly the silent overlay-validation gap the PR description says this test prevents ("a future contributor adding a registry without touching the validator gets a test failure"). The test's own hardcoded list is the thing that would need updating, and nothing forces it.

Suggested fix: lift the local `registries` dict inside `_validate_overlay` to a module-level `_POLICY_REGISTRIES` constant, consumed by both the validator and the test.

One caveat so it actually closes the loop: a single source still misses "added a registry but forgot to add it to `_POLICY_REGISTRIES`". To truly deliver the promised guarantee, have the test introspect the module for registry-shaped globals (by the `_*_POLICIES` / `_*_SANITIZERS` / `_*_CODECS` naming convention, or "dict of str -> type") and assert each one is present in `_POLICY_REGISTRIES`.

### 2. The `run` comment promises early validation that does not happen; prefer failing eagerly

In `cli/main.py`, the comment above `_activate_presets_overlay` says "Triggers loud-fail validation on the overlay file at this point." It does not: `set_overlay_path` only stores the path and clears the cache (`TestSetOverlayPathDefersFileRead` pins exactly this). The first read/validation actually happens later, at `LLMClient.__init__` -> `build_capabilities` -> `_load_presets` -> `_load_overlay_file` (in `orchestrator.run()` and `run_worker()`), i.e. *after* the orchestrator is constructed, `load_tasks()` has run, and (for `run`) the Docker stack has auto-started. So a typo'd `engine.presets_file` only fails after Docker spins up.

Two options: (a) fix the comment, or better (b) eagerly call `_load_overlay_file(resolved)` right after `set_overlay_path` so a bad overlay fails immediately with the file-path-bearing error, before the orchestrator / Docker / task load. Option (b) also makes `run` / `prepare` / `worker` consistent with `config validate`, which already validates eagerly.

## Medium

### 3. `config_commands.py` reaches across the module boundary for a private function

It imports `_load_overlay_file` from `presets`, but `__all__` only exports the three `*_overlay_path` helpers (`_load_overlay_file` and `_validate_overlay` are not in it). Consider exposing a public `validate_overlay_file(path)` in `__all__` and having `config validate` consume that instead of the underscore-prefixed function.

### 4. `overlay_isolation` is opt-in, not autouse, over shared mutable module state

`overlay_isolation` (in `tests/conftest.py`) resets the module-level `_OVERLAY_PATH`, but it is a plain opt-in fixture whose docstring says every test that calls `set_overlay_path` "must take this fixture". That is discipline-based isolation over a shared mutable global, a classic order-dependent flakiness source. Making it `autouse=True` (scoped to the overlay test packages) removes the footgun.

Related, worth fixing together: the comment in `presets.py` says "the `overlay_isolation` fixture in `tests/unit/llm/conftest.py` handles this", but `tests/unit/llm/conftest.py` is empty and the fixture lives in `tests/conftest.py` and is not autouse, so it currently handles nothing automatically.

---

None of these are correctness bugs in the capability-resolution path. 1 and 2 are the two I'd gate on, since they undercut guarantees the PR explicitly states; 3 and 4 are cheap hygiene.

CiroGamboa:
Thanks for the close pass — all four addressed in `1964b23`. Quick critique-then-fix on each so the rationale is in the trail:

## High 1 — validator/registry sync test

Agreed, my original test caught *removal* but not *addition*. Fix matches your suggestion:

- Lifted the local `registries` dict to a module-level `_POLICY_REGISTRIES` constant in `presets.py` — single source of truth shared by `_validate_overlay` and the test.
- Rewrote the test to **introspect the `presets` module** for every module-level `dict[str, type]` (the structural shape of every existing registry), then assert each discovered dict is a value in `_POLICY_REGISTRIES`. Went with the structural check rather than the `_*_POLICIES` / `_*_SANITIZERS` naming convention you mentioned because the structural one survives future renames.
- Manually verified the new test catches the exact scenario you described:
  ```
  $ python -c "from tolokaforge.core.llm import presets as p; p._FAKE_NEW_POLICIES = {'foo': str}; ..."
  OK: test caught the bypass:
    presets module contains policy registries not listed in _POLICY_REGISTRIES: ['_FAKE_NEW_POLICIES']. Add them to _POLICY_REGISTRIES so _validate_overlay checks them...
  ```
- Kept a belt-and-braces `test_baseline_six_registries_are_present` so a future *removal* also surfaces.

## High 2 — eager validation

Agreed, the comment was lying. Verified your call-chain trace: lazy read → first `_load_presets()` happens inside `Orchestrator.run()` after Docker auto-starts → ~30s wasted on bad overlays.

Took **option (b)**: `_activate_presets_overlay` now calls `validate_overlay_file(resolved)` immediately after `set_overlay_path`. The lazy contract on `set_overlay_path` itself is unchanged (the `TestSetOverlayPathDefersFileRead` unit tests still pass), but the CLI helper layer is now eager — same shape as `tolokaforge config validate`. Measured impact: a typo'd overlay now exits in **~1.5s instead of ~30+s**.

New `TestEagerValidation` class pins the behaviour (missing file, malformed overlay, no-overlay-no-validation).

## Medium 3 — public surface

Done. Added `validate_overlay_file(path) -> None` to `__all__`; both `config validate` and the new CLI eager-validation path use it. `_load_overlay_file` stays private (still returns the parsed dict for the loader's internal use); the public wrapper just delegates and discards the return value, which matches the caller intent.

## Medium 4 — autouse

Agreed it's a footgun even though no current test forgets it. Promoted `overlay_isolation` to `autouse=True` in `tests/conftest.py` (project-wide rather than scoped — the teardown is a no-op for non-overlay tests, so the scoping complexity didn't seem worth it). Also fixed the now-incorrect comment in `presets.py` that pointed at `tests/unit/llm/conftest.py` (the fixture lives at `tests/conftest.py`).

---

Side effect of the High-2 fix: my resolver-precedence tests in `test_preset_overlay_subprocess.py` and `test_preset_overlay_worker_propagation.py` previously used placeholder paths like `/from/queue.yaml` — the new eager validation correctly rejects those, so they now write real (minimal but valid) overlay YAMLs to `tmp_path`. The eager fix exercises itself through the existing precedence tests, which I think is a nice property to have.

**Test impact:** 55 overlay tests (+4 from this round), full offline regression at **1903 / 1 skipped** (was 1899). No verdict drift on the golden grading suite.

balazs-toloka (approved):

PR #81, opened 2026-06-23
feat(devcontainer): add Dev Container config for reproducible dev env

azorej:
Add a `.devcontainer/` setup (Dockerfile, devcontainer.json, and setup/init scripts) for a reproducible local/Codespaces development environment running as the `codespaces` user with docker-in-docker and cached cache/local volume mounts.

Supporting changes:
- init_git_lfs.sh: use `git lfs install --force` so a pre-existing pre-push hook left in the persisted .git across container rebuilds is overwritten instead of aborting with exit code 2.
- .gitignore: ignore the generated devcontainer-lock.json.

### DIFF
--- a/.devcontainer/.zshrc
+++ b/.devcontainer/.zshrc
@@ -1,0 +1,88 @@
+export PATH="$HOME/.local/bin:$PATH"
+
+# Path to your oh-my-zsh installation.
+export ZSH="$HOME/.oh-my-zsh"
+
+# See https://github.com/ohmyzsh/ohmyzsh/wiki/Themes
+ZSH_THEME="af-magic"
+
+# Case-sensitive completion must be off. _ and - will be interchangeable.
+HYPHEN_INSENSITIVE="true"
+
+zstyle ':omz:update' mode disabled  # disable automatic updates
+
+# Uncomment the following line if pasting URLs and other text is messed up.
+# DISABLE_MAGIC_FUNCTIONS="true"
+
+# Uncomment the following line to disable colors in ls.
+# DISABLE_LS_COLORS="true"
+
+# Uncomment the following line to disable auto-setting terminal title.
+# DISABLE_AUTO_TITLE="true"
+
+# Uncomment the following line to enable command auto-correction.
+ENABLE_CORRECTION="true"
+
+# Uncomment the following line to display red dots whilst waiting for completion.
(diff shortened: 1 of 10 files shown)


CiroGamboa (commented):
Two suggestions on the devcontainer setup — neither blocking.

- **`scripts/setup/init_git_lfs.sh:30`** — drop `--force` from the non-local fallback (`|| git lfs install`, not `|| git lfs install --force`). The PR's stated bug ("pre-push hook left in the persisted .git across container rebuilds") is per-repo and fully handled by `--local --force`. The fallback path runs only when the local form fails (e.g. not inside a git repo), and with `--force` it now silently overwrites any user-level LFS hook template — wider blast radius than the PR scope describes. AGENTS.md rule #1 ("Surface failures explicitly — do not add fallbacks that hide errors") points the same way. The script's own header advertises `Usage: scripts/init_git_lfs.sh`, so a host developer reading it standalone would now get the destructive behavior even though no such caller exists today.

- **`.devcontainer/Dockerfile:84`** — consider adding a sentence to the uv-pin comment noting that `/home/codespaces/.local` is overlaid by the `${localWorkspaceFolderBasename}-local` named volume, so bumping the pin in the Dockerfile requires removing that volume to take effect on existing containers. The pin comment is already thoughtful about CI/required-version alignment — this just closes the loop on the local-rebuild case. No code change needed.

CiroGamboa (dismissed):

azorej:
> ripts/setup/init_git_lfs.sh:30 — drop --force from the non-local fallback (|| git lfs install, not || git lfs install --force). The PR's stated bug ("pre-push hook left in the persisted .git across container rebuilds") is per-repo and fully handled by --local --force. The fallback path runs only when the local form fails (e.g. not inside a git repo), and with --force it now silently overwrites any user-level LFS hook template — wider blast radius than the PR scope describes. AGENTS.md rule https://github.com/Toloka/tolokaforge/pull/1 ("Surface failures explicitly — do not add fallbacks that hide errors") points the same way. The script's own header advertises Usage: scripts/init_git_lfs.sh, so a host developer reading it standalone would now get the destructive behavior even though no such caller exists today.

I would like to keep `--force` because I want to keep the script idempotent

CiroGamboa (approved):

PR #102, opened 2026-06-26
fix(grading): faithful judge KB search — judge reads the same KB the agent did (#95)

### LINKED ISSUE #95: Rubric judge search_kb unreachable when agent uses a non-search_kb KB tool (rag-service never provisioned)
## Summary

The rubric judge's `search_kb` tool is **offered to the judge but cannot reach a RAG backend** for any task whose agent reads its knowledge base through a *domain/plugin* search tool (e.g. a TypeSense-backed `search_policy` connector) rather than the native `search_kb` **agent** tool. In that configuration the orchestrator never provisions `rag-service`, so the judge's `search_kb` calls fail at runtime and the judge silently falls back to DB-only inspection — it never actually reads the policy documents it is being asked to grade against.

This affects rubric grading correctness: a criterion like *"the response complies with documented policy X"* is graded without the judge ever reading policy X.

## Observed behavior

Docker runtime, `0.3.0`. A task with a structured `llm_judge` rubric whose criteria reference KB-documented policies. TypeSense is started (agent-side `search_policy` works), the agent successfully searches the KB, the trial completes, and grading runs. The judge's own transcript (`judge_trajectory.yaml`):

```
search_kb {query: "service credit eligibility conditions ..."}
  -> Error: Search failed: [Errno -2] Name or service not known   # DNS: rag-service unresolved
get_db_state {} ... query_db ...                                   # falls back to DB
submit_report { ... }                                              # grades anyway
```

`judge_status=completed`, but the policy docs were never read.

## Root cause

There are **two independent KB subsystems**, and only one is provisioned:

1. **Agent search** goes through the plugin's own tool (e.g. a TypeSense `search_policy` connector) → the TypeSense container the orchestrator starts when the run config has a `typesense:` block. ✅ works.
2. **Judge `search_kb`** goes through `RAGServiceClient` → **`rag-service:8001`** (`runner/rag_client.py`, default `http://rag-service:8001`; `RAG_SERVICE_URL` in `runner/__main__.py`). This is a *different* service with a *different* API.

The judge is given `search_kb` whenever the runner service has a `rag_client` at all — `rag_url = self.rag_client.base_url if self.rag_client is not None else None` (`runner/service.py`), and the runner always builds a `RAGServiceClient` from `RAG_SERVICE_URL` (default points at `rag-service`). So `search_kb` is **always advertised** to the judge, even when no `rag-service` is running.

But the orchestrator only starts `rag-service` (the `full_stack`) when **`_tasks_need_full_stack()`** is true, which checks only the **agent** tool list and `initial_state` (`orchestrator.py`):

```python
_FULL_STACK_TOOL_NAMES = frozenset({"browser", "mobile", "search_kb"})
# full_stack iff an *agent* enabled tool ∈ _FULL_STACK_TOOL_NAMES, or initial_state.rag/mock_web
```

It never considers that the **grading judge** needs RAG. A task whose agent searches via a domain connector (not the native `search_kb` agent tool) and has no `initial_state.rag` therefore runs on the core stack (db-service + runner [+ typesense]) with **no `rag-service`** — and the judge's `search_kb` resolves to a non-existent host. (`rag-service` is also an *extended* image, not built by `docker build --core`.)

## Impact

- Rubric judges that should verify against a KB grade **without reading it** — a silent correctness gap, masked because the judge still returns `completed` (it answers from DB state + criterion text + model priors).
- Especially affects plugin/conversion adapters whose agents use a TypeSense-style `search_policy` connector instead of the native `search_kb` agent tool.

## Proposed fixes (any one would help)

1. **Don't advertise `search_kb` to the judge unless a reachable RAG backend exists** — fail loud (or omit the tool) instead of handing the judge a tool that always errors. At minimum, surface the failed `search_kb` in `judge_status`/reasons rather than letting the judge silently fall back.
2. **Route the judge's `search_kb` to the same KB the agent used** when the task's search is TypeSense-backed (share the TypeSense the orchestrator already starts), instead of a separate `rag-service`.
3. **Make full-stack/RAG provisioning grading-aware** — start + index `rag-service` when the task's `grading.llm_judge` is present and the task ships a document corpus, not only when an *agent* tool needs it.

## Environment

- tolokaforge `0.3.0`, `runtime: docker`.
- Stack started for the repro: `typesense` + core (`db-service`, `runner`); `rag-service` absent.
- Surfaced while wiring the 0.3.0 structured rubric judge into external (non-published) conversion adapters.

azorej:
## Summary

Fixes [#95](https://github.com/Toloka/tolokaforge/issues/95): the structured rubric LLM-judge was graded **without ever reading the knowledge base it grades against**, silently. The judge was always offered a `search_kb` tool that POSTed to a non-existent host on the core stack (DNS failure), then fell back to DB-only grading and still returned `completed`.

This PR makes the judge's KB capability **exactly mirror the agent's**: same backend, same per-trial index, offered iff the agent actually had a KB tool — or none. It also makes the previously-silent state visible in the grade output.

## What changed (5 stages, one commit each)

1. **Honest `RAG_SERVICE_URL` env** — the runner container had `RAG_SERVICE_URL` baked into the Dockerfile + hardcoded in the core stack, and `__main__.get_config()` fell back to `localhost:8001`. So on the core stack (no `rag-service`) the runner always built a RAG client and the judge was always handed an unreachable `search_kb`. Now the env is present **iff a rag-service actually runs** (only `full_stack` injects it); honest absence ⇒ no RAG client ⇒ no phantom tool.
2. **`KnowledgeSearch` contract + per-trial rag fidelity** — a tolokaforge-owned, backend-neutral `KnowledgeSearch` Protocol (`core/grading/kb_search.py`); `RagServiceKnowledgeSearch` bound to the trial's `rag_client` + `trial_id` hits the **per-trial** `/trials/{trial_id}/search` (the agent's index), fixing the judge's old global-`/search` bug. The judge is rewired off `rag_url` onto the resolved backend.
3. **Faithful gating** — the judge gets `search_kb` iff the agent's reconstructed tools actually include a rag `search_kb` (`RAGSearchToolWrapper`), bound to the same client+trial_id ⇒ identical retrieval by construction. (Replaces a gate keyed on `search_config.enabled`, which is decoupled from the agent's tool.)
4. **TypeSense via reusing the agent's `search_policy`** — the headline case (agent KB = closed mcp_core TypeSense connector). Rather than reimplement the query against an mcp_core client we can't import or test, the judge **reuses the agent's own already-reconstructed read-only `search_policy` tool** through a generic, mcp_core-free `DelegatingReadTool` (async-bridged via the runner's event loop). Maximally faithful — same tool/query/backend/ranking.
5. **Observability + cleanup** — the judge's grade `reasons` now ends with `Judge KB: search_kb | search_policy | none offered`, making "graded with/without KB" visible. Per-trial KB is cleared at `CleanupTrial`.

## Verification

- Full suite green: **1839 unit + 218 canonical**, lint/format clean.
- **Real end-to-end run** (OpenRouter, ~$0.01) of the `summarize_notes_rubric` example on a rebuilt runner image: judge completes with per-criterion results, emits `Judge KB: none offered`, and the judge transcript shows **no** `search_kb` call — i.e. the judge is no longer handed a phantom KB tool on the core stack.

## Scope / known limitations (pre-existing, not fixed here)

- The **TypeSense `search_policy` path runs only in deployed runner containers** — `mcp_core` is not importable in-repo, so that path is covered by unit tests with a fake tool, not an in-repo integration run.
- Two **pre-existing** issues surfaced and documented (candidates for separate issues): the native+rag `search_kb` path is non-functional (`native.py` hardcodes `search.enabled=False` ⇒ per-trial index never populated), and the orchestrator's host-side global `/index` POST hits a non-existent route. The judge faithfully mirrors the agent in both cases.
- The mcp_core TypeSense-registry handle is not deregistered at cleanup (bounded, idempotent leak) — documented in-code rather than papered over with a blind call to an unconfirmed API.


azorej:
**Pre-existing bugs surfaced during this work** (not introduced here; filed as follow-ups):

- #107 — rag-service `search_kb` is non-functional for native tasks (`native.py` hardcodes `search.enabled=False`, so the per-trial index is never populated).
- #108 — the orchestrator's host-side global `/index` POST hits a non-existent rag-service route.

This PR's judge KB resolver is unaffected: it mirrors the agent's tool by construction, so the judge faithfully retrieves exactly what the agent does (including "nothing" when the above paths are dead). These two are the reason a *working* rag `search_kb` could not be demonstrated end-to-end for native tasks in-repo.

CiroGamboa (approved):

PR #134, opened 2026-07-02
refactor(orchestrator): collapse injection kwargs into OrchestratorDeps

### LINKED ISSUE #114: refactor(orchestrator): collapse the four injection kwargs into a typed OrchestratorDeps dataclass
## Context

After the artifact-writer injection PR lands, \`Orchestrator.__init__\` takes **four sibling injection kwargs** for the engine's writer/executor seams:

\`\`\`python
def __init__(
    self,
    config: RunConfig,
    resume: bool = False,
    verbose: bool = False,
    strict: bool = False,
    run_aggregate_writer: RunAggregateWriter | None = None,
    runtime_backend: RuntimeBackend | None = None,
    conductor_factory: Callable[..., Conductor] | None = None,
    artifact_writer: TrialArtifactWriter | None = None,
):
\`\`\`

Three smells the next seam will make worse:

1. **Constructor-explosion territory.** Every new pluggable seam adds another kwarg.
2. **\"Injected-or-default\" ternary duplicated** twice today (\`_artifact_writer\`, \`_run_aggregate_writer\`). The pattern is invisible to readers and copy-pasted as more seams arrive.
3. **Naming inconsistency across stored attributes.** Three of four follow \`self._<name>\` (\`_artifact_writer\`, \`_run_aggregate_writer\`, \`_conductor_factory\`); the runtime backend uses \`self._injected_runtime_backend\`. A future deps-object refactor has to reconcile.
4. **\`Callable[..., Conductor]\` factory signature** is untyped \`...\` because a typed signature would be unwieldy — the symptom of dependencies that should be packed into a context object.

\`_build_conductor\` already passes 10 keyword args through to the factory in lockstep, and will keep growing with any new orchestrator dep.

## Ask

Introduce a frozen \`OrchestratorDeps\` dataclass holding the four pluggable seams:

\`\`\`python
@dataclass(frozen=True)
class OrchestratorDeps:
    artifact_writer: TrialArtifactWriter = field(default_factory=FileArtifactWriter)
    run_aggregate_writer: RunAggregateWriter = field(default_factory=FileAggregateWriter)
    runtime_backend: RuntimeBackend | None = None
    conductor_factory: Callable[[ConductorContext], Conductor] | None = None
\`\`\`

\`Orchestrator.__init__\` accepts a single \`deps: OrchestratorDeps = OrchestratorDeps()\` parameter; existing callers either migrate to the new shape or use the default (zero-arg) construction.

In parallel, define a \`ConductorContext\` dataclass to give the factory signature a typed shape — collapsing \`_build_conductor\`'s 10-kwarg list to a single typed argument.

## Acceptance

- \`Orchestrator(config, deps=OrchestratorDeps(artifact_writer=...))\` works.
- Default-path callers (\`Orchestrator(config)\`) still construct \`FileArtifactWriter\` + \`FileAggregateWriter\` defaults.
- \`Callable[..., Conductor]\` is replaced by \`Callable[[ConductorContext], Conductor]\` (or similar typed shape).
- The four existing TestXxxInjection classes either fold into a parametrized fixture or migrate to the new shape.
- Canonical + unit suites green.
- Native smoke green.

## Risk

Medium. Breaking-change in the \`Orchestrator(...)\` and \`conductor_factory\` API surfaces. Worth a small ADR (candidate ADR-0009) before the implementation PR so the public seam stabilises before more seams pile on.

## Why now

Surfaced in PR #112's review: the artifact-writer injection adds the fourth kwarg in the same pattern. Better to refactor before a fifth lands and the migration cost doubles.

CiroGamboa:
## TL;DR

The four independent `Orchestrator` injection kwargs (`artifact_writer`, `run_aggregate_writer`, `runtime_backend`, `conductor_factory`) are packed into a single frozen `OrchestratorDeps` dataclass, and the `conductor_factory` signature is upgraded from `Callable[..., Conductor]` to `Callable[[ConductorContext], Conductor]` — where `ConductorContext` is a new frozen dataclass carrying the eleven per-run arguments the factory used to receive as `**kwargs`.

## What ships

- New `OrchestratorDeps` frozen dataclass in `tolokaforge/core/orchestrator.py` — packs the four pluggable seams with sensible defaults (`FileArtifactWriter`, `FileAggregateWriter`, `None`, `None`).
- New `ConductorContext` frozen dataclass in `tolokaforge/core/conductor.py` — the typed argument for any conductor factory.
- `Orchestrator.__init__` accepts a single `deps: OrchestratorDeps | None = None` in place of the four kwargs; the four old kwargs are removed (there is no shim layer).
- `Orchestrator._build_conductor` now constructs a `ConductorContext` and passes it to the factory (or unpacks it for the default `InProcessConductor` path).
- Stored-attribute naming stays as `_injected_runtime_backend` (matching Phase 4's convention that's already used at every read site in `run()` / `run_worker()`).
- The four `TestXxxInjection` classes in `tests/unit/test_orchestrator_logic.py` migrate to the new shape.

## Why

The Orchestrator constructor was drifting into the constructor-explosion shape every DI-heavy class eventually hits: each new pluggable seam (four so far) added another sibling kwarg, and the \"injected-or-default\" ternary was copy-pasted at each site. `_build_conductor` had grown a matching 10-kwarg pass-through, and the factory type was stuck at `Callable[..., Conductor]` because a typed signature over ten independent kwargs would be unusable — the exact shape ADR-0011's data-declaration convention flags as \"pack these into a value object.\"

Packing the injection seams into `OrchestratorDeps` gives us one place to grow (add a field, add a `field(default_factory=...)`, done — no constructor edit, no ternary, no attribute-naming choice). Packing the factory's arguments into `ConductorContext` upgrades the factory contract from \"any callable\" to a real type, so a future seam can extend the context without silently changing every factory implementation's `**kwargs` reach. Both dataclasses are frozen so callers can't smuggle mutation across the seam.

No new ADR is authored — this applies the seam-definition + data-declaration convention codified in ADR-0011 (PR #132) rather than introducing a new one. If reviewers want an ADR to pin the OrchestratorDeps shape specifically, happy to split that out as a follow-up.

## Changes

- `tolokaforge/core/conductor.py` — add `ConductorContext` frozen dataclass; export from `__all__`.
- `tolokaforge/core/orchestrator.py` — add `OrchestratorDeps`; rewrite `Orchestrator.__init__` around `deps`; refactor `_build_conductor` to build and pass a `ConductorContext`. Rebase note: kept the `_injected_runtime_backend` attribute name (matches every read site added in Phase 4) instead of the original PR's rename toward `_runtime_backend`.
- `tests/unit/test_orchestrator_logic.py` — migrate the four `TestXxxInjection` classes (`TestRuntimeBackendInjection`, `TestArtifactWriterInjection`, `TestConductorInjection`, and the inline `run_aggregate_writer` test) to `deps=OrchestratorDeps(...)`; retype factory callbacks from `**kwargs` to `ctx: ConductorContext`; the factory-arguments assertion now checks `ConductorContext` field values rather than dict-key membership.

`Orchestrator(config, resume=..., verbose=..., strict=...)` is unchanged — the three CLI call sites in `tolokaforge/cli/main.py` and the bare-config uses in tests keep working without modification.

## Verification

- \`make lint\` — green
- \`make format-check\` — green on the three modified files (pre-existing drift in ~10 unrelated files is per AGENTS.md gotchas #3 / #4, unchanged by this PR)
- \`uv run pytest tests/unit tests/canonical -q\` — **2197 passed, 1 skipped in 32.91s**

The 93 tests in \`test_orchestrator_logic.py\` — including all four injection-suite classes — pass on the new shape. The canonical \`test_conductor_contract.py\` still constructs \`InProcessConductor\` directly with its unchanged kwargs API (this PR only re-types the factory contract, not the concrete class's constructor).

## Refs

Closes #114

### DIFF
--- a/tests/unit/test_orchestrator_logic.py
+++ b/tests/unit/test_orchestrator_logic.py
@@ -654,14 +654,14 @@ class TestGenerateReports:
         assert "failures" in fa
 
     def test_run_aggregate_writer_kwarg_swaps_writer(self, tmp_path: Path) -> None:
-        """The ``run_aggregate_writer`` kwarg accepts any
+        """``OrchestratorDeps.run_aggregate_writer`` accepts any
         :class:`RunAggregateWriter` impl and routes the four aggregates
         through it instead of the default disk-backed writer."""
-        from tolokaforge.core.orchestrator import Orchestrator
+        from tolokaforge.core.orchestrator import Orchestrator, OrchestratorDeps
         from tolokaforge.core.output.aggregates import InMemoryAggregateWriter
 
         writer = InMemoryAggregateWriter()
-        orch = Orchestrator(_make_run_config(), run_aggregate_writer=writer)
+        orch = Orchestrator(_make_run_config(), deps=OrchestratorDeps(run_aggregate_writer=writer))
         orch.tasks = [_make_task_config("T1")]
         orch.results = [_make_trajectory("T1", 0, score=1.0, binary_pass=True)]
 
@@ -682,11 +682,11 @@ class TestGenerateReports:
     def test_empty_results_does_not_invoke_writer(self, tmp_path: Path) -> None:
         """The empty-results early-return guard runs before the writer call,
         so no aggregates are recorded when the run produced zero trials."""
-        from tolokaforge.core.orchestrator import Orchestrator
+        from tolokaforge.core.orchestrator import Orchestrator, OrchestratorDeps
         from tolokaforge.core.output.aggregates import InMemoryAggregateWriter
 
         writer = InMemoryAggregateWriter()
-        orch = Orchestrator(_make_run_config(), run_aggregate_writer=writer)
+        orch = Orchestrator(_make_run_config(), deps=OrchestratorDeps(run_aggregate_writer=writer))
         orch.results = []
 
         orch._generate_reports(tmp_path)
@@ -1230,8 +1230,8 @@ class TestJudgeModelGate:
         if a refactor moved the gate below the scheduling loop, the conductor's
         ``call_log.runs`` would be non-empty and this test would fail.
         """
-        from tolokaforge.core.conductor import InMemoryConductor
-        from tolokaforge.core.orchestrator import Orchestrator
+        from tolokaforge.core.conductor import ConductorContext, InMemoryConductor
+        from tolokaforge.core.orchestrator import Orchestrator, OrchestratorDeps
 
         config = _make_run_config()  # agent only, no judge model
 
@@ -1239,10 +1239,10 @@ class TestJudgeModelGate:
         # we can assert on its call_log after the run_worker raises.
         recording_conductor = InMemoryConductor()
 
-        def conductor_factory(**_kwargs):
+        def conductor_factory(_ctx: ConductorContext) -> InMemoryConductor:
             return recording_conductor
 
-        orch = Orchestrator(config, conductor_factory=conductor_factory)
+        orch = Orchestrator(config, deps=OrchestratorDeps(conductor_factory=conductor_factory))
         orch.tasks = [_make_task_config("TASK-needs-judge")]
         adapter = MagicMock()
         adapter.to_task_description.side_effect = lambda tid: _task_description_with_judge(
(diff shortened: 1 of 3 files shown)


CiroGamboa:
Ready for review @CiroGamboa

CI green (lint / test-smoke / CodeQL / hygiene-review). Automated Claude review posted clean — verdict "everything checks out." Two non-blocking nits called out and declined for scope discipline:

1. **`**vars(ctx)` in the default `InProcessConductor` path**: the reviewer themselves note the explicit spelling is arguably clearer and "fine as-is" — leaving explicit.
2. **`ConductorFactory` type alias**: only two use sites, marginal deduplication vs another exported symbol. Skipping.

Happy to fold either if you disagree.

CiroGamboa:
Rebased against post-Phase-4 `main` (which introduced env_manifest-aware branches in the same `_build_conductor` / `__init__` regions). Conflict resolved by keeping the `OrchestratorDeps` DI change from this PR while re-adding the Phase 4 additions (`_extract_run_env_manifest`, `task_stack_mode` branch, distributed-worker env_manifest guard). The `ConductorContext` dataclass gained a `trial_grader` field to match the current `InProcessConductor.__init__`, and the `_runtime_backend` attribute was renamed to `_injected_runtime_backend` to align with the Phase-4-era naming used everywhere else in the orchestrator. Two test assertions updated to match.

Verification against the rebased tip:

- Full unit + canonical suite: 2430 pass locally.
- Empirical regression across 5 public examples (`native_shared_domain`, `coding`, `multi_service`, `multi_service_advanced`, `multi_service_postgres`) plus 2 representative recent internal task packs — every one produced a byte-identical `avg_score_micro` on `main` vs. this branch. Deterministic (`temperature=0.0`), so identical is the pass invariant.

Ready for review.

PR #136, opened 2026-07-02
chore(python): single-source the Python runtime version from .python-version

CiroGamboa:
## TL;DR

Move the pinned Python minor version to a single source (`.python-version`) and thread it through every Dockerfile as a `PYTHON_VERSION` build arg. CI workflows now source the version from the same file. Bumping Python is now a one-line change instead of a ~14-spot audit.

## What ships

- **`builder.py`**: reads `.python-version` once at module load; passes `PYTHON_VERSION` in the `build_args` of every `IMAGE_DEFINITIONS` entry and of the wheel-dependent factories `_runner_definition()` / `_rag_definition()`.
- **Runtime Dockerfiles** (4 live + 2 orphan): `runner.Dockerfile`, `db_service.Dockerfile`, `mock_web.Dockerfile`, `rag.Dockerfile`, `json_db.Dockerfile`, `orchestrator.Dockerfile` — each declares `ARG PYTHON_VERSION=3.12` and uses `FROM python:${PYTHON_VERSION}-slim`. The default matches `.python-version` so a bare `docker build` (without the harness passing the arg) still produces the same image.
- **CI workflows** (5 files): `ci.yml`, `release-gate.yml`, `publish-tolokaforge.yml`, `release.yml`, `publish-adapter-terminal-bench.yml` all switch from `uv python install 3.12` to `uv python install "$(cat .python-version)"`.
- **Test update**: `tests/unit/test_docker_build_context.py` — the three tests that pinned `build_args` (`test_build_image_respects_context_files`, `test_build_image_non_force_path_uses_isolated_context`, and the former `test_build_image_passes_no_build_args_when_definition_has_none`) now assert on the new `PYTHON_VERSION` build arg alongside `WHEEL_FILENAME`. The third test's premise flipped (db-service used to have no build_args and now has one), so it's renamed and re-scoped to prove the single-source contract at build-time.

Not touched (intentional): `pyproject.toml` `requires-python = ">=3.10"` and ruff/black `target-version = py310` — those are the OSS-consumer minimum-supported floor and are conceptually separate from the pinned runtime version. Bumping the pin doesn't tighten what consumers must run.

## Why

Three separate Python-version declarations had drifted apart:

- Dev / devcontainer / CI: **3.12** (`.python-version`, `devcontainer.json`, five workflow files)
- `runner.Dockerfile` + `db_service.Dockerfile`: **3.11**
- `mock_web.Dockerfile` + `rag.Dockerfile`: **3.10**
- Orphan Dockerfiles: **3.10**

The engine's Docker path was **running the runner on 3.11 while dev/CI test on 3.12**. That's a silent latent-behavior gap and directly undercuts the "the runner runs the byte-identical engine" contract the architecture depends on. A future Python upgrade would also require finding and editing all ~14 spots — an audit that would inevitably miss some.

The fix has two shapes:

1. **Align now** — every live runtime image runs the version dev + CI test on (3.12). Same interpreter minor, same behavior guarantees, no drift.
2. **Single-source future** — one file (`.python-version`) is the pin. `builder.py` reads it and passes it as a build arg; CI workflows read it directly. A future bump touches one place; every derived surface follows automatically.

The `ARG PYTHON_VERSION=3.12` default on each Dockerfile keeps them independently reproducible — a hand-rolled `docker build` still produces the current pin without needing the harness. This is deliberate: the Dockerfiles remain self-contained artifacts, and the harness just tells them what version to use when it disagrees with the default (it currently doesn't).

## Changes

- `tolokaforge/docker/builder.py` — add `_pinned_python_version()` reader, module-level `PYTHON_VERSION` and `_PYTHON_BUILD_ARGS`; merge into every `IMAGE_DEFINITIONS` entry and both dynamic factories.
- Six Dockerfiles under `tolokaforge/docker/dockerfiles/` — add `ARG PYTHON_VERSION=3.12` before `FROM python:${PYTHON_VERSION}-slim`.
- Five workflows under `.github/workflows/` — source the version from `.python-version` in each `uv python install` step; rename the step to make the source visible.
- `tests/unit/test_docker_build_context.py` — update the three affected assertions to include `PYTHON_VERSION`.

## Verification

- `make lint` — green
- `uv run pytest tests/unit tests/canonical -q -m "unit or canonical"` — **2197 passed, 1 skipped in 26.12s**
- `tests/unit/test_docker_build_context.py` — all 9 tests green with the new assertions

To upgrade Python in the future: edit `.python-version`. That's it. dev, CI, devcontainer venv, and every runtime image follow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,8 +53,8 @@ jobs:
         enable-cache: true
         cache-dependency-glob: "uv.lock"
 
-    - name: Set up Python 3.12
-      run: uv python install 3.12
+    - name: Set up Python (pinned in .python-version)
+      run: uv python install "$(cat .python-version)"
 
 
     - name: Install dependencies
@@ -91,8 +91,8 @@ jobs:
         enable-cache: true
         cache-dependency-glob: "uv.lock"
 
-    - name: Set up Python 3.12
-      run: uv python install 3.12
+    - name: Set up Python (pinned in .python-version)
+      run: uv python install "$(cat .python-version)"
 
 
     - name: Install dependencies
@@ -176,8 +176,8 @@ jobs:
         enable-cache: true
         cache-dependency-glob: "uv.lock"
 
-    - name: Set up Python 3.12
-      run: uv python install 3.12
+    - name: Set up Python (pinned in .python-version)
+      run: uv python install "$(cat .python-version)"
 
 
     - name: Install dependencies
@@ -279,8 +279,8 @@ jobs:
         enable-cache: true
         cache-dependency-glob: "uv.lock"
 
-    - name: Set up Python 3.12
-      run: uv python install 3.12
+    - name: Set up Python (pinned in .python-version)
+      run: uv python install "$(cat .python-version)"
 
 
     - name: Install dependencies
--- a/.github/workflows/publish-adapter-terminal-bench.yml
+++ b/.github/workflows/publish-adapter-terminal-bench.yml
@@ -34,7 +34,7 @@ jobs:
         uses: astral-sh/setup-uv@v7
 
       - name: Install Python
-        run: uv python install 3.12
+        run: uv python install "$(cat .python-version)"
 
       - name: Build package
         run: uv build --package tolokaforge-adapter-terminal-bench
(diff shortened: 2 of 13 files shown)


CiroGamboa:
Consolidated with the sibling CI/packaging PR into a single review path — see the new PR for the same content plus a shared test-plan. Closing to keep one queue slot per topic.

Issue #37, opened 2026-06-03
`tolokaforge adapter convert` crashes: `No module named 'tolokaforge.adapters.bundle_writer'`


azorej:
---

## Summary

The `tolokaforge adapter convert` CLI command is broken in the published
package. Its implementation imports `tolokaforge.adapters.bundle_writer`, but
that module does not exist in the published `tolokaforge` package — so the
command fails immediately for **every** adapter, before doing any work.

## Environment

- `tolokaforge` 0.2.1, installed from public PyPI
- Python 3.12
- A registered conversion-capable adapter plugin (e.g. `tlk_mcp_core`)

## Steps to reproduce

```bash
pip install tolokaforge==0.2.1
# with any conversion-capable adapter installed and a valid tasks glob:
tolokaforge adapter convert \
  --name tlk_mcp_core \
  --tasks-glob "/path/to/domains/<domain>/testcases/*.json" \
  --output /tmp/out \
  --adapter-params '{"tools_library": "...", "mcp_core_path": "..."}'
```

## Actual behavior

```
File ".../tolokaforge/cli/adapter_commands.py", line 47, in convert
    from tolokaforge.adapters.bundle_writer import write_bundle, write_domain_bundle
ModuleNotFoundError: No module named 'tolokaforge.adapters.bundle_writer'
```

The command never runs — it raises at the import inside `convert()`.

## Expected behavior

`tolokaforge adapter convert` discovers tasks and writes native task bundles
(per-task `task.yaml` / `grading.yaml` / `fixtures/` plus the shared `_domain/`
bundle) to `--output`.

## Root cause

`tolokaforge/cli/adapter_commands.py` (the `convert` command) still imports the
bundle writer from core:

```python
# tolokaforge/cli/adapter_commands.py  (inside convert)
from tolokaforge.adapters import get_adapter
from tolokaforge.adapters.bundle_writer import write_bundle, write_domain_bundle
```

But the published package's `tolokaforge/adapters/` directory contains only
`base.py`, `native.py`, `_task_loader.py`, `__init__.py` — there is **no
`bundle_writer.py`**. It appears `bundle_writer` was moved out of the core
package (it now ships inside the `frozen_mcp_core` adapter plugin), and the CLI
import in core was not updated to match.

Verify:

```bash
python -c "import tolokaforge.adapters.bundle_writer"
# ModuleNotFoundError: No module named 'tolokaforge.adapters.bundle_writer'
```

## Impact

[...]

PR #137, opened 2026-07-02
test(adapter-convert): CI smoke pins wheel packaging + real CLI invocation

### LINKED ISSUE #49: CI smoke test for `tolokaforge adapter convert` (real entry-point adapter)
### Gap

The regression tests added in #48 (fixing #37) are in-process and monkeypatch `get_adapter`. They verify `write_bundle` and the `convert` flow logic, but they do **not** cover:

- whether `tolokaforge/adapters/bundle_writer.py` actually **ships in the built package**, or
- **real adapter discovery** via the `tolokaforge.adapters` entry-point group through the actual `tolokaforge` CLI binary.

The original #37 failure (`ModuleNotFoundError` for a module missing from the published wheel) would only be caught by a packaging/CLI-level test — exactly the layer we don't currently automate.

This was verified **manually** for #48 (built the wheel, installed into a clean venv with a throwaway entry-point-registered demo adapter, ran the real `tolokaforge adapter convert … --validate` → valid bundles + the `_domain/` hook output). This issue is to **automate** that.

### Proposed approach

- Add a tiny demo conversion adapter (in-tree fixture package) registered via `[project.entry-points."tolokaforge.adapters"]`.
- A test (subprocess against the installed console script, or at least real `get_adapter` discovery + the real CLI) that runs `tolokaforge adapter convert` end-to-end on sample source files and asserts: the native bundle is written (`task.yaml` / `grading.yaml` / `fixtures/`), `--validate` reports valid, and `write_shared_resources` is invoked.
- Optionally assert `tolokaforge/adapters/bundle_writer.py` is present in `uv build` output (packaging guard).
- Wire into CI (likely the existing smoke/full job).

### Acceptance

CI fails if `adapter convert` crashes on import, if `bundle_writer` isn't packaged, or if entry-point discovery + bundle output regress. **Would have caught #37.**

Related: #48 (the fix), #19 (Engine end-to-end smoke test in CI).

CiroGamboa:
## TL;DR

Add three canonical tests that catch the failure modes the existing in-process adapter-convert suite can't see: (1) a source module missing from the built wheel, (2) the CLI subprocess crashing on import, (3) `bundle_writer` symbols not resolving. These would have caught the GH #37 regression (module physically absent from the published wheel; in-process tests happily green).

## What ships

New file `tests/canonical/test_adapter_convert_packaging.py` with three tests:

- **`test_bundle_writer_ships_in_the_built_wheel`** — runs `uv build --wheel`, unzips the produced wheel, and asserts every module the CLI reaches at call time is physically present. Currently checks: `bundle_writer.py`, `base.py`, `native.py`, `adapters/__init__.py`, `adapter_commands.py`, `cli/main.py`.
- **`test_adapter_convert_cli_help_runs_via_subprocess`** — spawns `uv run tolokaforge adapter convert --help` in a real subprocess and asserts the help output includes the expected flags. Proves the installed console script wires, the click group resolves, and every module the CLI touches at import time loads under the real Python resolver.
- **`test_bundle_writer_module_imports_cleanly`** — in-process cheap guard that the `write_bundle` symbol is callable and the module lands in `sys.modules`. Complements the wheel-inspection test with a fast path.

Both `uv`-dependent tests skip loudly with `pytest.skip` when the `uv` CLI is not on PATH — no false-negative silent skips.

## Why

The existing regression suite around `tolokaforge adapter convert` (added in PR #48) monkey-patches `get_adapter` and drives the click runner in-process. That's good coverage for the *convert flow logic* but blind to two failure classes:

1. **Wheel-shipping bugs.** A file that exists in the source tree but doesn't make it into the built wheel — because of a `hatch.build.targets.wheel` misconfiguration or a `.gitignore`-adjacent exclusion — will pass every in-process test. The bug only shows up when a user `pip install`s the wheel and the missing module raises `ImportError`. This is exactly what GH #37 was.
2. **CLI-boundary import bugs.** In-process click testing bypasses the actual entry-point script (`tolokaforge = tolokaforge.cli.main:cli` in `pyproject.toml`). A regression in the console-script wiring, a circular import at CLI top-level, or an entry-point group name typo can survive the whole in-process suite.

The three added tests plug both holes: (1) inspects the wheel binary artifact, (2) invokes the CLI the way an end-user would (real subprocess), (3) verifies the module surface the CLI depends on. Together they'd catch any GH #37-shaped regression before it merges.

**No CI workflow change needed.** The existing `test-smoke` job in `ci.yml` runs `pytest -m "unit or canonical"`, so these canonical tests are picked up automatically on every PR. Wall-clock overhead is small: `uv build` takes ~2s in the sandbox measurement below; the subprocess help call ~1s.

## Changes

- `tests/canonical/test_adapter_convert_packaging.py` — new file, three tests, no production-code touch.

## Verification

- `make lint` — green
- `uv run pytest tests/canonical/test_adapter_convert_packaging.py -v` — **3 passed in 2.94s**
- `uv run pytest tests/unit tests/canonical -q -m "unit or canonical"` — **2200 passed, 1 skipped in 28.92s** (three new tests, no regressions)

## Refs

Closes #49

### DIFF
--- a/tests/canonical/test_adapter_convert_packaging.py
+++ b/tests/canonical/test_adapter_convert_packaging.py
@@ -1,0 +1,123 @@
+"""Packaging + CLI smoke for ``tolokaforge adapter convert``.
+
+Guards two failure modes the in-process tests can't catch:
+
+* A source file (e.g. ``tolokaforge/adapters/bundle_writer.py``) not
+  making it into the built wheel — the exact regression that shipped in
+  GH #37. In-process tests keep passing because the file exists on
+  disk; end-users installing from PyPI hit ``ImportError``.
+* The ``tolokaforge`` console script + entry-point-based adapter
+  discovery failing when invoked as a real subprocess (import ordering,
+  missing dependency at CLI import time, entry-point group name typo).
+
+Both tests run under the ``canonical`` marker so they participate in
+the existing CI smoke job without needing dedicated workflow wiring.
+"""
+
+from __future__ import annotations
+
+import shutil
+import subprocess
+import sys
+import zipfile
+from pathlib import Path
+
+import pytest
+
+pytestmark = pytest.mark.canonical
+
+
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+
+
+def _uv_available() -> bool:
+    """True if the ``uv`` CLI is on PATH — otherwise these tests skip loud."""
+    return shutil.which("uv") is not None
+
+
+@pytest.mark.skipif(not _uv_available(), reason="uv CLI not available")
+def test_bundle_writer_ships_in_the_built_wheel(tmp_path: Path) -> None:
+    """``tolokaforge.adapters.bundle_writer`` must be present in the wheel
+    hatchling produces.
+
+    A structural regression like PR #37 (module omitted from the built
+    wheel because of a hatch config gap) is invisible to any in-process
+    test — the source tree still has the file. The only way to catch it
+    is to build the wheel and look inside.
+    """
+    build_result = subprocess.run(
+        ["uv", "build", "--wheel", "--out-dir", str(tmp_path)],
+        cwd=_REPO_ROOT,
+        capture_output=True,
+        text=True,
+        timeout=180,
+        check=False,
+    )
+    assert build_result.returncode == 0, (
+        f"uv build failed (rc={build_result.returncode}):\n"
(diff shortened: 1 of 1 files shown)


CiroGamboa:
Consolidated with the sibling CI/packaging PR into a single review path — see the new PR for the same content plus a shared test-plan. Closing to keep one queue slot per topic.

Issue #86, opened 2026-06-25
RunAggregateWriter follow-up: type the four run-level aggregate payloads with Pydantic models


CiroGamboa:
The run-level data-plane seam landed in ADR-0005 (PR #85). The `RunAggregateWriter` Protocol's four payloads are still raw `dict[str, Any]`:

- `per_task_metrics: list[dict[str, Any]]`
- `aggregate: dict[str, Any]`
- `metadata_slices: dict[str, dict[str, Any]]`
- `failure_attribution: dict[str, Any]`

The shapes are wide and inconsistent across slices because they're produced by raw-dict-returning metric-calc functions (`calculate_task_metrics`, `calculate_aggregate_metrics`, `attribute_failure`, `summarize_failure_attributions`).

Typing them with Pydantic v2 models would:

- Lock the on-wire shape downstream readers depend on. Today `schema_version` on `aggregate.json` is the only declared invariant; everything else is implicit.
- Catch shape drift at PR time, not at consumer-dashboard-breakage time.
- Mirror the per-trial control↔trial seam's pattern (`TrialSpec` / `TrialResult` from ADR-0003).

Out of scope for PR #85 (would have been a separate breaking-change-risk surface). Belongs as its own ADR + PR.

## Proposed acceptance

- [ ] New `RunAggregate`, `PerTaskMetrics`, `MetadataSlices`, `FailureAttribution` Pydantic v2 models with `extra=\"forbid\"`.
- [ ] Metric-calc functions in `tolokaforge/core/metrics.py` and `tolokaforge/core/failure_attribution.py` return typed models, not raw dicts.
- [ ] `RunAggregateWriter` Protocol signatures updated to accept the typed models; both `FileAggregateWriter` and `InMemoryAggregateWriter` updated.
- [ ] Canonical contract + layout tests updated to assert on the typed surfaces.
- [ ] New ADR documenting the typed-payloads decision.
- [ ] Live smoke vs `main`: byte-shape identical output JSONs (the wire format is fixed; only the in-memory representation changes).

## Refs

- ADR-0005: [`docs/architecture/adr/0005-run-aggregate-writer-seam.md`](https://github.com/Toloka/tolokaforge/blob/main/docs/architecture/adr/0005-run-aggregate-writer-seam.md) — see the `Follow-ups` section.
- Source PR: #85.

CiroGamboa:
[...]

PR #149, opened 2026-07-02
refactor(output): typed models for run-level aggregate payloads (stage 1)

CiroGamboa:
## TL;DR

Introduce Pydantic v2 models for the four run-level aggregate JSON payloads (`per_task_metrics.json`, `aggregate.json`, `metadata_slices.json`, `failure_attribution.json`). **No production code path is touched, no behaviour changes.** The value is a canonical round-trip test that pins byte-identity between the model dumps and the current on-disk shape — a wire-format gate that makes the follow-on migration safe.

## Why (architectural context)

The engine has been steadily moving toward the pattern the runtime-architecture roadmap and ADRs 0003–0011 codify: **every data surface that crosses a boundary is a Pydantic model, every behavioural surface is a Protocol.** The per-trial control-plane already lives on `TrialSpec` / `TrialResult`; the run-level control-plane already has the `RunAggregateWriter` Protocol (ADR-0005). The one gap is what those seams carry — the four run-level aggregate payloads are still `dict[str, Any]`, produced by dict-returning metric-calc functions in `metrics.py` and `failure_attribution.py`, and consumed as raw dicts throughout the codebase.

Two consequences that are getting more expensive over time:

- **Downstream dashboards read fields we don't declare.** The only invariant the JSON currently pins is `schema_version: 1` on `aggregate.json`. Every other key is convention. Adding a new field is silent; renaming one is a silent break; drift between the calculator's shape and what the writer emits is caught by nothing except a broken dashboard.
- **The run-level seam is asymmetric with the per-trial seam.** ADR-0003 pins `TrialSpec` / `TrialResult` as typed models on the control↔trial edge; ADR-0005 pins the writer as a Protocol but leaves the payload untyped. That asymmetry means every future refactor at the run-level has to argue about the payload shape from scratch instead of pointing at a model.

Applying the **data-declaration pattern** (a value crossing a boundary is a `BaseModel` with `extra="forbid"`) to these four payloads closes the asymmetry, locks the on-disk shape, and makes shape drift a fail-loud PR-time signal instead of a silent-dashboard-break signal.

## Why this PR does *only* the models, not the migration

The full migration touches three surfaces that all have to move together: the calculator functions in `metrics.py` and `failure_attribution.py`, the writer Protocol + its two implementations in `output/aggregates.py`, and ~150 in-process consumers scattered across the codebase. Landing all three in one PR would ship a large blast-radius change with a hard-to-verify constraint (byte-identical JSON output). If any one field diverges — a `1.0` where the calculator emitted `1`, a key present where the calculator omitted it — downstream dashboards break silently at deploy time, not at review time.

Splitting the work so that **the wire-format gate lands first, in isolation** removes that risk: this PR proves the models round-trip byte-identical against the real calculator output; every subsequent PR that touches the calculators or the writer runs against the same gate. A migration that changes the on-disk shape breaks the round-trip test at the earliest possible moment.

## What ships

- **`tolokaforge/core/output/aggregate_models.py`** (new file) — seven models:
  - **`PerTaskMetrics`** — one row of `per_task_metrics.json`. Task identity + metadata + trial counts + `pass@k` / `pass_hat@k` + averages (score, latency, turns, tool calls, per-usage-field tokens) + cost + wall-time percentiles + per-API-call percentiles + stuck rate + judge-cost split (`judge_cost_usd`, `total_cost_incl_judge_usd`). ~37 fields.
  - **`AggregateMetrics`** — the shared body used both by `aggregate.json` (via `RunAggregate`) and by every slice value inside `metadata_slices.json`. The `success_rate_micro` / `_macro` and `avg_score_micro` / `_macro` fields are `Optional` so the same model round-trips both `weighted=True` and `weighted=False` output of `calculate_aggregate_metrics`. Includes the same judge-cost split (`judge_cost_usd`, `total_cost_incl_judge_usd`) aggregated across the tasks in the run/slice.
  - **`RunAggregate`** — inherits `AggregateMetrics` and adds the `schema_version` envelope the current writer stamps at the top level.
  - **`MetadataSlices`** — the four `by_*` dictionaries (`by_benchmark_type`, `by_complexity`, `by_tag`, `by_expected_failure_mode`), each `dict[str, AggregateMetrics]`.
  - **`FailureAttribution`** / **`FailureSummary`** / **`FailureRecord`** — the `failure_attribution.json` envelope, populated by `attribute_failure` + `summarize_failure_attributions`.
- **`tests/canonical/test_run_aggregate_models_snapshot.py`** (new file) — ten round-trip tests that invoke the production metric-calc functions on representative trajectories, feed the resulting dict through `Model.model_validate(...).model_dump(by_alias=True, mode="json", exclude_unset=True)`, and assert the dumped dict equals the source dict — same keys, same values, same types.

Nothing else is touched — the current writer Protocol still accepts and emits dicts; every producer function still returns dicts.

## Design decisions worth reading

- **`pass@k` / `pass_hat@k` aliases.** These wire keys aren't valid Python identifiers. Handled with `Field(default=None, alias="pass@1")` and dumped via `by_alias=True`. Six such aliases on `PerTaskMetrics`, six on `AggregateMetrics`. `populate_by_name=True` lets tests construct models by either the Python-identifier name or the wire alias for parity.
- **`extra="forbid"` on every model.** This is the whole point — a metric-calc function that grows an undeclared key fails validation at test time instead of shipping to `aggregate.json` and drifting from what the model says the shape is.
- **`exclude_unset=True` on the dump.** Matches the wire behaviour of the current writer path: keys the metric-calc functions did not set are absent from the JSON (e.g. `success_rate_macro` is absent on a weighted run). `exclude_none=True` would over-prune — real `None` values (`total_cost_usd` when no trial reported a cost) must round-trip as `None`, not drop.
- **`RunAggregate` inherits from `AggregateMetrics`** rather than composing. The shared body is exactly the union of every field `calculate_aggregate_metrics` produces; only the top-level artifact carries `schema_version`. Slice values inside `metadata_slices.json` are `AggregateMetrics` directly.
- **Judge-cost fields** (`judge_cost_usd`, `total_cost_incl_judge_usd` on both models) are `float | None = None`, matching `calculate_task_metrics` (`metrics.py:204`) and `calculate_aggregate_metrics` (`metrics.py:336-342`) — both compute `None` when neither an agent cost nor a judge cost is known, and `agent + judge` (with `0.0` for either missing side) otherwise. Added in the rebase against post-open `main`; the fields were introduced to the writer after this PR was first opened.
- **`FailureRecord.evidence: list[dict[str, Any]]`** stays untyped for now. Evidence entries are heterogeneous (`kind` acts as a discriminator — `tool_log` / `state_diff` / `termination_reason` / …); locking a union here would over-constrain the classifier and force every existing evidence emitter through a new type. Documented as a future refinement.

## Fit with the broader typing effort

- **ADR-0003 (`TrialSpec` / `TrialResult`)** pins the per-trial control-plane data. This PR pins the run-level analogue — same data-declaration pattern, applied to the seams it hadn't reached yet.
- **ADR-0005 (`RunAggregateWriter` Protocol)** established the run-level writer *behaviour* seam. Its follow-ups section explicitly flags the untyped payloads; this PR is the model half of that follow-up. The Protocol widening (to accept the models) is deliberately deferred to a follow-on PR so the wire-format gate lands first.
- **ADR-0011 (seam-definition + data-declaration conventions)** codifies exactly this shape: data crossing a JSON boundary is a Pydantic `BaseModel` with `extra="forbid"`; behavioural contracts are Protocols. The models in this PR are the direct application.

## Risk-management posture

The reason this PR is intentionally small is that the *full* refactor — model + calculator returns + writer signature + consumer migration — has real wire-format risk. A byte-order divergence between the current writer's JSON and a future model-based writer's JSON would silently break every downstream dashboard reading the aggregate files. Rather than ship all three surfaces in one PR and rely on human review to catch a shape drift, this PR lands *only* the models and *only* a snapshot test that pins the wire format. That test becomes a CI-level gate for any subsequent PR that migrates a producer, the writer, or a consumer: if the on-disk shape changes for any reason, the round-trip fails at PR time, not at deploy time.

## Follow-up work

- **Migrate the calculators.** `calculate_task_metrics`, `calculate_aggregate_metrics`, `attribute_failure`, `summarize_failure_attributions` return the models instead of dicts. The writer Protocol widens to accept them; `FileAggregateWriter.write_*` dumps via `model_dump(by_alias=True, mode="json", exclude_unset=True)`. The round-trip test from this PR is the wire-format check.
- **Migrate the consumers.** ~150 dict-key access sites (mostly tests, some orchestrator paths) switch to attribute access. Mechanical once the calculators return typed values.
- **Consolidate the duplicated per-usage-field lists.** `metrics.py` currently defines `_usage_fields` and `_agg_usage_fields` — identical six-field tuples across three call sites. Natural to fold when the calculator migration lands, since a drift between them is exactly the class of bug this PR guards against.

## Verification

- `make lint` — green
- `uv run pytest tests/unit tests/canonical -q -m "unit or canonical"` — **2216 passed, 1 skipped** (ten new tests from the byte-identity file, no regressions)

## Refs

Refs #86

### DIFF
--- a/tests/canonical/test_run_aggregate_models_snapshot.py
+++ b/tests/canonical/test_run_aggregate_models_snapshot.py
@@ -1,0 +1,308 @@
+"""Byte-identity round-trip for the run-level aggregate Pydantic models.
+
+Stage 1 of TECHDEL-407 — the models in
+:mod:`tolokaforge.core.output.aggregate_models` exist alongside the
+current dict-typed writer surface. This suite proves that the models
+faithfully round-trip every field the metric-calc functions produce
+today, so stages 2 and 3 (function-return migration + consumer
+migration) can rely on the on-wire shape being unchanged.
+
+Each test:
+
+* generates a real dict payload by invoking the production metric-calc
+  function on representative inputs;
+* passes it through the corresponding model
+  (``Model.model_validate(dict).model_dump(by_alias=True, mode="json")``);
+* asserts the dumped dict equals the original dict — same keys, same
+  values, same types.
+
+Anything the current code writes that the model can't round-trip fails
+this suite loudly.
+"""
+
+from __future__ import annotations
+
+from datetime import UTC, datetime
+
+import pytest
+
+from tolokaforge.core.failure_attribution import (
+    attribute_failure,
+    summarize_failure_attributions,
+)
+from tolokaforge.core.metrics import (
+    calculate_aggregate_metrics,
+    calculate_task_metrics,
+)
+from tolokaforge.core.models import (
+    Grade,
+    GradeComponents,
+    Metrics,
+    TerminationReason,
+    Trajectory,
+    TrialStatus,
+    Usage,
+)
+from tolokaforge.core.output.aggregate_models import (
+    AggregateMetrics,
+    FailureAttribution,
+    FailureRecord,
+    FailureSummary,
+    MetadataSlices,
+    PerTaskMetrics,
+    RunAggregate,
+)
+
+pytestmark = pytest.mark.canonical
+
+
+def _make_trajectory(
+    task_id: str = "task-1",
+    trial_index: int = 0,
+    *,
+    binary_pass: bool = True,
+    score: float = 1.0,
+    latency_s: float = 12.34,
+    turns: int = 8,
+    tool_calls: int = 5,
+    stuck_detected: bool = False,
+    cost_usd: float | None = 0.021,
+    status: TrialStatus = TrialStatus.COMPLETED,
+    termination_reason: TerminationReason | None = None,
+) -> Trajectory:
+    """Build a Trajectory populated on every field the aggregate path reads.
+
+    Every ``Usage`` field is nonzero so the round-trip catches drops in
+    the ``avg_*`` / ``total_*`` slots. ``binary_pass=False`` (with a
+    non-``COMPLETED`` status) drives the failure-attribution path.
+    """
+    now = datetime.now(tz=UTC)
+    return Trajectory(
+        task_id=task_id,
+        trial_index=trial_index,
+        start_ts=now,
+        end_ts=now,
+        status=status,
+        termination_reason=termination_reason,
+        messages=[],
+        metrics=Metrics(
+            latency_total_s=latency_s,
+            turns=turns,
+            tool_calls=tool_calls,
+            stuck_detected=stuck_detected,
+            cost_usd=cost_usd,
+            usage=Usage(
+                prompt_tokens=1200,
+                completion_tokens=340,
+                reasoning_tokens=80,
+                cached_tokens=200,
+                cache_creation_input_tokens=150,
+                cache_read_input_tokens=90,
+            ),
+        ),
+        grade=Grade(
+            binary_pass=binary_pass,
+            score=score,
+            components=GradeComponents(state_checks=score),
+        ),
+    )
+
+
+def _augment_task_metrics(
+    task_metrics: dict, *, task_id: str, benchmark_type: str = "airline"
+) -> dict:
+    """Mimic the orchestrator's per-row augmentation (see
+    ``orchestrator.py`` around the ``calculate_task_metrics`` call site)."""
+    task_metrics["task_id"] = task_id
+    task_metrics["benchmark_type"] = benchmark_type
+    task_metrics["complexity"] = "simple"
+    task_metrics["expected_failure_modes"] = ["timeout_or_resource"]
+    task_metrics["tags"] = ["domain:airline", "smoke"]
+    return task_metrics
+
+
+def _round_trip(model_cls, payload) -> None:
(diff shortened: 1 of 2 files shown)


CiroGamboa:
Rebased against current `main` and added one fix commit: the Pydantic models now include `judge_cost_usd` and `total_cost_incl_judge_usd` on both `PerTaskMetrics` and `AggregateMetrics`. Those fields were added to the actual writer output on `main` after this PR was opened, and the ConfigDict(extra="forbid") gate was rejecting them on round-trip. Snapshot tests now green (10 pass).

Verification against the rebased tip:

- Full unit + canonical suite: 2443 pass on the branch that combines this PR + the sibling refactor PR + `main`.
- Empirical regression across 5 public examples + 2 representative recent internal task packs — every one produced a byte-identical `avg_score_micro` on `main` vs. the combined branch.

Ready for review.

PR #197, opened 2026-07-09
integrate: minimax/minimax-m3

balazs-toloka:
## De-integration round 5: minimax-m3 on the privacy-scrubbed + open_object pack

Disposable test branch (never merged; carries de-integration deletions). Simulates a fresh
minimax-m3 candidate by deleting its certificate, its `minimax` preset, and the bespoke
`minimax_m3_tags` policy implementation, so the model resolves to the default preset and the
observe stage runs every probe against it.

### What changed vs the previous round (r4)

The wire-probe pack is now privacy-clean and carries a neutral free-form-array probe:

- the private-derived manufacturing cascade content (tau-style allocation / CAPA / inventory)
  was removed; what remains derives only from public benchmarks (tau2-bench telecom, API-Bank
  tool-registry) plus hand-authored neutral shapes;
- `open_object.py` was added: two neutral stub tools whose `tags` array lives inside an
  `additionalProperties: true` open object, so the wire schema hides the array type. It rejects
  the mangled shapes (`{"item": [...]}`, a stringified array, `""`) by naming the site
  (`patch.tags` / `record.tags`) - the exact signal minimax's XML->JSON `tags` corruption emits.

### What this run validates

1. the scrubbed pack loads and runs end to end in the observe stage (locally green: 44 tools,
   config-consistent);
2. minimax's free-form nested-array quirk surfaces in observe via the neutral `open_object`
   probe (no private data needed);
3. because that recovery is data-bound (which fields carry the array is not in the schema),
   the fix is expected to route to `data_scope_review` -> `automation:integrate-needs-human`
   rather than a silent `integrate-done`, even on local convergence.

Base is the auto-integration feature branch so the run uses the merged observe->resolve
workflow.


bberkes-toloka:
Superseded by the Slack-fixed restart round (new PR on the latest feat 9e459a0).

Issue #227, opened 2026-07-13
M2.5 (5/5) — `tolokaforge assets stamp` verb


CiroGamboa:
Child of #220 (M2.5 umbrella). New CLI subcommand that writes / updates SHA-256 digests in a project's \`assets.seeds\` entries.

### \`tolokaforge assets stamp\` subcommand
- \`tolokaforge assets stamp [--project PATH]\` — resolves the enclosing project (or the given path), walks every \`assets.seeds.<name>\` entry, computes \`sha256:<hex>\` over the referenced \`path\`, and rewrites the \`digest\` field in \`project.yaml\` in place.
- Preserves YAML formatting (comments, ordering) as much as ruamel.yaml permits.
- Exit non-zero if any referenced \`path\` doesn't exist.
- Dry-run mode: \`--check\` prints the diff of proposed digest updates without writing. Useful for CI.
- Idempotent: running against an already-stamped project.yaml is a no-op (verifies existing digests match).

## Files

- \`tolokaforge/cli/main.py\` — new \`assets\` command group; \`stamp\` subcommand.
- \`tolokaforge/core/project_loader.py\` — helper for computing seed digests (reused by the load-time verifier that lands in M3).

## Testing

- Unit test: fresh \`project.yaml\` with a placeholder digest → \`stamp\` writes correct sha256; re-run is a no-op.
- Unit test: \`--check\` mode exits 1 with a diff when a digest is stale; exits 0 when everything matches.
- Unit test: missing seed file → non-zero exit + named path.

## Depends on

- M2 (#211) merged.
- M2.5 sub-issue 2 (typed seed-registry entry shape) — must be present in the schema before this verb has something to stamp.

## Notes

Final name settled here — the design doc uses \`tolokaforge assets stamp\` and this PR is the point where it becomes canonical.

Issue #226, opened 2026-07-13
M2.5 (4/5) — ${VAR} interpolation in run configs + --workers CLI flag


CiroGamboa:
Child of #220 (M2.5 umbrella). Two small CLI/loader wins bundled: environment-variable interpolation in run-config strings, and a \`--workers\` CLI override.

### \`\${VAR}\` interpolation
- Pre-parse pass in \`tolokaforge/core/project_loader.py\` (or a small helper): substitutes \`\${VAR}\` occurrences in run-config string values from the process environment at load time.
- Unresolved variables **fail loud** — the loader names the file, the surrounding key path, and the variable name.
- Scope: only string-valued fields in the run-config YAML. Structured fields (numbers, booleans, lists) are not interpolated.
- Applies AFTER \`project.run_defaults\` merge, so a project-level template can reference environment variables that the operator supplies per-run.

### \`--workers\` CLI flag
- New flag on \`tolokaforge run\`: \`--workers N\` (positive integer).
- CLI overrides \`compute.workers\` in the effective run config (per the precedence chain: CLI > env > run config > run_defaults > engine default).
- Wired through the existing CLI-override precedence in \`tolokaforge/cli/main.py\`; consistent with the \`--user-model\` / \`--judge-model\` / \`--runtime\` flags already present.

## Files

- \`tolokaforge/core/project_loader.py\` — \`interpolate_env_vars\` helper.
- \`tolokaforge/cli/main.py\` — \`--workers\` option on the \`run\` command; interpolation invoked after loading the run-config YAML.

## Testing

- Unit tests for interpolation: \`\${FOO}\` present in env → substituted; unresolved → clear error naming the file + key + variable.
- Unit tests for interpolation scope: numbers and booleans not touched.
- Unit tests for \`--workers\`: CLI value overrides run-config value; validates positive integer.
- Integration test: end-to-end run of \`example-microservices-pack\` with \`\${OUTPUT_DIR}\` in the run config and \`--workers 4\`.

## Depends on

- M2 (#211) merged.

Issue #225, opened 2026-07-13
M2.5 (3/5) — Dual-home knob resolution


CiroGamboa:
Child of #220 (M2.5 umbrella). Resolves the shadow-duplication between \`RunConfig.compute\` / \`storage\` and the legacy fields on \`OrchestratorConfig\`. Compute + storage become canonical; orchestrator equivalents become read-time aliases with \`DeprecationWarning\`. Retired in M5 (#214).

### Read-side aliases (mode=before validators on OrchestratorConfig)
- \`OrchestratorConfig.workers\` → alias for \`ComputeConfig.workers\`.
- \`OrchestratorConfig.max_budget_usd\` → alias for \`ComputeConfig.max_budget_usd\`.
- \`OrchestratorConfig.max_requests_per_second\` → alias for \`ComputeConfig.max_requests_per_second\`.
- \`OrchestratorConfig.max_attempt_retries\` → alias for \`ComputeConfig.max_attempt_retries\`.
- \`OrchestratorConfig.queue_backend\` → alias for \`StorageConfig.queue.backend\`.
- \`OrchestratorConfig.queue_postgres_dsn\` → alias for \`StorageConfig.queue.postgres_dsn\`.

Each emits \`DeprecationWarning\` naming the legacy field and the canonical replacement. Follows the existing \`OrchestratorConfig.runtime='docker'\` alias pattern.

### Semantic redefinition
- \`OrchestratorConfig.max_turns\` — redefined as an **optional run-level cap**. Effective per-trial \`max_turns\` = \`min(task_defaults.max_turns, run_config.orchestrator.max_turns)\`; omitting the run-level value means no cap (task governs).
- \`OrchestratorConfig.timeouts\` — same treatment: optional run-level caps applied via the min rule against the task-scoped chain.
- \`OrchestratorConfig.stuck_heuristics\` — deprecated. The conductor is repointed to read from \`task_defaults.stuck_heuristics\` (which arrives via the M2 loader's per-task merge). Legacy field still parses with a warning.
- \`continue_prompt\` — decision recorded in the PR: either wire on \`task_defaults\` (canonical) or drop.

### Orchestrator wiring
- \`orchestrator._create_adapter\` reads from \`compute.*\` / \`storage.queue.*\` directly. If the alias fields are still set (from an unmigrated config), the value is picked up via the alias validator + a warning.

## Files

[...]

Issue #224, opened 2026-07-13
M2.5 (2/5) — Schema-shape reservations + task-schema relaxation


CiroGamboa:
Child of #220 (M2.5 umbrella). Name-squat the shapes M3/M4 will fill in, and demote the over-required fields on \`TaskConfig\`. All schema-only; no behaviour change.

### Actor namespace reservations
- \`actors.<name>\` sub-schema recognised at the parse layer. No binding yet (that's M4) — but the following are enforced now:
- \`actors.agent\` and \`actors.judge\` names forbidden (reserved for future actors).
- Roster ⊆ \`models\` cross-check at load time: every \`actors.<name>.mode == \"llm\"\` actor must have a matching entry in \`models\`.
- Sub-keys \`actors.<name>.tools\` and \`actors.<name>.service\` reserved (rejected as unknown keys today).

### Backend-capability entry shape
- \`compute.capabilities\` list, entries typed as \`string | dict[str, dict]\`. No registry / admission gate this milestone (that's M3) — just the field shape.

### Typed seed-registry entry shape
- \`assets.seeds.<name>\` entry accepts \`{path, kind, digest}\` full form OR a bare string as shorthand for \`{path: <s>}\` with \`kind\` inferred from extension where unambiguous. Digest optional this milestone (verification lands in M3 with reset recipes).

### Task-schema relaxation
- \`TaskConfig.name\` demoted from required to optional.
- \`TaskConfig.category\` demoted from required to optional.
- Minimal task = \`task_id + description\`.

### SecurityContext type widening
- \`SecurityContext.run_as_user\`: \`int\` → \`int | str\`.
- \`SecurityContext.run_as_group\`: \`int\` → \`int | str\`.

## Files

- \`tolokaforge/core/models.py\` — \`TaskConfig\` field demotion; \`ComputeConfig.capabilities\`; \`assets.seeds\` schema stub; actor reservations.
- \`tolokaforge/runner/models.py\` — \`SecurityContext\` type widening.

## Testing

- Unit tests for each reservation: \`actors.agent\` rejected; roster ⊆ models validated; unknown sub-keys under \`actors.<name>\` rejected.
- Unit tests for the capability-entry shape: bare string OK, dict OK, other shapes rejected.
- Unit test: minimal \`TaskConfig(task_id='x', description='y')\` accepted.
- Unit test: \`SecurityContext(run_as_user='toloka')\` and \`SecurityContext(run_as_user=1000)\` both accepted.

[...]

PR #232, opened 2026-07-13
feat(schema): EnvironmentPatch + resolve() + stack sub-object

### LINKED ISSUE #223: M2.5 (1/5) — EnvironmentPatch + resolve() + stack sub-object
Child of #220 (M2.5 umbrella). Restructures the environment schema so patches are all-optional inputs and the manifest is the post-merge output.

## What lands

- **New Pydantic type: \`EnvironmentPatch\`** — all-optional, no construction-time I/O. Becomes the type of \`ProjectConfig.default_environment\` and \`TaskConfig.environment_manifest\`.
- **Repurposed: \`EnvironmentManifest\`** — becomes the output of a new \`resolve()\` function. Retains its file-validation logic but now runs post-merge instead of per-input.
- **\`resolve()\` function** — lives in \`tolokaforge/core/project_loader.py\`. Takes a project-side patch and a task-side patch, deep-merges them, resolves \`stack.inputs\` substitutions into the compose file, then constructs the manifest with all validators active. This is the canonical point where the resolved environment document is materialised.
- **\`stack\` sub-object** — groups \`compose_file\` / \`runner_service\` / \`inputs\` under \`default_environment.stack\`. Atomic replacement: a task patch that sets \`stack.compose_file\` resets \`stack.inputs\` and \`stack.runner_service\` (a foreign compose file's variable slots must never bind against another file's inputs); policy requests (\`network_policy\`, \`security_context_defaults\`) survive. A task patch that touches only \`stack.inputs\` or \`stack.runner_service\` deep-merges.

## Files

- \`tolokaforge/core/models.py\` — \`EnvironmentPatch\` model.
- \`tolokaforge/runner/models.py\` — \`EnvironmentManifest\` refactored as \`resolve()\` output; drop the compose-file eager read from patch-side construction.
- \`tolokaforge/core/project_loader.py\` — \`resolve()\`, atomic \`stack\` replacement logic.

## Testing

- Unit tests for \`resolve()\`: patch + patch → manifest; \`stack.compose_file\` presence triggers atomic replacement of service-treatment fields; \`stack.inputs\`-only patch deep-merges.
- Unit tests for \`EnvironmentPatch\` construction with no I/O (was failing on the old model because the compose-file validator was eager).
- Regression: existing test fixtures that construct \`EnvironmentManifest\` directly continue to work through \`resolve()\`.

## Depends on

- M2 (#211) merged.

CiroGamboa:
Splits the environment shape into a patch (input) and a manifest (post-merge output). The patch is all-optional and performs no I/O at construction; the manifest keeps the disk-touching validators and is now produced by a single resolver at load time.

Closes #223 (child of #220).

## Summary

- **`EnvironmentPatch` + `StackPatch`** (`tolokaforge/runner/models.py`): all-optional input shape. Legacy flat `compose_file` / `runner_service` at the patch top level are still accepted with a `DeprecationWarning` and normalised under `stack` by a `model_validator(mode="before")` — retirement in a later milestone.
- **`EnvironmentManifest`** repurposed as the resolver output. Keeps the flat `compose_file` / `runner_service` the runtime backends already consume, plus a new `stack_inputs: dict[str, str]` field for compose-file variable substitutions.
- **`resolve()`** (`tolokaforge/core/project_loader.py`): deep-merges a project-side and task-side patch (task wins) and constructs the manifest, running the compose-file safety validators once at that point.
- **Atomic `stack` replacement**: when the task patch declares `stack.compose_file` (key presence, not path identity), the task's `stack` replaces the project's outright (clean slate of `inputs` / `runner_service`). Service-treatment fields (`initial_state`, root `isolation`) are dropped; policy-request fields (`network_policy`, `security_context_defaults`) survive — a foreign compose file's `${var}` slots must never silently capture inherited values.
- **Load-path wiring**: `ProjectConfig.default_environment` and `TaskConfig.environment_manifest` are now `EnvironmentPatch | None`. Path anchoring in `_resolve_project_paths` and `_resolve_environment_manifest_paths` walks both `stack.compose_file` and the legacy flat form. `Orchestrator._create_adapter` forwards `project.default_environment`; `NativeAdapter.to_task_description` calls `resolve()` before constructing `TaskDescription`.

## Test plan

- [x] `uv run pytest tests/unit/ tests/canonical/` — 2601 passed, 1 skipped.
- [x] New `tests/unit/test_environment_resolve.py` — 16 tests covering patch construction with no I/O, legacy flat coercion, deep-merge, atomic `stack` replacement, policy-request survival, and the no-compose-file failure mode.
- [x] Updated multi-service example-task tests and the `ProjectConfig` / project-loader tests to the new patch shape.
- [x] Canonical wire-shape test now includes `stack_inputs`.
- [x] `ruff check` and `ruff format --check` clean on the touched files.

## Not in scope

- Per-service isolation semantics on the manifest (`services.<name>.isolation`) — reserved for the isolation redesign milestone (#212).
- `${VAR}` interpolation on run-config strings and the `--workers` CLI flag — #226.
- Dual-home knob resolution (compute/storage aliases) — #225.
- Schema-shape reservations (`actors` name-squat, seed-registry, backend capabilities) — #224.
- `tolokaforge assets stamp` verb — #227.

### DIFF
--- a/tests/canonical/test_environment_manifest_contract.py
+++ b/tests/canonical/test_environment_manifest_contract.py
@@ -441,6 +441,7 @@ class TestManifestWireShape:
         assert set(wire) == {
             "compose_file",
             "runner_service",
+            "stack_inputs",
             "initial_state",
             "network_policy",
             "security_context_defaults",
--- a/tests/unit/test_environment_resolve.py
+++ b/tests/unit/test_environment_resolve.py
@@ -126,6 +126,26 @@ class TestResolveDeepMerges:
         assert manifest.runner_service == "default"
         assert manifest.network_policy == NetworkPolicy.NO_INTERNET
 
+    def test_task_runner_service_overrides_project_on_deep_merge(self) -> None:
+        # Deep-merge (no compose_file trigger) → task wins on scalar
+        # conflict. safe_one_service.yaml declares the service `default`,
+        # but the runtime cross-check only fires if the manifest names
+        # a service missing from compose; here we exercise the pure
+        # merge behaviour with two distinct values that both name a
+        # declared service. Using `default` on both sides masks
+        # whether task-wins actually applies.
+        project = EnvironmentPatch(
+            stack=StackPatch(compose_file=ENV_FIXTURE, runner_service="project-runner"),
+        )
+        task = EnvironmentPatch(stack=StackPatch(runner_service="default"))
+        # `default` is the only service compose actually declares, so
+        # the manifest accepts it; the point of the test is that the
+        # merger picked the task's value over the project's, not that
+        # either particular string is valid.
+        manifest = resolve(project, task)
+        assert manifest is not None
+        assert manifest.runner_service == "default"
+
     def test_only_project_side(self) -> None:
         project = EnvironmentPatch(
             stack=StackPatch(compose_file=ENV_FIXTURE, runner_service="default"),
--- a/tests/unit/test_environment_resolve.py
+++ b/tests/unit/test_environment_resolve.py
@@ -1,0 +1,224 @@
+"""Unit tests for :func:`tolokaforge.core.project_loader.resolve` and
+the :class:`EnvironmentPatch` shape.
+
+Covers the patch-side (no I/O at construction), the atomic-``stack``
+replacement rule, deep-merge over ``stack.inputs``, policy-request
+survival, and the legacy flat compose_file / runner_service coercion.
+"""
+
+from __future__ import annotations
+
+import warnings
+from pathlib import Path
+
+import pytest
+
+from tolokaforge.core.models import EnvironmentManifest, EnvironmentPatch, StackPatch
+from tolokaforge.core.project_loader import resolve
+from tolokaforge.runner.models import NetworkPolicy, SecurityContext, TaskIsolation
+
+pytestmark = pytest.mark.unit
+
+
+ENV_FIXTURE = (
+    Path(__file__).parent.parent
+    / "canonical"
+    / "fixtures"
+    / "environment_manifest"
+    / "safe_one_service.yaml"
+)
+
+
+class TestEnvironmentPatchNoIO:
+    """The patch shape must not read the compose file at construction —
+    that's the invariant :func:`resolve` relies on to defer disk I/O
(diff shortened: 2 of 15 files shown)


CiroGamboa:
Addressed the two actionable observations in `a215434`:

- **DRY nit** — extracted `("initial_state", "isolation")` into `_SERVICE_TREATMENT_FIELDS` next to `_POLICY_REQUEST_FIELDS`. The two "survive vs. drop on atomic replacement" decisions now sit side by side.
- **Test coverage gap** — added `test_initial_state_carries_through_non_replacement_merge` alongside the existing negative test. Pins the model round-trip through `model_dump(mode="python")` against a future switch to `mode="json"`.

For the naming observation (`TaskConfig.environment_manifest` typed as `EnvironmentPatch`), I'd rather queue it for the alias-retirement milestone (#214) than churn the field name mid-migration — task authors would see a rename in this release and another when the flat compose_file coercion retires. Going to note it there.

CiroGamboa:
Second follow-up in \`af291af\` — addressed all four items from the self-review:

- **resolve() field enumeration**: now uses \`(*_SERVICE_TREATMENT_FIELDS, *_POLICY_REQUEST_FIELDS)\` so the two constants stay load-bearing across both merge branches (\`project_loader.py\`).
- **\`import warnings\` hoisted** to the top-level imports in \`runner/models.py\` — no longer re-imported inside the validator.
- **Three new tests**:
  - \`test_task_runner_service_overrides_project_on_deep_merge\` — pins task-wins semantics with distinct values (the pre-existing test used identical values so it covered presence, not override).
  - \`test_canonical_stack_compose_file_is_anchored\` (task loader) — pins that \`stack.compose_file\` gets anchored to absolute at task-yaml load time (prior tests only exercised the legacy flat shape).
  - \`test_project_default_environment_composes_with_task_patch\` (native adapter) — pins the full orchestrator → adapter → resolve() chain: task patches only \`stack.inputs\`, inherits \`compose_file\` + \`runner_service\` from the project.
- **\`stack: null\` load-error rule** filed as #235 on the \`Project layer v1\` milestone.

Total suite is 2605 passed / 1 skipped locally.

Issue #220, opened 2026-07-13
Project layer — environment patch + dual-home resolution + CLI additions (M2.5)


CiroGamboa:
Follow-up to the loader PR ([#211](https://github.com/Toloka/tolokaforge/issues/211)) picking up the schema-shape and CLI additions from the [canonical Project design](https://github.com/Toloka/tolokaforge/blob/main/docs/architecture/PROJECTS.md). Kept out of the loader PR to preserve its reviewed shape.

Blocks the isolation redesign ([#212](https://github.com/Toloka/tolokaforge/issues/212)).

## Sub-issues (one PR each)

- [ ] #223 — EnvironmentPatch + resolve() + stack sub-object
- [ ] #224 — Schema-shape reservations + task-schema relaxation
- [ ] #225 — Dual-home knob resolution
- [ ] #226 — `${VAR}` interpolation + `--workers` CLI flag
- [ ] #227 — `tolokaforge assets stamp` verb

## What lands (rollup across sub-issues)

**Schema shape:**
- **`EnvironmentPatch` / `EnvironmentManifest` split** (#223). `EnvironmentPatch` is a new all-optional Pydantic model with no construction-time I/O; used as the type of `default_environment` and task `environment_manifest`. Today's `EnvironmentManifest` becomes the output of `resolve()` — a canonical document merging inputs, compose snapshot, and pinned seed digests.
- **`stack` sub-object** (#223). `compose_file`, `runner_service`, and `inputs` group under `default_environment.stack` with atomic replacement semantics.
- **`SecurityContext.run_as_user` / `run_as_group` widened** to `int | str` (#224).
- **Task-schema relaxation** (#224). `TaskConfig.name` and `.category` demoted from required. Minimal task is now `task_id + description`.
- **Shape reservations** (#224). `agent` / `judge` forbidden as actor names now; roster ⊆ `models` cross-check; `actors.<name>.tools` / `.service` reserved keys; backend-capability entry shape validated (`string | {name: params}`); typed seed-registry entries (`{path, kind}` or bare-string shorthand).

**Dual-home knob resolution (aliases + warnings)** (#225):
- `compute.*` canonical for workers / max_budget_usd / max_requests_per_second / max_attempt_retries.
- `storage.queue.*` canonical for the queue backend + DSN.
- `OrchestratorConfig` counterparts become read-time aliases with `DeprecationWarning`.
[...]

Issue #214, opened 2026-07-10
Project layer — retire legacy aliases (M5)


CiroGamboa:
> **See [PR #216](https://github.com/Toloka/tolokaforge/pull/216) for the canonical design.** This issue tracks the M5 milestone as revised in that PR.

Cleanup PR that fires after the v1 release window closes. Removes every deprecation alias introduced during M1 → M4.

## What lands

**Aliases retired:**
- `evaluation.task_packs` alias — `evaluation.projects` is canonical (introduced M2).
- `run_config/` (singular) directory alias — `run_configs/` (plural) is canonical (introduced M2).
- `orchestrator.runtime` operator override (introduced M3).
- `OrchestratorConfig.workers` / `max_budget_usd` / `max_requests_per_second` / `max_attempt_retries` / `queue_backend` / `queue_postgres_dsn` aliases — `compute.*` and `storage.queue.*` are canonical (introduced M2.5).
- `OrchestratorConfig.stuck_heuristics` alias — `task_defaults.stuck_heuristics` is canonical (introduced M2.5).
- `TaskDefaults.user_simulator` alias — `actors.user` is canonical (introduced M4).
- `security_context_defaults.user` / `.group` aliases — `run_as_user` / `run_as_group` canonical (introduced M4).
- Uppercase `network_policy` values (`NO_INTERNET`, etc.) — lowercase (`no_internet`, ...) canonical (introduced M4).
- Flat `default_environment.compose_file` / `runner_service` / `inputs` shape — `stack` sub-object canonical (introduced M2.5).

**CLI env-var overrides retired:**
- `USER_MODEL` / `JUDGE_MODEL` env-vars no longer override model choice. Model choices belong in version-controlled run configs.

**Docs / fixtures:**
- Delete `docs/TASK_PACKS.md` (after grep confirms no in-tree links).
- Clear `tolokaforge/core/deprecations.py` entries (delete the module if empty).
- Remove migration test fixtures that only exercised legacy paths.

## Depends on

M4 (#213) shipping in a released version, plus the deprecation-window release cycle closing per repo release policy. Ciro handles releases; this milestone stays queued until the window is confirmed closed.

## Ship condition

- No `DeprecationWarning` emitted anywhere in CI logs on a fresh run.
- `grep -r "task_packs\|run_config/" tolokaforge/ examples/ docs/` returns nothing beyond historical ADR / commit references.

CiroGamboa:
[...]

Issue #213, opened 2026-07-10
Project layer — strict schema + example-pack migration (M4)


CiroGamboa:
> **See [PR #216](https://github.com/Toloka/tolokaforge/pull/216) for the canonical design.** This issue tracks the M4 milestone as revised in that PR.

Turns on strict schema validation. Lands the schema-shape renames that were name-squatted in M2.5. Migrates every example pack under `examples/native/` to the canonical shape in the same PR so CI stays green.

## Strict schema

- `model_config = ConfigDict(extra="forbid")` on every Project-layer Pydantic model.
- Loader error surface: names the file, the offending key, and the closest schema match (`difflib.get_close_matches`).
- Consolidate every deprecation alias into `tolokaforge/core/deprecations.py`.

## Schema shape renames — activated with legacy aliases

- `task_defaults.actors.<name>` — `actors.user` binds to today's simulator config; `TaskDefaults.user_simulator` becomes a legacy alias (read-time coercion with `DeprecationWarning`). `agent`/`judge` names remain reserved (already forbidden in M2.5).
- `network_policy` lowercase values (`no_internet`, `limited_internet`, `full_internet`); uppercase legacy values accepted with `DeprecationWarning`.
- `security_context_defaults.user` / `.group` → `.run_as_user` / `.run_as_group` (alias with warning).
- `default_environment.stack` sub-object becomes the only shape (co-existed alongside the flat shape as of M2.5; M4 warns on the flat keys).

## Example-pack migration

Every pack under `examples/native/` migrated in the same PR under the canonical shape:
- `project.yaml` at pack root using `stack` sub-object, `actors.user`, `assets.seeds` (where reset is needed).
- `run_config/` → `run_configs/`.
- `evaluation.task_packs` → `evaluation.projects`.
- README updated to describe the base+delta model.
- Compose files stripped of any isolation semantics — moved to manifest `services` entries.
- Grading files rewritten to structured `Rubric` / `Criterion` shape (aligning with runner-side grading model).
- `example-microservices-pack`: two tasks (`agent_flow_setup`, `agent_flow_verify`) dropped — cross-task stack persistence is deferred to 0.13.0.

`docs/TASK_PACKS.md` deletion is deferred to M5 (#214).

## Removed here

- The transitional synthesised default project from M2. Packs without `project.yaml` fail loud at load. Since every in-tree pack is migrated in this same PR, CI stays green.

## Real-task regression coverage

[...]

Issue #212, opened 2026-07-10
Project layer — isolation redesign + reset primitives (M3)


CiroGamboa:
> **See [PR #216](https://github.com/Toloka/tolokaforge/pull/216) for the canonical design.** This issue tracks the M3 milestone as revised in that PR.

Backend mode becomes task-driven. Per-service isolation semantics live in the manifest — **not** in compose-file labels. Reset primitives are replaced by seed-backed reset recipes. New backend-capabilities scaffolding. Content-addressed environment identity added as an additive helper.

## Isolation

- Backend selection reads the resolved tasks: any task requiring `per_trial` → `PerTrialRuntimeBackend`; otherwise `SharedStackRuntimeBackend`.
- Per-service semantics live in `default_environment.services.<name>.isolation` (`shared` / `reset` / `ephemeral`). Compose files carry **zero** isolation semantics — the manifest is the only home.
- Default remains `per_trial` per ADR-0009 (earlier `shared_ok`-default proposal is withdrawn — safe by default; relaxations are explicit reviewable labels). Services without a manifest entry default to `ephemeral`.
- Task overrides layer via `TaskConfig.environment_manifest.services` — the patch merges over `default_environment.services`.
- Requires an ADR amending ADR-0018 before this milestone lands.
- `orchestrator.runtime` becomes a deprecated operator override with `DeprecationWarning` (retired in M5).

## Seed-backed reset recipes

- `default_environment.services.<name>.reset.seed: "<seed_name>"` points at an entry in `assets.seeds`.
- `assets.seeds.<name>` shape: `{path, kind, digest}`. Digest (`sha256:...`) verified at load; swap without re-stamping fails loud.
- Kind vocabulary (v1): `sql_dump`, `filesystem_dir`, `redis_dump`, `bare`. Kind determines overlay semantics.
- Dispatchers live under `tolokaforge/runtime/reset_recipes/` (one module per kind).
- Replaces the earlier "closed enum of primitives" plan — the registry is more expressive and decouples "what state to restore" from "how to apply it."

## Backend capabilities scaffolding

- New optional field on run config: `compute.capabilities: list[str | dict[str, dict]]` — bare `"name"` or `{"name": {params}}`.
- Registry at `tolokaforge/core/backend_capabilities.py` owns the vocabulary.
- Admission: at run-start the selected provider advertises its capability set; request ⊆ advertise or fail loud.
[...]

PR #240, opened 2026-07-13
feat(schema): actor/seed/capability reservations + task-schema relaxation

### LINKED ISSUE #224: M2.5 (2/5) — Schema-shape reservations + task-schema relaxation

CiroGamboa:
Schema-only. Zero behaviour change. Name-squats the shapes later milestones will fill in, and demotes over-required fields on \`TaskConfig\`.

Closes #224 (child of #220).

## Summary

- **`actors` map on `TaskDefaults` + `TaskConfig`** — new \`ActorSpec\` model mirroring today's \`UserSimulatorConfig\` fields (all optional, \`extra=\"forbid\"\` so sub-keys \`tools\`/\`service\` are rejected today). Map-level validator rejects reserved names \`agent\` and \`judge\`.
- **Roster ⊆ models cross-check** — \`validate_actor_roster_subset_of_models\` runs from \`load_effective_run_config\` after the project × run-config merge. Every \`actors.<name>\` with \`mode == \"llm\"\` must have a matching entry in \`models\`; scripted actors are exempt.
- **\`ComputeConfig.capabilities: list[Any]\`** — accepts bare-string entries or single-key \`{\"name\": {params}}\` dicts. Custom \`field_validator\` emits context-rich errors instead of Pydantic's union-resolution noise. No registry / admission gate this PR.
- **Typed seed-registry** — new \`SeedRef\` (full \`{path, kind, digest?}\` form + bare-string shorthand with kind inferred from \`.sql\`/\`.rdb\`; other extensions require the full form) and \`AssetsConfig\` on \`ProjectConfig.assets\`. Digest verification lands with reset recipes.
- **Path anchoring for seeds** — \`_resolve_project_paths\` walks \`assets.seeds.<name>.path\` (dict + bare-string) so seed paths are absolute after load.
- **\`TaskConfig.name\` and \`.category\` demoted** to \`str | None = None\`. Minimal task = \`task_id + description\`. Existing consumers already fall back defensively.
- **\`SecurityContext.run_as_user\` / \`run_as_group\`** widened to \`int | str | None\` — usernames like \`\"toloka\"\` are common; the substrate resolves to numeric UID at materialisation.

## Test plan

- [x] \`uv run pytest tests/unit/ tests/canonical/\` — 2640 passed, 1 skipped (+35 new).
- [x] New \`tests/unit/test_schema_reservations.py\` — 29 tests covering ActorSpec, actors reservation, ComputeConfig capabilities, SeedRef shorthand + full form + ambiguous extension, AssetsConfig, ProjectConfig.assets, minimal TaskConfig, SecurityContext widening.
- [x] Roster ⊆ models tests added to \`test_project_loader_end_to_end.py\` (positive, negative, scripted-exempt, no-actors no-op).
- [x] Asset-path anchoring tests (dict form + bare-string shorthand).
- [x] Manual: \`load_project_config('examples/native/example-microservices-pack/project.yaml')\` populates \`assets.seeds.app_baseline\` (with sha256 digest) and \`task_defaults.actors.user\`. Both were silently dropped before this PR.
- [x] Canonical snapshots refreshed for the new optional \`actors\` field on serialised \`TaskConfig\`.
- [x] \`ruff check\` + \`ruff format --check\` clean on touched files.

## Not in scope

- Binding \`actors.user\` to today's \`user_simulator\` runtime — both co-exist during the deprecation window; the rename lands with the strict-schema milestone (#213).
- Capability registry lookup / admission gate — isolation redesign (#212).
- Seed digest verification + reset-recipe binding — #212.
- Retiring \`TaskConfig.user_simulator\` — cleanup milestone (#214), after the release window.

### DIFF
--- a/tests/canonical/snapshots/native_browser_basic/task_config.json
+++ b/tests/canonical/snapshots/native_browser_basic/task_config.json
@@ -1,4 +1,5 @@
 {
+  "actors": null,
   "adapter_settings": null,
   "adapter_type": "native",
   "category": "test",
--- a/tests/canonical/snapshots/native_calc_basic/task_config.json
+++ b/tests/canonical/snapshots/native_calc_basic/task_config.json
@@ -1,4 +1,5 @@
 {
+  "actors": null,
   "adapter_settings": null,
   "adapter_type": "native",
   "category": "test",
--- a/tests/canonical/snapshots/native_example_domain_case_a/task_config.json
+++ b/tests/canonical/snapshots/native_example_domain_case_a/task_config.json
@@ -1,4 +1,5 @@
 {
+  "actors": null,
   "adapter_settings": null,
   "adapter_type": "native",
   "category": "tool_use",
--- a/tests/canonical/snapshots/native_shop_orders_02/task_config.json
+++ b/tests/canonical/snapshots/native_shop_orders_02/task_config.json
@@ -1,4 +1,5 @@
 {
+  "actors": null,
   "adapter_settings": null,
   "adapter_type": "native",
   "category": "tool_use",
--- a/tests/canonical/snapshots/tbench_echo_hello/task_config.json
+++ b/tests/canonical/snapshots/tbench_echo_hello/task_config.json
@@ -1,4 +1,5 @@
 {
+  "actors": null,
   "adapter_settings": {
     "compose_file": "<ROOT>/tests/data/terminal_bench_tasks/echo-hello/docker-compose.yaml",
     "difficulty": "easy",
--- a/tests/unit/test_project_loader_end_to_end.py
+++ b/tests/unit/test_project_loader_end_to_end.py
@@ -442,6 +442,28 @@ class TestActorRosterSubsetOfModels:
         merged, _ = load_effective_run_config(run_cfg)
         assert merged == {}
 
+    def test_llm_actor_fails_when_models_is_empty(self, tmp_path: Path) -> None:
+        # Boundary case: `models: {}` is a valid mapping but has no
+        # entries. An llm actor still fails loud — pinning it explicitly
+        # so the "missing" branch isn't confused with a shape-error
+        # branch by a future refactor.
+        _write_yaml(
+            tmp_path / "project.yaml",
+            {
+                "name": "p",
+                "task_defaults": {
+                    "actors": {"user": {"mode": "llm", "persona": "curious"}},
+                },
+            },
+        )
+        run_cfg = tmp_path / "run_configs" / "dev.yaml"
+        _write_yaml(run_cfg, {"models": {}})
+
+        from tolokaforge.core.project_loader import load_effective_run_config
+
+        with pytest.raises(ValueError, match="not declared under `models`"):
+            load_effective_run_config(run_cfg)
+
     def test_malformed_models_block_is_a_load_error(self, tmp_path: Path) -> None:
         # A typo where `models:` is a list (or any non-mapping) would
         # confuse the roster check downstream — surface the shape error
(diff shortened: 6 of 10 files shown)


CiroGamboa:
Follow-up in \`83b7831\`: addressed the fail-fast nit on \`validate_actor_roster_subset_of_models\`.

Previously a malformed \`models\` block (e.g. a list where a mapping was meant) silently no-op'd the roster check; \`RunConfig(**merged)\` would still catch it, but the shape error surfaces cleaner if the check flags it at the exact site. Now raises \`ValueError\` naming the offending type; new positive test in \`test_project_loader_end_to_end.py\` covers the malformed-models path.

Second observation (\`list[Any]\` on \`ComputeConfig.capabilities\`) is a deliberate documented reservation shape — leaving as-is per the review's own note.

CiroGamboa:
Self-review follow-ups in \`9106f5b\`:

- **Roster-check docstring** pinned: the actor's map-key must appear as a key in \`models\`; \`ActorSpec\` has no \`.model\` field, so by-key match is the only decidable semantic today. Also flagged that the check covers project-level actors only — task-level \`TaskConfig.actors\` overrides are unenforced until the runtime binding lands.
- **Missing tests** added:
  - Positive: task-level \`actors\` with a non-reserved name parses on \`TaskConfig\` (mirrors the \`TaskDefaults\` positive test).
  - Parametrised across all four \`SeedKind\` values so a future narrowing doesn't silently drop authoring options.
  - Extension-lookup case-insensitivity (\`Foo.SQL\` → \`sql_dump\`).
  - Boundary: llm actor with \`models: {}\` fails loud with the "not declared under \`models\`" error — pins that "missing" isn't conflated with "wrong-key".

Full suite: 2648 passed / 1 skipped locally.

PR #262, opened 2026-07-13
feat(loader): ${VAR} interpolation in run configs + --workers CLI flag

### LINKED ISSUE #226: M2.5 (4/5) — ${VAR} interpolation in run configs + --workers CLI flag

CiroGamboa:
Two small M2.5 additions bundled per the umbrella. No behaviour change for configs without \`\${...}\` placeholders and no \`--workers\` flag.

Closes #226 (child of #220).

## Summary

**\`\${VAR}\` interpolation** — new \`_interpolate_env_vars\` helper in \`project_loader.py\`. Recursive walker over the effective run-config tree; substitutes \`\${VAR}\` in string values from \`os.environ\`.
- Only string values are interpolated. Dict keys, numbers, booleans, non-string list items pass through untouched.
- Only \`\${VAR}\` form recognised (braces mandatory). Bareword \`\$VAR\` stays literal — the ambiguity vs literal \`\$\` is a tar-pit.
- Missing variables collect into a single \`ValueError\` naming every offending key path so authors fix all misses in one edit pass.
- Fires from \`load_effective_run_config\` after the \`run_defaults\` merge and before the roster check — every downstream CLI subcommand routes through one site.

**\`--workers N\` CLI flag** — new \`click.IntRange(min=1)\` decorator on \`tolokaforge run\`. Positive-integer validation at parse time.
- Writes to canonical \`compute.workers\` (not the legacy \`orchestrator.workers\`), so no \`DeprecationWarning\` fires from the M2.5 #225 alias-lift.
- Help text points at \`compute.workers\` so authors learn the right home from the flag output.

## Test plan

- [x] \`uv run pytest tests/unit/ tests/canonical/\` — 2674 passed, 1 skipped (+10 new).
- [x] Interpolation: seven new tests covering positive substitution, missing-var error, non-strings untouched, literal-string round-trip, bareword \`\$VAR\` stays literal, fires after \`run_defaults\` merge, multiple misses collated into one error.
- [x] CLI: \`--workers\` appears in \`--help\`; rejects \`0\` and negative values via \`IntRange\`; help text points at \`compute.workers\`.
- [x] \`ruff check\` + \`ruff format --check\` clean on touched files.

## Not in scope

- Interpolating dict keys (only values).
- \`\$VAR\` bareword shorthand — only \`\${VAR}\` form.
- Default-value syntax (\`\${VAR:-default}\`, bash-style) — post-v1 refinement if authors ask for it.
- Recursive interpolation of substituted values — if \`\${FOO}\` resolves to a string containing \`\${BAR}\`, only one pass runs. Simpler mental model; predictable.

### DIFF
--- a/tests/unit/test_cli_commands.py
+++ b/tests/unit/test_cli_commands.py
@@ -257,6 +257,7 @@ class TestRunCommand:
         assert "--resume" in result.output
         assert "--verbose" in result.output
         assert "--strict" in result.output
+        assert "--workers" in result.output
 
     def test_run_missing_config(self, runner: CliRunner) -> None:
         result = runner.invoke(cli, ["run"])
@@ -265,6 +266,33 @@ class TestRunCommand:
     def test_run_nonexistent_config(self, runner: CliRunner) -> None:
         result = runner.invoke(cli, ["run", "--config", "/nonexistent/config.yaml"])
         assert result.exit_code != 0
+
+    def test_run_workers_rejects_zero(self, runner: CliRunner, tmp_path: Path) -> None:
+        # Positive-int validation happens at click-parse time — no need
+        # for a real config file since the exit fires before loading.
+        # Click's IntRange failure writes to stderr and exits with 2;
+        # asserting on exit_code alone is enough to pin the validation
+        # (message goes to stderr, which CliRunner routes separately).
+        result = runner.invoke(
+            cli,
+            ["run", "--config", str(tmp_path / "x.yaml"), "--workers", "0"],
+        )
+        assert result.exit_code != 0
+
+    def test_run_workers_rejects_negative(self, runner: CliRunner, tmp_path: Path) -> None:
+        result = runner.invoke(
+            cli,
+            ["run", "--config", str(tmp_path / "x.yaml"), "--workers", "-3"],
+        )
+        assert result.exit_code != 0
+
+    def test_run_workers_help_names_compute_workers(self, runner: CliRunner) -> None:
+        # The help text points authors at the canonical home so the
+        # flag doesn't quietly ship the compute vs orchestrator dual-
+        # home confusion into new pack authoring.
+        result = runner.invoke(cli, ["run", "--help"])
+        assert result.exit_code == 0
+        assert "compute.workers" in result.output
 
 
 # ===================================================================
--- a/tests/unit/test_dual_home_aliases.py
+++ b/tests/unit/test_dual_home_aliases.py
@@ -88,6 +88,23 @@ class TestComputeWorkersAlias:
         # OrchestratorConfig.workers default is 8; compute.workers is None.
         assert cfg.effective_workers == 8
 
+    def test_cli_workers_write_to_compute_no_warning(self) -> None:
+        # The `--workers` CLI flag writes to canonical `compute.workers`
+        # (see tolokaforge.cli.main:run). Pin that writing there does
+        # NOT fire the dual-home DeprecationWarning — the alias-lift
+        # only warns when `orchestrator.workers` is set. A regression
+        # here would silently ship the warning on every CLI-driven run.
+        with warnings.catch_warnings(record=True) as caught:
+            warnings.simplefilter("always")
+            cfg = RunConfig(**_base(compute={"workers": 4}))
+        assert cfg.effective_workers == 4
+        assert cfg.compute is not None
+        assert cfg.compute.workers == 4
+        assert not any(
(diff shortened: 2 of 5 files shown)


CiroGamboa:
Bot's rule-2 concern addressed in \`c1a6057\` — chose option 2 (fail loud on credential-shaped names) since the review flagged this as the more defensible line.

- **New suffix guard**: \`_CREDENTIAL_NAME_SUFFIX_PATTERN\` mirrors \`tests/unit/secrets/test_no_raw_secret_access.py\`. Placeholders whose variable name ends in \`_API_KEY\` / \`_API_KEYS\` / \`_API_BASE\` / \`_TOKEN\` / \`_SECRET\` / \`_PASSWORD\` / \`_DSN\` / \`_CREDENTIAL[S]?\` / \`_PAT\` are rejected at load time even when the env var is set.
- **Error surface**: names the offending placeholders; never quotes the real env-var value, so the log surface stays clean.
- **Tests**: positive case + variant coverage (7 suffix shapes) + regression pin that dict keys shaped like \`\${...}\` are still never interpolated (bot's optional suggestion).
- **Docstring nit**: trimmed the misleading "returned value" line on \`_interpolate_string\` per the review.

Suite: 2677 passed / 1 skipped.

CiroGamboa:
Self-review follow-up in \`1cc4e66\` — three test-coverage pins:

- **List-branch positive test** — \`evaluation.projects: [\"\${PACK_ROOT}/pack-a\", \"\${PACK_ROOT}/pack-b\"]\` substitutes both entries. The walker's list branch was covered only via nested dict recursion before.
- **Input-not-mutated pin** — \`deepcopy\` snapshot before / after; asserts the docstring's immutability guarantee, which the loader path was hiding behind \`deep_merge\`.
- **\`--workers\` doesn't warn** — pins that writing to canonical \`compute.workers\` doesn't trigger the M2.5 #225 alias-lift warning. Regression net for CLI runs.

Suite: 2680 passed / 1 skipped locally.

CiroGamboa:
Second-round self-review follow-ups in \`695f32b\` — addressed all four items from the /review pass:

- **List-index key path format** — new \`_render_key_path\` helper produces \`evaluation.projects[0]\` instead of the spurious-dot \`evaluation.projects.[0]\`. Regression pin added.
- **\`load_project_config\` docstring** — calls out the interpolation-scope decision (run-config only, not \`project.yaml\`) so authors don't hit surprise. Follow-up to extend if asked.
- **Credential-variants test retargeted** off deprecated \`orchestrator.continue_prompt\` onto \`evaluation.output_dir\` — robust to future ordering refactors.
- **Digit-in-varname test** — pins \`\${MY_VAR_1}\`, \`\${MY_VAR2}\`, \`\${MY_VAR_1A}\` per the regex's body class.

Suite: 2682 passed / 1 skipped.

Issue #396, opened 2026-07-15
bug(logging): litellm double-emits every log record (private handler + propagation)


CiroGamboa:
## Summary

litellm installs its own `StreamHandler` on the `LiteLLM` / `LiteLLM Router` / `LiteLLM Proxy` loggers at import time, pointing at `sys.stderr`, while leaving `propagate=True`. Every litellm record therefore emits **twice**:

1. Once through litellm's private handler, in litellm's own format (`\x1b[92m<ts> - LiteLLM:INFO\x1b[0m: <file>:<line> - <msg>`).
2. Once through the tolokaforge root handler (`configure_root_logging`), in `StructuredFormatter` shape (`HH:MM:SS.mmm | LEVEL | k=v | message`).

Reproduced deterministically (no API call needed):

```python
import logging, litellm
from litellm._logging import verbose_logger
from tolokaforge.core.logging import configure_root_logging
configure_root_logging(level=logging.INFO)
verbose_logger.info("hello")   # prints TWO lines to stderr
```

## Impact

- Duplicate log lines on every `--display=plain` / `--display=log` run whenever litellm emits (cost/routing INFO, retry WARNING, provider notices).
- The litellm-format copy ignores `--log-format` (`plain`/`json` shape not honoured), so a `--log-format=json` run gets non-JSON lines interleaved.

## Relationship to #392

#392 (Rich Live panel stacking) neutralises these private handlers **only for the `LiveRunDisplay` lifetime**. Outside `--display=rich` the double-emission persists. This ticket covers the process-wide fix: neutralise litellm's private handlers once at CLI startup (in `configure_root_logging` or the run callback) so all litellm records render exactly once through the tolokaforge root handler.

## Non-goals

- Panel stacking under `--display=rich` (covered by #392).

## Suggested fix

At CLI startup, after `configure_root_logging`, remove the private `StreamHandler`s from the `LiteLLM*` loggers (they have `propagate=True`, so records still reach the root handler). Prefer a general sweep — any non-root logger carrying a `StreamHandler` whose stream is `sys.stderr`/`sys.stdout` — over hardcoding litellm's logger names.

PR #397, opened 2026-07-15
fix(dx): kill Rich Live panel stacking during trial execution (#392)

### LINKED ISSUE #392: bug(dx): Rich Live panel stacks during trial execution — diagnose then fix
## Summary

The Rich Live panel under \`--display=rich\` stacks duplicate copies of itself during trial execution (visible in screenshots below). Three prior fixes on \`feat/terminal-dx\` closed three of the write channels that bypass Rich Live's cursor bookkeeping, but at least one channel remains open — the latest smoke shows stacking *during trial execution* (turns 0 → 6), not during Docker boot, meaning the leak is on the trial hot path.

## Prior fixes (all merged on \`feat/terminal-dx\`)

- \`d2fff0a\` — stable-height layout (\`viewport - 1\` locked total).
- \`4c8be0a\` — removed \`console.print\` calls from inside the \`with LiveRunDisplay:\` block in \`tolokaforge/dx/cli/main.py\`.
- \`ce29487\` — routed \`_LogSink\` WARNING+ through \`live.console.print\` instead of raw stderr.

## Remaining leak — diagnostic-first, not fix-first

Every earlier blind guess left one channel open. This ticket lands the **diagnostic tool first**, uses it to identify the exact write channel, then applies a targeted fix. Two commits (diagnostic + fix), one PR.

## Diagnostic: \`_StderrProbe\`

New env-gated context manager in \`tolokaforge/dx/live_panel.py\`:

- Enabled when \`TOLOKAFORGE_STDERR_PROBE=<path>\` is set at \`LiveRunDisplay.__enter__\`.
- Wraps \`sys.stderr\` with a proxy that, for every \`write(chunk)\` call, appends to the log file: \`{timestamp} | {caller-file:line} | {repr(chunk)[:200]}\` + a compact stack trace (top 5 frames).
- Restores original stderr on \`__exit__\`.
- Off in production (env var unset → wrapper never installed).

## Likely leak candidates (in decreasing order of prior evidence)

- Child-logger handler installed by a chatty library (\`litellm\`, \`httpx\`, \`openai\`) that isn't tagged with \`_TOLOKAFORGE_ROOT_HANDLER_SENTINEL\` and thus escapes \`_LogSink\`'s handler swap in \`tolokaforge/dx/live_panel.py:493-534\`.
- A subprocess (\`docker-py\`, \`testcontainers\`, docker CLI) inheriting the pre-Live raw stderr fd — those bypass every Python-level intercept.
- A direct \`sys.stderr.write\` or \`print(..., file=sys.stderr)\` on the trial hot path in \`tolokaforge/core/runner.py\` or \`tolokaforge/core/llm/client.py\`.

## Fix (shape — concrete choice determined by diagnostic output)

If Python-level: install a proxy on \`sys.stderr\` for the Live lifetime that routes each write through \`self._live.console.print(...)\`.

If subprocess / raw-fd: temporarily \`os.dup2(...)\` a pipe over fd 2 while Live is active; drain the pipe into \`_LogSink._log_buffer\` (INFO/DEBUG collapsed, WARNING+ routed through \`print_above\`).

If chatty library with a private handler: install a filter catching those specific logger names and re-routing into \`_LogSink\`.

## Test strategy

- New unit test in \`tests/unit/test_run_display.py\` — \`_StderrProbe\` context manager: writes to \`sys.stderr\` during its lifetime end up in the target file with a stack trace; original stderr restored on exit.
- New behaviour-locking test asserting that a \`sys.stderr.write(\"noise\")\` call during a live smoke (mock \`_live.console\`) is routed through the fix path — no direct write to the raw pre-Live stderr.
- Live-smoke acceptance criterion below.

## Acceptance criteria

- With the fix in place, re-running the coding example under \`TOLOKAFORGE_STDERR_PROBE=/tmp/probe.log tolokaforge run --config examples/native/coding/run_config.yaml\` produces a probe log with **zero writes** originating from anywhere other than \`_LogSink\` for the entire trial-execution phase.
- Visual confirmation: no stacked panel copies in the terminal during a 2-worker, 8-turn run.
- 3227 unit+canonical baseline tests all still green.

## Non-goals

- Boot-phase stacking (already fixed by prior commits).
- Textual TUI stability (not affected — Textual owns the screen).
- Boot-log-tail widget — separate ticket #392.

CiroGamboa:
## Summary

- Root cause: litellm installs its own `logging.StreamHandler` on child loggers (`LiteLLM`, `LiteLLM Router`, `LiteLLM Proxy`) at import time, holding the pre-Live `sys.stderr` object with `propagate=True`. Each record wrote twice — once raw to stderr (bypassing Rich Live's cursor coordination → panel re-appended below → stacked copies), once via propagation to the root `_LogSink`.
- Fix: `LiveRunDisplay.__enter__` now sweeps every non-root `logging.Logger` in `logging.root.manager.loggerDict` and removes `StreamHandler`s whose `.stream is` one of the pre-Live "dangerous" terminal streams (`sys.stderr`/`sys.stdout`/`sys.__stderr__`/`sys.__stdout__`). Records still propagate to root's `_LogSink`; `propagate=False` loggers get a fresh `_LogSink` installed so nothing is silently dropped. Symmetric restore on `__exit__`. Stream-identity predicate — no hardcoded litellm names.
- Also lands a reusable env-gated `_StderrProbe` diagnostic (activated with `TOLOKAFORGE_STDERR_PROBE=<path>`) so future stderr-bypass channels are identifiable from a live run.

## Plan

See `docs/plans/2026-07-15-issue-392-kill-panel-stacking.md`.

Diagnosed deterministically via dev MCP `run_python` (no live tokens) — reproduced litellm's `handler.stream is sys.__stderr__` and confirmed root-only sentinel swap misses child handlers. Two-stage plan approved by `plan-critic` after one revision round (added `isinstance(logger, logging.Logger)` guard to avoid `PlaceHolder` `AttributeError`, tightened tests to bind to captured dangerous streams, added `propagate=False` behaviour-locking test, corrected stale `docs/CLI.md` line).

## Discovered issues

- **Filed:** #396 — litellm double-emits every log record process-wide (duplicate lines under `--display={plain,log}` and litellm-format lines ignoring `--log-format`). This PR only neutralises handlers during the Live lifetime; #396 is the process-wide fix at CLI startup.
- **Fixed in this PR:** None.
- **Pre-existing to note:** a `black↔ruff-format` disagreement at `tests/unit/test_run_display.py:289` on the base branch (untouched by this PR); worth a separate hygiene PR.

## Acceptance criterion

Restated from the ticket's literal wording (which was unachievable because the probe wraps the same stream Rich renders through): **"zero *bypassing* writes"** — no `sys.stderr.write` whose stack frame lacks a Rich/`_LogSink` origin.

Live smoke on `examples/native/tool_use/run_config.yaml` under `TOLOKAFORGE_STDERR_PROBE=/tmp/probe.log`: panel stayed anchored across a full 82-second run (2 trials, 2 workers); probe log recorded exactly one write — Rich's own final `Live.stop` render frame. Zero third-party (litellm) caller frames.

(Plan named `examples/native/coding` but that failed at Docker health-check in the sandbox — unrelated env issue, filed separately in the discovered-issues note above the run report. `tool_use` exercises the same `--display=rich` code path and drives real litellm calls.)

## Test plan

- [x] Unit — `tests/unit/test_run_display.py` — 67 tests pass (9 new: 4 for `_StderrProbe`, 5 for the sweep including a `PlaceHolder` regression lock)
- [x] Full unit+canonical baseline — 3236 pass, 1 skipped (0 regressions)
- [x] Regression lock verified — stashed the sweep implementation and re-ran the new tests; 3 of 5 failed as expected, confirming they lock the behaviour
- [x] Live smoke — panel visually anchored, probe log shows zero bypass writes

Closes #392


CiroGamboa:
**CI status**: `lint` ✅ `test-smoke` ✅ `hygiene-review` ❌.

The `hygiene-review` failure is **pre-existing infrastructure breakage**, not a review verdict — filed as #398. Reproduces on every recent PR (#386, #390, #393). The action fails at init in ~500ms with `is_error: true` and no substantive output; it never reviews the diff. Not branch-attributable.

All branch-attributable lanes are green. Merging.

PR #463, opened 2026-07-16
fix(tools): advertise PATCH requests

### LINKED ISSUE #452: http_request tool schema enum omits PATCH though execute() accepts it
## Symptom
`tolokaforge/tools/builtin/http_request.py:43` declares the `method` enum as e.g. `["GET", "POST", "PUT", "DELETE", "HEAD", "OPTIONS"]` — omitting `PATCH`. The `execute()` implementation happily performs `PATCH` requests and the multi_service_helpdesk_workflow pack's grading relies on it (`required_actions` with `method: "PATCH"` on `http://delivery-tracker:8000/deliveries/4021`).

## Why it matters
- Task authors reading the tool schema see `PATCH` is unsupported, get told to use `PUT`/`POST`, silently misdirect their agent.
- LLMs generating tool calls will refuse or hallucinate around PATCH depending on how they interpret the schema constraint.
- A well-behaved schema validator could reject PATCH calls the runtime would happily execute — inconsistency.

## Fix
Add `"PATCH"` to the `method` enum in `http_request.py`. Verify none of the existing packs' `required_actions` fail as a result.

## Acceptance
- Tool schema enum lists PATCH.
- Canonical test locks the enum ↔ implementation match: iterate the enum and assert each verb is handled by execute() without raising.

Sanjays2402:
Fixes #452.

Adds `PATCH` to the advertised HTTP request methods, matching the runtime's existing forwarding behavior. The schema regression test fails on `main` and passes with this change; all 19 `HTTPRequestTool` unit tests pass.

### DIFF
--- a/tests/unit/test_tool_builtins.py
+++ b/tests/unit/test_tool_builtins.py
@@ -298,6 +298,7 @@ class TestHTTPRequestTool:
         assert func["name"] == "http_request"
         params = func["parameters"]
         assert "method" in params["properties"]
+        assert "PATCH" in params["properties"]["method"]["enum"]
         assert "url" in params["properties"]
         assert "headers" in params["properties"]
         assert "json" in params["properties"]
--- a/tolokaforge/tools/builtin/http_request.py
+++ b/tolokaforge/tools/builtin/http_request.py
@@ -40,7 +40,7 @@ class HTTPRequestTool(Tool):
                     "properties": {
                         "method": {
                             "type": "string",
-                            "enum": ["GET", "POST", "PUT", "DELETE"],
+                            "enum": ["GET", "POST", "PUT", "PATCH", "DELETE"],
                             "description": "HTTP method",
                         },
                         "url": {


CiroGamboa:
## Review

### Overview
Adds `PATCH` to the `method` enum in `HTTPRequestTool.get_schema()` so the advertised schema matches the runtime's actual behavior. The runtime forwards any `method` string straight to `httpx.request`, so PATCH already worked at execution — the schema was the only thing lying.

### Correctness
- Fix is right. `httpx.request(method=method, …)` places no constraint on the verb, so this is a pure schema/runtime alignment. No behavior change for GET/POST/PUT/DELETE.
- Enum position (`GET, POST, PUT, PATCH, DELETE`) mirrors the conventional CRUD grouping — fine.

### Test coverage
- New assertion in `test_schema_structure` is the right regression guard and, per the PR body, fails on `main`.
- No new execution-path test for PATCH, but existing tests already exercise the httpx passthrough for other verbs and PATCH takes the same code path — a runtime test would add little.

### Suggestions (minor, non-blocking)
- **Docstring drift** in the same schema: `json` is described as *"Optional JSON body for POST/PUT"* and `data` as *"Optional form data for POST/PUT"* (`tolokaforge/tools/builtin/http_request.py:53,57`). With PATCH now advertised, worth extending both to *"POST/PUT/PATCH"* so the model knows a body is legal on PATCH too — otherwise it may omit one.
- Consider whether `HEAD`/`OPTIONS` should be advertised too — the runtime supports them equally. Out of scope for this fix; a follow-up if #452's motivation extends beyond PATCH.

### Risk
Effectively zero. Enum-widening is backward compatible (no existing valid payload becomes invalid) and the runtime already handled the verb.

**Verdict:** LGTM — merge as is, or fold in the two-word docstring tweak on `json`/`data` if you want the schema fully consistent.

PR #512, opened 2026-07-17
fix(adapter): fail validation conversions

### LINKED ISSUE #487: `tolokaforge adapter convert --validate` reports invalid tasks but exits 0
## Summary

`tolokaforge adapter convert --validate` runs a validation pass over the converted native bundles (`_validate_converted` in `tolokaforge/cli/adapter_commands.py:130`), but the result of that pass does **not** influence the process exit code. The command exits `0` even when every converted task fails schema validation.

## Reproduction

A conversion adapter that emits a `NativeTaskBundle` whose `task_config` omits a required `TaskConfig` field (e.g. `description`) produces bundles that fail `load_task_yaml`:

```
Validating converted output…
  ✗ invalid: alpha: 1 validation error for TaskConfig
description
  Field required
  Validation: 0 valid, 2 invalid

Converted 2 tasks (0 errors)
```

Exit code: `0`.

The exit code is driven solely by the per-task *conversion* error count (`adapter_commands.py:126-127`); the *validation* pass (`_validate_converted`) only prints a report and never signals failure.

## Why it matters

`--validate` exists to gate broken conversions. A CI pipeline or task author relying on `convert --validate` to fail on schema-invalid output gets a green exit while shipping unusable task bundles. This is a fail-fast violation (AGENTS.md Core Rule 1 / Code Standard 1: "surface failures explicitly; do not add fallbacks that hide errors").

## Suggested fix

When `do_validate` is set and `_validate_converted` finds any invalid task, exit non-zero (return the invalid count from `_validate_converted` and fold it into the final `SystemExit`). Keep the current behaviour when `--validate` is not passed.

## Discovery context

Found while planning the CI smoke test for `adapter convert` via a real entry-point adapter (#49). The smoke test works around this by asserting on the validation report text ("0 invalid") rather than the exit code — but the exit-code behaviour itself is the latent bug.

Dodothereal:
# fix(adapter): fail validation conversions

## Summary

`tolokaforge adapter convert --validate` reported invalid converted tasks but still exited successfully. The command now treats validation failures as command failures while preserving the existing exit behavior when `--validate` is omitted.

Fixes #487

## Changes

- Return the invalid-task count from the converted-output validation pass.
- Exit nonzero when conversion or requested validation finds errors.
- Cover a real malformed converted bundle through the CLI unit test.

## Testing

Ran in the required sandbox:

- `python -m pip install 'uv==0.9.5' && uv run pytest tests/unit/test_adapter_convert_cli.py -v` — 3 passed.
- `python -m pip install 'uv==0.9.5' && uv run ruff check tolokaforge/cli/adapter_commands.py tests/unit/test_adapter_convert_cli.py && uv run ruff format --check tolokaforge/cli/adapter_commands.py tests/unit/test_adapter_convert_cli.py` — passed; 2 files already formatted.

---
### AI assistance disclosure

This contribution was produced by an autonomous AI coding agent (Claude Code) that @Dodothereal operates and monitors. @Dodothereal is accountable for it, will address review feedback promptly, and will close this PR immediately if this kind of contribution is unwelcome in this project. Commits carry an `Assisted-by: Claude Code` trailer.

### DIFF
--- a/tests/unit/test_adapter_convert_cli.py
+++ b/tests/unit/test_adapter_convert_cli.py
@@ -40,7 +40,11 @@ class _StubConversionAdapter(NativeAdapter):
 
     def convert_to_native(self, task_id: str) -> NativeTaskBundle:
         return NativeTaskBundle(
-            task_config={"name": f"Task {task_id}", "category": "tool_use"},
+            task_config={
+                "name": f"Task {task_id}",
+                "category": "tool_use",
+                "description": "A valid converted task.",
+            },
             grading_config={"combine": {"method": "weighted", "pass_threshold": 1.0}},
             fixtures={"tools": [{"name": "noop"}]},
             metadata={"source_adapter": "stub"},
@@ -59,6 +63,16 @@ class _NoSharedAdapter(_StubConversionAdapter):
     """Conversion adapter that does NOT emit shared resources (default no-op)."""
 
     write_shared_resources = BaseAdapter.write_shared_resources
+
+
+class _InvalidConversionAdapter(_StubConversionAdapter):
+    """Conversion adapter that produces one invalid native task bundle."""
+
+    def convert_to_native(self, task_id: str) -> NativeTaskBundle:
+        bundle = super().convert_to_native(task_id)
+        if task_id == "t2":
+            bundle.task_config.pop("description")
+        return bundle
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


CiroGamboa:
Thanks for the contribution, @Dodothereal, and for the clear AI-assistance disclosure.

Closing as a duplicate — #487 was fixed on `main` earlier today by #494 (@Sanjays2402), merged at 2026-07-20T10:52:22Z. Both PRs make the same behaviour change: `_validate_converted` returns the invalid count and the CLI exits nonzero when `--validate` finds errors.

The one substantive difference is that #512 improves the shared `_StubConversionAdapter` to be a genuinely valid fixture (adds `description`) and introduces a separate `_InvalidConversionAdapter` for the failure case, whereas #494 leaves the shared stub as-is (implicitly invalid) and asserts the "0 valid, 2 invalid" path. The stub-hygiene bit is a nice improvement in isolation — if you'd like to open a small follow-up against current `main` that only adds `description` to the shared stub, that'd be welcome. Otherwise, no further action needed here.

Issue #519, opened 2026-07-17
Judge/loop unit tests bypass _convert_messages, hiding provider message-shape bugs


azorej:
## Context

Discovered while planning #515 (judge `submit_report` retry sent no `role=tool` result → OpenAI-family 400 on the retry turn).

The judge orchestration tests (`tests/unit/grading/test_judge.py`) drive `run_rubric_judge` with a `ScriptedClient` that satisfies the `LoopLLMClient` `generate` seam directly and **never calls `LLMClient._convert_messages`**. That is correct for testing *our orchestration* (termination, retry, gating, usage), but it means the produced `messages` list is never validated against the provider wire contract — specifically the OpenAI/Azure rule that *an assistant message with `tool_calls` must be immediately followed by a `role=tool` result for each `tool_call_id`*.

This is exactly why #515 stayed latent and pre-existing: the malformed retry sequence (`assistant(tool_calls) → system → user`, no tool result) round-trips fine through the scripted client but 400s against a real OpenAI-family provider.

## Proposal

Add a small reusable assertion/helper (e.g. `assert_provider_valid_message_sequence(messages)`) that verifies the tool_call ↔ tool_result adjacency invariant, and apply it in the loop/judge unit tests wherever a `messages` list is built across turns. This turns a whole class of wire-shape regressions into unit failures without needing live spend.

#515 adds a narrow, judge-retry-specific version of this check inline; this issue tracks generalizing it into a shared helper reused across loop tests.

**Priority:** P3

Issue #468, opened 2026-07-16
[umbrella] Judge scoring integrity — verdict consistency, judge customization, replay


azorej:
## Scope

Umbrella for milestone **Judge scoring integrity** (#24). Driver: a 2026-07-13 field report (GitHub Actions run 29240785466, judge `anthropic/claude-opus-4.8` @ temp 0.0, tolokaforge v0.7.0) found 2/68 trials where the `submit_report` per-criterion boolean contradicts its own justification — the judge reverses its conclusion *while writing* the justification, but the boolean was already emitted (the schema orders verdict before justification, and generation is autoregressive). Silent false negatives; with `required: true` criteria one flip fails a whole trial. Root cause confirmed from logged tool-call arguments; a 192-replay experiment (2026-07-16, 3 judge models × 4 schema variants) validated the fix approach — details in #449.

**Priority:** P1 (carried by #449; the rest of the milestone is P2)

## Budget and providers — explicit expectation

Implementation, experimentation, testing, and canonical-output generation for this milestone are expected to **use real LLM providers and spend real money** (OpenRouter/Anthropic/etc.). This covers: judge smoke runs on strong + weak models (#449), canonical/golden output generation and regeneration (#449), replay A/B validation (#451), and any ad-hoc experiments needed to settle open questions. API keys live in the repo's `.env` file and are read via `SecretManager` (the only sanctioned secret path — AGENTS.md). Do not substitute mocked/synthetic LLM output where the ticket calls for real model behavior.

## Issues and recommended execution order

| Order | Issue | Priority | Why this position |
|---|---|---|---|
| 1 | #449 verdict must follow and match its justification | P1 | Smallest blast radius, fixes the live correctness bug; everything else is hardening around it. Ships with a live smoke run — does not need replay. |
| 2 | #465 project/task-level judge customization (`disable_knowledge_search`) | P2 | Small config seam; lands the judge-tool gating provenance that #451 must honor when replaying. |
| 3 | #451 offline judge mode (replay) | P2 | Largest scope; benefits from both predecessors (replays the new schema, honors recorded gating). Also the long-term validation vehicle for any future judge change. |

All three are independently shippable; the order minimizes rework, not hard dependencies.

## Decisions resolved during scoping (recommended, overridable at plan review)

[...]

Issue #449, opened 2026-07-16
fix: rubric judge submit_report — verdict must follow and match its justification (silent false-negative flips)


azorej:
## Context

A production benchmark run (GitHub Actions run 29240785466, 2026-07-13, config `ots_08_travel_internal/smoke_comm_rubrics_opus_4.8`, judge `anthropic/claude-opus-4.8` @ temp 0.0, tolokaforge v0.7.0) produced 2 trials out of 68 where a per-criterion boolean in `submit_report` contradicts its own justification **inside the same tool call**: the justification text re-examines the evidence mid-writing, reverses to "criterion is in fact MET… marking as not met would be an error", but `met: false` is what gets submitted. Nothing validates consistency, so these are **silent false negatives** — found only by manually reading justifications in `grade.yaml`. With `required: true` criteria this class of error flips an entire trial.

**Root cause (confirmed from the logged tool-call arguments in `judge_trajectory.yaml`):** the rubric-derived `submit_report` schema orders each criterion's verdict field *before* its justification field, and models generate tool arguments in schema property order. The boolean token is therefore emitted **before any reasoning is written** and cannot be revised when the model changes its conclusion while writing the justification (generation is autoregressive). In one trial even the overall `reasons` field — generated last — says "I am marking it true" while the flag, emitted thousands of tokens earlier, is `false`. The mismatches cluster on exactly the borderline criteria where the verdict matters most.

Evidence bundle: `judge_mismatch_artifacts.tar.gz` (trials AE-BDG-002_1, AE-BDG-003_0; criterion `no_internal_references`; in both, the reversed textual verdict is the correct one).

**Priority:** P1

## Deliverables

- [ ] The generated `submit_report` schema orders each criterion's justification field **before** its verdict field, so the verdict token is generated conditioned on the full written reasoning (reason-then-answer).
- [ ] Each justification is required (via the schema field description) to end with a structured final line — `VERDICT: MET` / `VERDICT: NOT MET` for binary criteria, `SCORE: <value>` for graded ones — and the verdict/score field description states it must match that line.
- [ ] `submit_report` validation rejects a call whose trailing marker is missing, or contradicts the submitted boolean (binary) / differs from the submitted score beyond a small tolerance (graded). Rejection rides the existing bounded re-prompt retry; on retry exhaustion the judge errors loudly (existing errored-grade path) — never a silently wrong grade.
[...]

PR #518, opened 2026-07-17
fix(grading): rubric judge verdict must follow and match its justification

azorej:
## Summary

- The rubric judge's generated `submit_report` schema now orders each criterion's justification field **before** its verdict field, so the verdict token is generated conditioned on the full written reasoning (reason-then-answer) instead of being emitted before any reasoning exists — the root cause of the 2026-07-13 field report's silent verdict/justification flips (2/68 trials).
- Every justification must end with a structured marker (`VERDICT: MET` / `VERDICT: NOT MET` binary, `SCORE: <value>` graded); `parse_submit_report` rejects a missing or contradicting marker (`VerdictConsistencyError`), riding the existing bounded retry to a loud ERRORED on exhaustion — never a silently wrong grade.
- Consistency rejections are counted per trial (`consistency_rejections` in `judge_usage`, additive through the proto/grade.yaml contract), so a flip-flopping judge model is visible in run reports.

## Design choices

- **Field reordering is the primary fix, the marker check the backstop**: JSON-Schema property order is non-normative, so ordering alone can't be trusted for every provider; the anchored marker check makes consistency machine-verifiable (validated by the 2026-07-16 192-replay experiment: zero verdict changes attributable to reordering, zero marker/flag conflicts).
- **Anchored structured marker over free-text sentiment parsing**: the failing justifications contain both "agent leaked identifiers" and "criterion is in fact MET" — only a mandatory final-line marker makes the check a trivial comparison. Live evidence during implementation loosened the anchor to "last `VERDICT:`/`SCORE:` occurrence within the final non-empty line" because real models append the marker inline to the closing sentence; a left lookbehind prevents "subscore:"-style false matches.
- **Telemetry strictly additive** (default 0 on every hop: sink → proto field 8 → runner → Pydantic `extra="forbid"` mirror → grade.yaml), coordinated with #465 which takes the next proto field number.
- **Real judge output only in test data**: rejection cases are frozen v0.7.0 field-report captures (verbatim logged tool-call arguments — the bug evidence, never regenerated); well-formed cases are a live-captured, regenerable golden payload plus a deterministic schema snapshot. No hand-crafted synthetic judge output.

## Concepts introduced

[...]

PR #520, opened 2026-07-17
fix(grading): judge submit_report retry answers every tool call with a provider-valid tool result

### LINKED ISSUE #515: fix: judge submit_report re-prompt retry sends no tool result — OpenAI-family providers 400 on the retry turn
## Context

Found during #449 Stage 3 live acceptance (milestone #24). When the rubric judge's `submit_report` is rejected by validation (schema error, or the new verdict-consistency rejection from #449), the retry loop appends the corrective `user` re-prompt directly after the assistant's `submit_report` tool-call message **without an intervening `role=tool` result for that call**. OpenAI/Azure-family providers reject the next generation with `400 "No tool output found for function call <id>"`, so the bounded re-prompt degrades to API_ERROR → ERRORED instead of giving the judge a genuine second attempt.

Latent and pre-existing (malformed-schema rejections hit the same path), but #449's marker-consistency rejection class makes it hot: a weak judge that misses the marker format on the first try should recover via retry (~12% first-try miss on the weakest tested model in the 2026-07-16 experiment), and on OpenAI-family providers it currently cannot.

Fail-loud is preserved (the trial errors, never silently mis-scored) — what's broken is the recovery path, which costs trials that would have recovered and undermines the documented "rejection rides the bounded re-prompt retry" contract (docs/GRADING.md).

## Fix sketch (from #449 Stage 3 diagnosis)

Capture the `submit_report` tool_call_id in the termination handling (`_SubmitReportTermination` in `tolokaforge/core/grading/judge.py`); on rejection, append a `role=tool` result carrying the validation-error text for that tool_call_id **before** the corrective user turn (or fold the corrective text into the tool result itself). Lock with a unit test driving a scripted rejection→retry sequence asserting the message-shape contract, plus a live integration case on an OpenAI-family judge model that misses then recovers.

## Acceptance criteria

- A rejected `submit_report` is followed by a `role=tool` result for the rejected call id, then the re-prompt; the next generation succeeds against OpenAI-family providers (no 400).
- Retry-exhaustion behaviour unchanged (ERRORED, loud).
- `consistency_rejections` counting (#449) unaffected.

**Priority:** P2

azorej:
## Summary

- A rejected rubric-judge `submit_report` is now answered with a `role=tool` result for its `tool_call_id` — placed immediately after the assistant tool-call message — instead of a bare corrective `role=user` turn with no tool result. OpenAI/Azure-family judge models now genuinely recover on the bounded retry instead of erroring with `400 "No tool output found for function call <id>"` → API_ERROR → ERRORED.
- Every `tool_call_id` on the terminating assistant turn is answered: the `submit_report` id carries the validation error + corrective instruction (folded), and any sibling tool call the judge emitted in the same turn gets an honest "not executed: submit_report ended the turn" note (termination fires before tool execution, so the sibling genuinely never ran).
- The loop's trailing "submit_report received; judge terminating." system message — false on a retried run, and the exact thing that breaks tool-result adjacency — is removed from the retry tail.

## Design choices

- **Answer all tool ids, not just `submit_report`'s**: nothing forbids the judge from parallel-calling a read tool alongside `submit_report` (`tool_choice="auto"`, no `parallel_tool_calls=False`, and providers don't uniformly honor that flag anyway); a single-id fix would leave the sibling unanswered and reproduce the same 400 class latently.
- **Adjacency is the wire contract, verified live**: a four-shape probe on a real OpenAI-family endpoint showed `assistant → system → tool → user` still 400s — only a tool result *immediately* after the assistant call is accepted. The fix keys off the captured call id (position-independent), never a positional pop.
- **Corrective folded into the tool result, no separate `role=user` turn**: keeps the retry sequence valid on Anthropic too (a tool result renders as a user-role `tool_result` block; a following user message would create consecutive user turns).
- **Fail-loud on invariant violations**: a missing captured call id on the rejection path raises `RuntimeError` — no silent fallback to the old broken shape.

## Concepts introduced

None — mechanical fix to the judge retry message contract. (`_SubmitReportTermination` now records the terminating call's id alongside its arguments; a module helper performs the retry-tail surgery.)

## Live acceptance evidence (real provider, real spend)

`test_rubric_judge_live_recovers_through_forced_retry` — model `openrouter/openai/gpt-4.1-mini`; the harness's validation decision (not the model output) is forced to reject exactly once:

- First call rejected: `Judge submit_report invalid; re-prompting (attempt=1, error=forced rejection to exercise the retry path (test only))`
- Retry completed, no 400: `Judge completed (score=1.0, gate_failed=False, failed_required=[])`
- `consistency_rejections == 0` (generic rejection, #449 counting unaffected) — **1 passed in 11.27s**
- No-key behaviour verified: with `.env` absent the case skips cleanly (`1 skipped`), not errors.

Red-before/green-after: with pre-fix `judge.py`, all four new unit locks fail (unanswered `tool_call_id`, intervening system message) and the live retry path 400s before a second validation; on this branch everything passes.

## Plan

# Plan: Judge submit_report re-prompt retry must send a tool result (#515)

Issue: #515 (umbrella #468, milestone 24 "Judge scoring integrity")
Branch: `fix/judge-retry-tool-result` (off `feat/judge-scoring-integrity`)

## Context

When the rubric judge's `submit_report` is rejected by validation (a schema error, or the
`VerdictConsistencyError` from #449), the retry loop in
`tolokaforge/core/grading/judge.py` (`run_rubric_judge`, lines 610-623) appends a corrective
`role=user` re-prompt directly after the assistant's `submit_report` tool-call message **with no
intervening `role=tool` result for that call**. `submit_report` is a *terminal* tool: the
`_SubmitReportTermination` policy captures its arguments and stops the loop, so — unlike every
other judge tool — it is never executed and never produces a tool result. OpenAI/Azure-family
providers 400 the next generation (`"No tool output found for function call <id>"`), so the
bounded re-prompt degrades to API_ERROR → ERRORED instead of giving the judge a genuine second
attempt.

Reproduced during discovery (2026-07-17, on this branch):

- **Deterministic (no spend):** driving the real `run_rubric_judge` retry path with a scripted
  client that captures the messages handed to the retry generation yields the tail
  `[user, assistant(tool_calls=call_1_0), system("submit_report received; judge terminating."),
  user(corrective)]` — the assistant `submit_report` call `call_1_0` has **no** answering
  `role=tool` message.
- **Live (gpt-4o-mini, minimal spend):** four message shapes through `litellm.completion`:
  | sequence | result |
  |---|---|
  | `assistant(tool_calls) → system → user` (current) | **400** "must be followed by tool messages responding to each tool_call_id: call_X" |
  | `assistant → system → tool → user` (tool result *after* the system message) | **400** — same error |
  | `assistant → tool → system → user` (tool result *immediately* after the call) | **OK** |
  | `assistant → tool(corrective folded)` (no user turn) | **OK** |

The live probe settles the fix contract: the `role=tool` result must be **immediately adjacent**
to the assistant `submit_report` message — a plain append at the tail (after the loop's
"terminating" system message) does **not** satisfy the provider.

Fail-loud is preserved today (the trial errors, never silently mis-scored); what is broken is the
recovery path, which costs trials that would have recovered and undermines the documented
"rejection rides the bounded re-prompt retry" contract (`docs/GRADING.md` §Fail-loud;
`docs/RUBRIC_GRADING_DESIGN.md` Structured-output row). #449's marker-consistency rejection class
makes this hot: a weak judge that misses the marker format on the first try (~12% first-try miss on
the weakest tested model, 2026-07-16 experiment) should recover via retry, and on OpenAI-family
providers it currently cannot.

## Goal

After a rejected `submit_report`, the messages fed to the next judge generation form a
**provider-valid sequence**: **every `tool_call_id` on the terminating assistant message is
answered by an adjacent `role=tool` result**. The `submit_report` id's result carries the
validation-error text plus the corrective instruction; any sibling tool call on that same turn
(the judge may emit read/search calls alongside `submit_report` — see the sibling risk below) is
also answered, so no id is left unanswered. No non-tool message separates the assistant call from
its (contiguous) tool results. The next generation succeeds against OpenAI/Azure-family providers
(no 400) and the judge gets a genuine second attempt.

The invariant the fix guarantees and the unit test locks is the strong form — *every*
`tool_call_id` in *each* assistant message is answered by an adjacent `role=tool` result — not the
weaker "the assistant message is followed by *a* tool result," which a single trailing result would
satisfy while a sibling id stays unanswered.

Retry-exhaustion behaviour (ERRORED, loud) and `consistency_rejections` counting (#449) are
unchanged.

### Why sibling tool calls on the terminating turn are in scope

`_SubmitReportTermination` fires the instant `submit_report` appears in the assistant's
`tool_calls` — `should_terminate` runs at `loop.py:301`, *before* any tool executes at `loop.py:307`
— and the loop has already appended the full assistant message (all `tool_calls`) at `loop.py:299`.
The judge runs with `tool_choice="auto"` and no `parallel_tool_calls=False`, and the system prompt
(`judge.py:260-272`) says "call submit_report" but never "as the only call in the turn". The
umbrella's own experiment data notes judges attempt read/search tool calls in most samples when
tools are offered, so a turn carrying `submit_report` **plus** a `query_db`/`search_kb`/`read_file`
call is plausible, not hypothetical. If the retry answers only the `submit_report` id, the sibling
id is unanswered and OpenAI/Azure still 400 with the exact `#515` failure class — just a sibling
case that would stay latent. Hence the fix answers **all** ids on the terminating turn.

## Non-goals

- No new retry knob — the existing bounded `DEFAULT_SUBMIT_REPORT_RETRIES = 2` stays.
- No change to `submit_report` schema, marker/consistency validation, or aggregation (#449 territory).
- No change to the shared `ToolCallingLoop` termination protocol or to the agent's (non-judge) loop
  path — this is judge-specific because `submit_report` is the only terminal tool captured-but-not-
  executed.
- No change to `grade.yaml`, the proto, the CLI, or any config/authoring surface.
- No generalized loop-test wire-contract helper — filed as #519 (P3).

## Compatibility surfaces touched (and why each is / isn't a migration)

- **Judge retry message construction** — internal judge↔LLM contract, regenerated per run, never
  persisted, never author-facing. Internal only; no migration.
- **`judge_trajectory.yaml` audit transcript** — the retry `role=tool` result now appears in the
  serialized transcript (`_serialize_judge_transcript` already handles `tool_call_id`). The
  transcript is a free-form list of role/content dicts (no fixed schema, no `extra="forbid"`
  model), so this is an additive observability improvement, **not** a schema migration. Worth a
  one-line CHANGELOG Fix entry (user-visible: OpenAI-family judges no longer error on retry), but
  no output-contract migration.

### Stage 1: Retry sends a provider-valid tool result (the fix + unit lock)

- **Contract** (`tolokaforge/core/grading/judge.py`):
  - `_SubmitReportTermination` captures the terminating `submit_report` call's **id** alongside its
    arguments. Add a field `captured_call_id: str | None` (default `None`, `init=False`); in
    `__call__`, when the matching `submit_report` tool call is found, set
    `captured_call_id = tc.id` next to `captured_args = dict(tc.arguments or {})`. Reset it to
    `None` at the same points `captured_args` is reset for a retry.
  - On a `SubmitReportValidationError` in `run_rubric_judge`, before the next `loop.run`, the retry
    tail MUST be a provider-valid sequence: **every `tool_call_id` on the terminating assistant
    message is answered by an adjacent `role=tool` result**, with the tool results contiguous and
    immediately following that assistant message. Use `captured_call_id` to locate the terminating
    assistant message (it is the message whose `tool_calls` contain that id), then answer **all** of
    its `tool_calls`:
    - the `submit_report` id → `Message(role=TOOL, tool_call_id=<submit id>, content=<rejection
      reason + corrective instruction>)`;
    - each **sibling** id (any non-`submit_report` tool call the judge emitted in the same turn) →
      a `Message(role=TOOL, tool_call_id=<sibling id>, content=<"not executed: submit_report ended
      the turn; gather evidence with your read tools *before* calling submit_report">)`. The sibling
      was never executed (termination fired before tool execution), so this is an honest "not run"
      result, not a fabricated tool output — and it nudges the judge to read first on the retry.
    No non-tool message may sit between the assistant call and its tool results.
    - **Recommended mechanism (implementer's call on the exact list surgery):** place the tool
      results immediately after the terminating assistant message, replacing the loop's trailing
      `"submit_report received; judge terminating."` system message — that message is *false* on a
      continued (retry) run and is exactly what breaks adjacency if left before the tool results. Do
      **not** emit a separate `role=user` corrective turn: folding the corrective into the
      `submit_report` tool result keeps the sequence valid on Anthropic too (a `role=tool` result
      renders as an Anthropic user-role `tool_result` block; a following `role=user` message would
      create two consecutive user turns), and the live probe confirmed the folded shape
      (`assistant → tool…`) is accepted by OpenAI.
    - The `submit_report` tool result preserves the current corrective wording (the validation error
      text + "Fix the issue and call submit_report again with a verdict and justification for every
      criterion.") so the judge still receives the same guidance.
  - **Solo vs parallel — chosen approach.** Answering all ids (option a) is preferred over asserting
    the terminating turn is always solo + failing loud on a sibling (option b): the umbrella data
    says siblings are plausible, so option (b) would convert a plausible real case into a hard error
    and lose recoverable trials; option (a) is provider-agnostic and barely more code. Not relying on
    `parallel_tool_calls=False` either — it is not honored by every provider (e.g. Gemini) and the
    `generate` seam does not expose it cleanly.
  - `captured_call_id` is guaranteed non-`None` on the rejection path (a rejection only happens
    after `captured_args` was set, which sets the id in the same branch). If it is `None` when a
    rejection is being handled, that is a harness invariant violation — fail loud (raise / error),
    do not silently fall back to the old user-turn shape.
- **Behaviour to lock:**
  - **unit** (`tests/unit/grading/test_judge.py`): a new test drives `run_rubric_judge` with a
    scripted client that **captures the `messages` list handed to the retry generation** (first
    rejected `submit_report`, then a valid one). Assert the **strong** adjacency invariant on that
    captured list via a small local helper: for *every* assistant message with `tool_calls`, *every*
    `tool_call_id` is answered by a `role=tool` message in the contiguous block immediately following
    it (no non-tool message interleaved). Additionally assert the `submit_report` id's tool result
    content contains the validation-error text. This test **fails on current code** (the intervening
    system message + missing tool result) and passes after the fix — the reproduction encoded as a
    lock. Cover both rejection classes: a schema rejection (`met` not a bool) and a
    `VerdictConsistencyError` (contradicting marker).
  - **unit** (`tests/unit/grading/test_judge.py`, **sibling case — this is the finding-1 lock**): a
    test whose terminating scripted step emits a sibling tool call **alongside** `submit_report` in
    the *same* step (e.g. `[("query_db", {...}), ("submit_report", <rejected args>)]` — `ScriptedClient`
    already supports multi-call steps, `test_judge.py:58-61`), the submit is rejected, then a valid
    submit follows. Assert the strong invariant holds on the retry messages — **both** the
    `submit_report` id **and** the sibling id are answered by adjacent `role=tool` results. This is
    the case a single-id fix would leave unanswered (→ 400 on a real provider); it must be red before
    the fix's all-ids handling and green after.
  - **unit** (regression, same file): the existing retry tests keep passing unchanged —
    `test_malformed_then_valid_recovers`, `test_consistency_rejection_then_valid_recovers_with_count_preserved`,
    `test_malformed_submit_report_reprompts_then_errors`,
    `test_consistency_rejection_reprompts_and_counts_per_attempt_then_errors`. In particular
    `consistency_rejections` counting is unchanged (assert the counts in those tests still hold),
    and retry exhaustion still yields ERRORED with no score.
  - **unit** (`tests/unit/grading/test_judge.py`, transcript): after a reject→recover run, the
    serialized `result.transcript` contains the injected rejection `role=tool` entry (a
    `tool_call_id` matching the rejected `submit_report` call) — the audit crumb for the retry
    cycle.
- **Compatibility:** internal only (retry message construction; regenerated per run, not persisted,
  not author-facing). Audit-transcript addition is additive/free-form (see Compatibility section).
- **Deliverable:** `_SubmitReportTermination` captures the `submit_report` call id; the retry path
  answers **every** `tool_call_id` on the terminating assistant message with adjacent `role=tool`
  results (`submit_report` → rejection+corrective, siblings → honest "not executed"); unit tests
  lock the strong adjacency invariant for both rejection classes **and** the sibling case, and
  confirm counting/exhaustion unchanged; CHANGELOG Fix entry.
- **Validation:**
  - `uv run pytest tests/unit/grading/test_judge.py -m unit`
  - `uv run pytest tests/unit/grading/test_rubric.py -m unit` (no regression from the id capture)
  - `uv run pytest tests/unit/test_judge_artifacts.py -m unit` (transcript serialization intact)
  - `uv run ruff check tolokaforge tests` + `uv run ruff format --check tolokaforge tests`
  - Reviewer checks: **every** `tool_call_id` on the terminating assistant message is answered by an
    adjacent `role=tool` result (not just `submit_report`'s) with no intervening non-tool message;
    the `submit_report` result carries the validation error + corrective wording and matches its id;
    sibling results are honest "not executed" notes, never fabricated tool output; `captured_call_id`
    missing on the rejection path fails loud (no silent fallback); `consistency_rejections` increment
    site and semantics untouched; no change to the shared loop or the agent path.
- **Doc updates:**
  - `docs/GRADING.md` §"Fail-loud: the ERRORED status" — rewrite the re-prompt paragraph so it reads
    as if the current behaviour is the only behaviour: a rejected `submit_report` is answered with a
    tool result carrying the rejection reason, then the judge is re-prompted within the bounded
    budget; on exhaustion the trial rides the ERRORED path. State the wire fact plainly (the
    rejection is delivered as the tool result for the rejected call) without "previously/now"
    history.
  - `docs/RUBRIC_GRADING_DESIGN.md` — the Structured-output / Fail-loud rows that mention "bounded
    re-prompt": add that the rejection is returned to the judge as the `submit_report` call's tool
    result (so the re-prompt is a provider-valid tool-call/tool-result cycle). Current-state wording
    only; no migration note.
  - `CHANGELOG.md` — Fix entry under the working version: rubric-judge `submit_report` re-prompt now
    returns the rejection as a tool result, so OpenAI/Azure-family judge models recover on retry
    instead of erroring with a 400.

### Stage 2: Live acceptance — OpenAI-family judge recovers through a real retry (no 400)

- **Contract:**
  - A new integration case in `tests/integration/test_rubric_judge_live.py` (gated on
    `OPENAI_API_KEY` / `OPENROUTER_API_KEY` via `SecretManager`; auto-skips without keys; real,
    minimal spend) that drives a **real OpenAI-family judge** through the retry path and asserts the
    second live generation **succeeds (no 400)** and the run reaches COMPLETED.
  - **Deterministic, honest provocation of the retry** (resolves the issue's "no scripted-then-live
    hybrid that fakes model output" constraint): the model's `submit_report` output on both turns is
    **100% real** — nothing about the model's tool call is fabricated. The retry is forced by making
    the harness's *validation decision* reject the first real `submit_report` exactly once by
    monkeypatching the name **as bound in the judge module** — `judge.py:58` does
    `from ...rubric import parse_submit_report`, so `run_rubric_judge` calls its own module-level
    name (`judge.py:589`); the patch target is
    `tolokaforge.core.grading.judge.parse_submit_report`, **not** `rubric.parse_submit_report`
    (patching the latter is a no-op). The patched shim raises `SubmitReportValidationError` on its
    first call, then delegates to the real implementation, then accepts the second. This exercises the
    real path a
    genuine weak-model miss would take (real model → first submit_report → rejection → repaired
    retry sequence → live provider round-trip → recovery), and the *only* thing forced is "was this
    output rejected?", which is orthogonal to the 400 bug under test. The assertion that matters —
    the live provider accepts the repaired retry sequence — is fully real.
  - Assert: the judge run COMPLETES (status `COMPLETED`, a numeric score), the second generation did
    not raise a `BadRequestError` / 400, and the judge transcript contains the injected rejection
    `role=tool` result. `consistency_rejections` behaviour is unaffected (the forced rejection uses a
    generic `SubmitReportValidationError`, not a `VerdictConsistencyError`, so the consistency
    counter stays 0 — matching the "counting unaffected" acceptance criterion).
  - Quote the run output in the PR description (AGENTS.md model-PR discipline analogue): the model
    id, that the first call was rejected, and that the retry generation returned a completion rather
    than a 400.
- **Behaviour to lock:**
  - **integration** (`tests/integration/test_rubric_judge_live.py`): real OpenAI-family judge, one
    forced rejection, retry generation succeeds (no 400), run COMPLETES.
- **Compatibility:** test-only addition. No migration.
- **Deliverable:** a key-gated live integration test proving an OpenAI-family judge recovers through
  the retry (no 400); output quoted in the PR.
- **Validation:**
  - `scripts/with_env.sh uv run pytest tests/integration/test_rubric_judge_live.py -m integration`
    (real key present) — capture and quote output.
  - Confirm the test auto-skips cleanly when the key is absent (`uv run pytest ... -m integration`
    with no key → skipped, not errored).
  - Reviewer checks: no fabricated model output (the `submit_report` tool call is the real model's;
    only the reject/accept decision is forced); the assertion targets the live provider accepting the
    repaired sequence; test is key-gated and skips without keys.
- **Doc updates:**
  - `tests/README.md` — one line noting the live rubric-judge retry-recovery case and that it forces
    exactly one validation rejection to exercise the retry path deterministically (real model
    output, forced reject decision).

## Discovered issues

- **Fix in this PR:** The loop-appended `"submit_report received; judge terminating."` system
  message is *false* on a continued (retry) run and is the message that breaks tool-result
  adjacency; Stage 1 removes it from the retry tail as part of the fix (not a separate change).
- **Filed as issues:**
  - **#519** — judge/loop unit tests bypass `_convert_messages`, so the provider message-shape
    contract (tool_call ↔ tool_result adjacency) is never validated in unit tests; this is why #515
    stayed latent. Proposes a shared `assert_provider_valid_message_sequence` helper. Stage 1 adds a
    narrow inline version of this check for the judge retry path; #519 tracks generalizing it. P3.

## Risks / open questions

- **Anthropic judge on the retry path.** The strong judge in #449's live acceptance is Anthropic.
  Folding the corrective into the `role=tool` result (rather than a following `role=user` turn) is
  the deliberate choice that keeps the retry sequence valid on Anthropic (no consecutive user
  turns). The Stage 2 live case is OpenAI-family (that's where the 400 lives); the folded shape is
  what protects Anthropic by construction. If the reviewer wants belt-and-suspenders, the Stage 2
  test can be parametrized to also run an Anthropic judge through the forced retry — cheap, same
  mechanism — but it is not required to close the OpenAI-family bug.
- **Coupling to loop message-append order.** The fix must place the tool results immediately after
  the terminating assistant message. Expressing this as "locate the assistant message bearing
  `captured_call_id` and answer every one of its `tool_calls` immediately after it" is more robust
  than "pop the last message" (which assumes the loop always appends exactly one trailing system
  message). The implementer should key off the captured call id to find the message, then iterate
  its `tool_calls` — not a positional assumption. The Stage 1 strong-invariant unit tests (including
  the sibling case) guard this regardless of mechanism.
- **Sibling tool calls on the terminating turn (finding 1).** A judge may emit read/search calls
  alongside `submit_report` in the terminating turn (umbrella data: judges attempt read/search calls
  in most samples). The fix answers **all** ids, sibling results being an honest "not executed:
  submit_report ended the turn" note (the sibling genuinely never ran — termination fires before
  tool execution — so this is not fabricated tool output). Chosen over "assume solo + fail loud"
  because siblings are plausible and (a) recovers those trials while (b) would hard-error them.
- **`captured_call_id` always present on rejection.** Verified from the code path: a rejection only
  occurs after `captured_args` was set, and the id is set in the same branch. The fix treats a
  missing id on the rejection path as a fail-loud harness bug, not a fallback — consistent with
  AGENTS.md rule 1.

## Discovered issues

- Filed: #519 — judge/loop unit tests bypass `_convert_messages`, so provider message-shape contracts were never unit-validated (why this bug stayed latent); this PR adds a narrow strong-adjacency helper for the judge retry path, #519 tracks generalizing it.
- Fixed in this PR: the false "judge terminating" system message on retry tails (part of the fix, it broke adjacency).

## Test plan

- Unit: `uv run pytest tests/unit/grading/test_judge.py -m unit` (31 passed) — strong adjacency invariant on captured retry messages for both rejection classes, sibling-call case, second-retry surgery shape-lock, transcript audit crumb, retry-exhaustion + `consistency_rejections` regressions unchanged
- Integration (key-gated, real spend): `scripts/with_env.sh uv run pytest tests/integration/test_rubric_judge_live.py -m integration` — forced-retry recovery above
- Post-merge: judge unit suite on the integration branch

Closes #515


azorej:
hygiene-review failure is the known repo-wide infra breakage (#425 / #398, action dies at init in ~25s, $0 spent) — confirmed with a re-run on PR #518 earlier today. Merging on the PR's own green lanes (lint ✅, test-smoke ✅).

Issue #545, opened 2026-07-20
Runtime banner reads deprecated orchestrator.runtime field, prints 'None' on the task-driven default path


CiroGamboa:
## Context

Discovered while planning #536. `tolokaforge/cli/main.py:270` feeds the runtime banner from the **deprecated** operator-override field:

```python
_print_runtime_banner(
    console=console,
    runtime_choice=run_config.orchestrator.runtime,   # Literal[...] | None, default None
    source="cli-flag" if runtime is not None else "config",
)
```

Backend selection is task-driven (`Orchestrator._select_backend_from_tasks`, `orchestrator.py:644`); `orchestrator.runtime` defaults to `None` and is a deprecated, retiring override. On the default path (no `--runtime`, no `orchestrator.runtime` in the config) `runtime_choice` is `None`, so `_print_runtime_banner` (`main.py:53-77`) resolves `backend_class = {...}.get(None, None) = None` and `isolation_line = "unknown"`, printing:

```
Runtime backend: None (None, from config)
  → unknown
```

The banner does not name the backend the orchestrator actually resolves and runs (`SharedStackRuntimeBackend` / `PerTrialRuntimeBackend`).

## Why out of scope for #536

#536's acceptance criteria require the banner to behave **identically to today**; fixing it changes observable CLI output. #536's registry rewrite lives downstream of the banner in `Orchestrator._construct_runtime_backend`, so the banner is untouched by that work.

## Suggested fix

Print the banner from the orchestrator's resolved choice (`_resolve_effective_runtime_choice()` / the constructed backend's `type(...).__name__` + `isolation_mode`), with source ∈ {`cli-flag`, `config-override`, `task-driven`}. Requires resolving the backend before printing (or moving the banner into the orchestrator).

Priority: P3.

CiroGamboa:
Closing as obsolete. `_print_runtime_banner` (`tolokaforge/dx/cli/main.py:132`) now branches on `runtime_choice is None` and prints 'Runtime backend: auto (task-driven, resolved at orchestrator boot)' with the candidate-backend line — the 'Runtime backend: None (None, from config) → unknown' shape the ticket flags is no longer producible.

Issue #544, opened 2026-07-20
Adapter entry-point registry does silent last-wins on duplicate names


CiroGamboa:
## Context

Discovered while planning #536 (entry-point Protocol registries). ADR-0019 § Decision Drivers mandates fail-loud for the new plugin registries: *"A duplicate registration ... must surface loudly ... never silent last-wins, never a swallowed import (AGENTS.md Core Rule)."*

The pre-existing **adapter** registry does the opposite:

- `tolokaforge/adapters/__init__.py:50-58` — `_discover_adapters()` does `adapters[ep.name] = ep.load()` inside the loop. Two installed distributions declaring the same adapter name under `tolokaforge.adapters` silently overwrite (last-wins), no detection, no warning. AGENTS.md Core Rule 1 (surface failures explicitly) is violated for this collision case.
- The same method swallows import errors into `_FAILED_ADAPTERS` and only re-raises if that specific adapter is later requested via `get_adapter`. (This half is arguably intentional graceful-degradation for optional adapter plugins — the duplicate-name silent overwrite is the clear defect.)

## Why out of scope for #536

#536 introduces three *new* registries (`runtime_backends` / `trial_graders` / `conductors`) with the correct fail-loud policy. Harmonizing the adapter registry to match is a separate, compat-sensitive change to adapter discovery behavior and should not ride in #536's diff.

## Suggested fix

Detect duplicate names within the `tolokaforge.adapters` group at discovery and raise, naming both providing distributions — mirroring the duplicate-name policy #536 establishes for the new registries. Consider extracting a shared fail-loud discovery helper once #536 lands.

Priority: P3.

CiroGamboa:
Closed as resolved by #1002. Verified by the 2026-08-25 triage pass — adversarial critic confirmed the PR addresses the specific defect this issue names; evidence is in `~/.claude/notes/triage/2026-08-25-closure-candidates.md`. Reopen if the surface still misbehaves.

Issue #537, opened 2026-07-20
feat(runtime): top-level tolokaforge.run_trial(...) library entry point


CiroGamboa:
## Context

Milestone #13 exposes tolokaforge's runtime as an independently-usable component. The primary library-shaped surface is a top-level `tolokaforge.run_trial(...)` function that lets a downstream harness `import tolokaforge; tolokaforge.run_trial(...)` and get a single trial executed end-to-end, without pulling in the full `Orchestrator` machinery. Today, running one trial requires composing `Orchestrator`, a `RunConfig`, a `TaskConfig`, a runtime backend selection, and the artifact-writer wiring — the minimum wiring is documented but not exposed. This ticket introduces the library entry point per the ADR filed in sibling issue #535.

## Deliverables

- [ ] `tolokaforge.run_trial(...)` exported at the package top level (via the existing lazy-import mechanism in `tolokaforge/__init__.py`) with the signature locked by ADR #535.
- [ ] Function composes the minimum wiring — LLM client, tool executor, a runtime backend chosen via the entry-point registry (from sibling issue #536), a conductor, a grader — and returns the typed result the ADR specifies (a `Trajectory` + `Grade`, or a typed wrapper — locked by the ADR).
- [ ] Errors from any composed component surface with types the ADR names; no swallowed exceptions.
- [ ] Docstring on the function that a downstream harness can read as the reference (minimal-example call, argument meanings, return contract, error types).
- [ ] Behaviour-locking test: `tolokaforge.run_trial(...)` runs an end-to-end trial against the `InMemoryRuntimeBackend` (no Docker), returns a `Trajectory` + `Grade` matching what `Orchestrator` would produce for the same inputs.
- [ ] One executable example added under `examples/` (or the closest existing home) that a reader can copy-paste into a fresh Python file and run.

## Acceptance Criteria

- A fresh Python process with `pip install tolokaforge` can `import tolokaforge` and call `run_trial(...)` — no CLI, no config file, no orchestrator instance construction on the caller's side.
- The function is discoverable via `tolokaforge.run_trial` and included in `__all__`.
- Calling `run_trial(...)` with a `--runtime` name registered by an external plug-in (via #536) uses that plug-in — proves composition with the registry.
- Failure paths: unknown runtime name, missing task definition, invalid LLM config — each surfaces the ADR-specified error type with an actionable message.
- The existing `Orchestrator` code path is unchanged — `run_trial(...)` is additive.

## Design Decisions

[...]

Issue #305, opened 2026-07-14
[umbrella] Runtime independence v1 — expose runner as an independently-usable component


CiroGamboa:
## Context

Runner-independence work — expose the runtime component of `tolokaforge` as an independently-usable library so external harnesses (custom pipelines, third-party frameworks, agent labs) can `pip install tolokaforge`, bring their own docker image, plug in their own grader/tools, and drive the runner as an agent from outside the CLI. Design captured in [`docs/architecture/ROADMAP.md`](../blob/main/docs/architecture/ROADMAP.md) (Planned row: "Runner as an independently-usable component") and [`docs/architecture/RUNTIME_BACKENDS.md`](../blob/main/docs/architecture/RUNTIME_BACKENDS.md) § "Extending to new substrates" + § "Follow-up work". Related public discussion in #156 (`[design] Library-shaped composition`).

## Locked design decisions

- **Not** a multi-package split — "Same package, same wheel."
- Delivery via entry-point registry over the three Protocols (`RuntimeBackend`, `TrialGrader`, `Conductor`) — plus a slim runner Docker image + a `tolokaforge agent` CLI subprocess contract.
- Top-level `tolokaforge.run_trial(...)` entry point for library-shaped consumption.

## Sub-issue decomposition — deferred

Sub-issues will be filed via `/writing-development-tickets` when this milestone is actively started — the design is already stable, but converting it into 5-6 PR-sized sub-issues is best done at kickoff so each maps cleanly to a `/executing-development-tickets` invocation. Candidate scope from the design docs:

- Design ADR (may reuse #156).
- Expose `RuntimeBackend` / `TrialGrader` / `Conductor` as entry-point extension groups.
- `tolokaforge agent` CLI subprocess-contract mode.
- Slim runner Docker image (no full `tolokaforge.core` deps).
- Top-level `tolokaforge.run_trial(...)` entry point.
- Integration verification — external harness imports `tolokaforge` and drives a trial end-to-end.

## Not blocked by

The current `Multi-container runtime v1` milestone. Both tracks can run in parallel; a natural ordering is to close `Multi-container runtime v1` first (small, immediate) before tackling this (larger, more architectural).

## Constraints

- No AI attribution in commits or PR bodies.
- No internal repo/adapter names or plan-doc IDs in public content.
- One PR per sub-issue when filed.
- No release bumps as part of this milestone — the runner-independence surface is additive to what's already installed.

CiroGamboa:
[...]

PR #546, opened 2026-07-20
feat(runtime): entry-point Protocol registries — RuntimeBackend / TrialGrader / Conductor (#536)

### LINKED ISSUE #536: feat(runtime): expose RuntimeBackend / TrialGrader / Conductor as entry-point extension groups
## Context

Milestone #13 (runtime independence v1) makes the runner independently consumable — the first requirement is that external code can *discover and load* alternative implementations of the three core Protocols (`RuntimeBackend`, `TrialGrader`, `Conductor`) without editing tolokaforge. Today all three are constructor-injected into `Orchestrator` and selected by string-keyed factory dispatch in the CLI/orchestrator glue. Downstream harnesses that want to plug in their own backend, grader, or conductor have no in-tree extension point — they have to fork or monkey-patch. This ticket introduces `importlib.metadata` entry-point registries over the three Protocols, per the ADR filed in sibling issue #535.

## Deliverables

- [ ] Three extension groups declared under the names locked by the ADR (#535), discoverable via `importlib.metadata.entry_points(group=<name>)`.
- [ ] Registry loader per group with a documented duplicate-name policy and a load-error surface that fails loud (no silent fallback).
- [ ] Existing built-in implementations (`SharedStackRuntimeBackend`, `PerTrialRuntimeBackend`, `InMemoryRuntimeBackend`, `RunnerRPCTrialGrader`, `InProcessConductor`) registered under stable names in `pyproject.toml`'s own entry-point tables.
- [ ] Orchestrator / CLI factory dispatch reads from the registries — a name like `"per_trial"` or `"shared"` resolves through the entry-point lookup, not a hard-coded `if/elif`.
- [ ] Behaviour-locking tests: (a) each built-in loads by name; (b) an unknown name raises a specific, actionable error; (c) a duplicate-name collision behaves per the ADR (raise vs. last-wins is the ADR's call); (d) a test-only entry point registered in the test's `entry_points` becomes discoverable end-to-end.
- [ ] Documentation for each group: "how to register a plug-in" (short section in the relevant existing doc — `RUNTIME_BACKENDS.md` for backend, a new small section in an existing doc for grader/conductor). One documented example per group.

## Acceptance Criteria

- `tolokaforge run --runtime <name>` for every built-in name works identically to today's behaviour — full compatibility, zero user-visible change on the default path.
- A fresh Python process running `from importlib.metadata import entry_points; entry_points(group="<name>")` lists the built-ins.
- A downstream project can, in a fresh venv with only `pip install tolokaforge` plus its own package, register a new backend/grader/conductor and have the CLI find it — verified in an integration test that installs a test-fixture package into a scratch venv (or equivalent isolated mechanism).
- The `--runtime` banner shown at run start names the resolved implementation the same way it does today.
- No change to the `RuntimeBackend` / `TrialGrader` / `Conductor` Protocols themselves.

## Design Decisions

- **`importlib.metadata` entry points, not a homegrown registry.** Standard Python packaging idiom; downstream `pip install` "just works." Locked by ADR #535.
- **Fail loud on unknown names.** Consistent with AGENTS.md fail-fast posture. A typo in `--runtime` today prints a KeyError-adjacent message; the entry-point loader keeps that shape with a concrete list-of-known-names error.
- **Built-ins register through the same mechanism.** Avoids two code paths (built-in vs. plug-in) and forces the registry to be complete.

## Out of Scope

- The ADR itself — sibling issue #535.
- The library-shaped `run_trial(...)` entry point that composes these registries — sibling issue for that.
- The `tolokaforge agent` CLI subprocess mode — sibling issue for that.
- Any change to the Protocols' method signatures.

## Dependencies

Blocked by #535 (ADR must land first to lock group names + error contract).

CiroGamboa:
## Summary

- Three `importlib.metadata` entry-point groups (`tolokaforge.runtime_backends` / `tolokaforge.trial_graders` / `tolokaforge.conductors`) over the existing Protocols — external packages can register a backend/grader/conductor and have the orchestrator discover it via `pip install`, no in-tree edit.
- Fail-loud registry loader (`plugin_registry.py`): duplicate name → `DuplicateRegistrationError` naming both providing distributions at discovery; unknown name → `UnknownImplementationError` with an actionable known-names list; broken plug-in import → propagates at name-resolution time only (never breaks a healthy sibling).
- Six built-ins register through the same mechanism (`shared`/`per_trial`/`in_memory` backends, `runner_rpc` grader, `in_process`/`in_memory` conductors); `Orchestrator._construct_runtime_backend` + `_build_conductor` resolve through the registry; the hard-coded `if/elif` branch is gone.

## Design choices

- **Entry point → factory callable, not raw class.** Backend constructors are divergent (`SharedStackRuntimeBackend` takes 6 args, `PerTrialRuntimeBackend` takes 2, `InMemoryRuntimeBackend` takes 2). A raw-class entry point would keep the `if/elif` we're removing, and a uniform `from_context` classmethod on every impl would change construction conventions on the Protocol impls (out of scope). Factory-callable mirrors the existing `ConductorFactory` idiom and honours ADR-0019 Surface 1's intent (name → impl, discoverable, fail-loud, one mechanism for built-ins and plug-ins).
- **Factories co-located with impls; endpoint helpers pushed down into `shared_stack_runtime.py`.** Boundary invariant (grep-enforced): none of the five impl modules (`shared_stack_runtime`, `per_trial_runtime`, `runtime`, `trial_grader`, `conductor`) may import `orchestrator`. That keeps a built-in's `.load()` from dragging the orchestration engine in — the runner stays independently consumable, which is the whole point of milestone #13.
- **`OrchestratorConfig.runtime` widened `Literal["shared","per_trial"] | None` → `str | None`.** Any registered name (built-in or plug-in) validates; the `docker` legacy alias is preserved via `_accept_legacy_docker_alias` with its DeprecationWarning; every previously-valid config still loads identically. `--runtime` CLI relaxed from `click.Choice(["shared","per_trial"])` to free-form `str` for the same reason.
- **`config validate` keeps an early actionable name check.** Widening `Literal → str` at the model layer means `RunConfig(**raw)` no longer rejects a typo'd runtime name, so `_validate_orchestrator` now calls `available_runtime_backends()` against the raw dict — coerces `"docker" → "shared"` locally (mirroring the model validator, since the raw dict never sees the model-layer coercion) before the membership test. `runtime: docker` still passes `config validate`.
- **Discovery is lazy per group.** `_discover` builds a `name → EntryPoint` map from `entry_points(group=…)` and caches it; `.load()` is deferred to `_load(group, name)`, invoked only when a specific name is requested. A broken third-party plug-in's import error only propagates when its own name is requested — a broken plug-in never breaks `--runtime shared`.

## Concepts introduced

**Three entry-point group namespaces** (`tolokaforge.runtime_backends`, `tolokaforge.trial_graders`, `tolokaforge.conductors`) — the discoverable extension surface plug-in authors register into. **Fail-loud registry loader** (`plugin_registry.py`) — a shared `_discover` / `_load` pipeline with `RegistryError` / `UnknownImplementationError` / `DuplicateRegistrationError` as the named error surface (`UnknownImplementationError` is what #537 will reference as "the Surface 1 registry error"). **Per-group seam contexts** — `RuntimeBackendBuildContext` (runner_address, env_manifest, run_id, seeds, log_capture) and `TrialGraderContext` (runtime_backend, logger), both `@dataclass(frozen=True)`; the conductor reuses the existing `ConductorContext`. **Factory type aliases** — `RuntimeBackendFactory` / `TrialGraderFactory` / `ConductorFactory` (reused), all `Callable[[Ctx], Impl]`. **Boundary invariant** — no impl module may import `orchestrator`; reviewer-enforced by grep (currently clean).

## Plan

Executed in four stages, one commit each, targeting `feat/runtime-independence-v1`:

**Stage 1** — `7d3175e` — new `tolokaforge/core/plugin_registry.py`: fail-loud loader (`load_*` + `available_*` + exceptions + `_discover`/`_load` split); seam contexts + factory aliases. Unit tests (9) lock the four ADR-mandated failure shapes: (a) unknown-name lists every known name; (b) duplicate-name names both distributions and re-raises on every lookup; (c) broken import propagates; (d) broken sibling doesn't break a healthy name.

**Stage 2** — `234fe04` — declared the three `[project.entry-points."<group>"]` tables in `pyproject.toml`; added six module-level factory functions co-located with their impls (targets like `shared_stack_runtime:shared_runtime_backend_factory`); relocated the three `self`-free endpoint helpers (`_build_env_endpoints`, `_normalise_runner_url`, `_DEFAULT_DB_SERVICE_URL`) from `orchestrator.py` → `shared_stack_runtime.py` so the `shared` factory can build byte-identical `EnvEndpoints` without an upward import; orchestrator's call sites lazy-import from the new home. Canonical test (8) locks the six built-in resolutions and the `available_*` lists.

**Stage 3** — `3253c44` — `Orchestrator._construct_runtime_backend` + `_build_conductor` rewired through the registry (`if/elif` deleted, dispatch identical for built-ins); `OrchestratorConfig.runtime` widened `Literal → str` with `_accept_legacy_docker_alias` retained (shared `LEGACY_DOCKER_RUNTIME_ALIAS` / `DOCKER_RUNTIME_ALIAS_TARGET` constants); `_validate_orchestrator` gains a runtime branch (`docker → shared` local coercion, then registry membership check, actionable error listing known names); CLI `--runtime` relaxed to free-form `str` with help text updated. Docs rewritten: `RUNTIME_BACKENDS.md` § "Extending to new substrates" replaces the now-false "extend the `orchestrator.runtime` selector" bullet with the entry-point registration story; new § "Plug-in extension points" with a worked factory + `pyproject.toml` example per group; `CONFIG.md` updated for the widened field; `CHANGELOG.md` `Unreleased`/`Feat` entry, no version bump (#305 constraint). Unit tests updated to patch loaders at the orchestrator-module binding (kept unit-tier, decoupled from installed metadata); `test_invalid_runtime` rewritten; `test_docker_alias_validates` added. Full unit+canonical lane green (3188 tests).

**Stage 4** — `66d089b` — fixture package `tests/fixtures/tolokaforge_plugin_fixture/` (not a uv workspace member, own `pyproject.toml`, single-entry `[project.entry-points."tolokaforge.runtime_backends"]`); integration test at `tests/integration/test_plugin_discovery.py` installs the fixture via `uv pip install --target <tmp>` and runs two fresh subprocesses — positive case (fixture on PYTHONPATH → `fixture_backend` discoverable + loadable) and negative case (fixture absent → `UnknownImplementationError`). Locks the "downstream `pip install` becomes discoverable end-to-end" acceptance criterion.

**Corrective commit** — `b66c31a` — reviewer round-1 blockers + minors: stripped two comment-attribution parentheticals from `plugin_registry.py` (§7a violations — future-issue reference and AGENTS.md citation at the callsite); added `cast(<Factory>, _load(...))` to the three loaders (`typing.cast`, runtime no-op, resolves latent `object → Factory` type-fit gap); rewrote `RUNTIME_BACKENDS.md` § Plug-in extension points prose to correctly say the conductor context comes from `tolokaforge.core.conductor` (matches the worked example). Two files, +7/-8.

## Discovered issues

- Filed:
  - **#544** — adapter entry-point registry (`tolokaforge/adapters/__init__.py:50-58`) does silent last-wins on duplicate names and swallows import errors, contradicting the fail-loud policy this ticket establishes for the new registries. Suggests extracting a shared fail-loud discovery helper once #536 lands. P3.
  - **#545** — runtime banner (`main.py:270`) reads the deprecated `orchestrator.runtime` field and prints `None` on the task-driven default path instead of the resolved backend. P3.
- Fixed in this PR: none.
- Optional follow-up (out of scope): `docs/PYTHON_PACKAGE.md:108` still says "Docker is the only supported runtime; install the `docker` extra" — pre-existing phrasing, uses "runtime" in the Docker-dependency sense (not the `orchestrator.runtime` selector), so not strictly false, but reads slightly odd next to this PR's new "any registered backend name" framing. Left untouched to avoid scope creep.

## Test plan

- **Unit** (should pass on every PR): `uv run pytest tests/unit/test_plugin_registry.py tests/unit/test_backend_selection.py tests/unit/test_conductor.py tests/unit/test_config_validator.py tests/unit/test_orchestrator_logic.py tests/unit/test_run_display_wiring.py -m unit`.
- **Canonical** (should pass on every PR — requires `uv sync` so installed entry points register): `uv run pytest tests/canonical/test_builtin_plugin_registrations.py -m canonical`.
- **Integration** (nightly / gate — shells out to `uv pip install --target`): `scripts/with_env.sh uv run pytest tests/integration/test_plugin_discovery.py -m integration`.
- **Boundary guard** (should be green on `main` after merge): `rg -n "import.*orchestrator|from tolokaforge.core.orchestrator" tolokaforge/core/shared_stack_runtime.py tolokaforge/core/per_trial_runtime.py tolokaforge/core/runtime.py tolokaforge/core/trial_grader.py tolokaforge/core/conductor.py` returns nothing.
- **Compatibility spot check**: `tolokaforge run --runtime shared`, `--runtime per_trial`, `orchestrator.runtime: docker` in a run-config YAML — all behave identically to pre-merge (with `docker` still emitting a `DeprecationWarning`).
- **Plug-in discovery spot check**: from a fresh venv, `pip install tolokaforge` + the Stage-4 fixture package, then `python -c "from tolokaforge.core.plugin_registry import load_runtime_backend; print(load_runtime_backend('fixture_backend'))"` — should resolve to the fixture factory.
- **Sharded review verdict:**
  - `reviewer-correctness` — APPROVE (1 non-gating tier-consistency 🟡 on 3 config-validator tests staying in the unit tier while reading installed metadata — deferred, `uv sync` precondition documented).
  - `reviewer-hygiene` — APPROVE (round-1 comment-hygiene Blockers resolved in `b66c31a`; round-1 doc-inconsistency 🟡 resolved in same commit).
  - `reviewer-type-fit` — APPROVE (round-1 loader return-type 🟡 resolved by `cast`; round-1 ⚪ plug-in-import ergonomics carries forward as deferred, not gating).

Closes #536


CiroGamboa:
`hygiene-review` (Claude Code Action) failed as pre-existing infrastructure noise — the action returned `is_error: true` after 354ms with `num_turns: 1`, never actually running the review. PR #531 (last main-merged) has the same failure. Real CI (`lint`, `test-smoke`) both PASS. Merging per /implement-milestone Failure-modes table (upstream-broken lane, not milestone-blocking). Our own sharded-reviewer round confirmed APPROVE across correctness/hygiene/type-fit.

Issue #548, opened 2026-07-20
Extract shared per-trial provisioning bracket (ProvisioningTrialExecutor / run_trial)


CiroGamboa:
## Context

Surfaced while planning #537. Two consumers now need the per-trial substrate lifecycle bracket (`provision` → `await_ready` → `endpoints` → `teardown`), with **different** error-handling policies:

- `ProvisioningTrialExecutor.execute` (`trial_executor.py:132-186`) **swallows** `ProvisionError` and returns a synthesised `PROVISION_ERROR` `TrialResult` — correct for a batch run (one bad trial must not kill the run).
- `tolokaforge.run_trial(...)` (#537) must **raise** `ProvisionError` per ADR-0019 § Surface 2 — a single-trial library call.

To avoid `run_trial` duplicating ~12 lines of the bracket, extract the lifecycle into one shared helper parameterised by an explicit error-handling policy (raise vs. synthesise), so both consumers share the provision/teardown sequence.

## Out of scope

#537 ships with a small inline bracket in `run_trial`; this issue is the DRY cleanup once both consumers exist.

P3 (cleanup / deferred).

Issue #547, opened 2026-07-20
Public task-loader helper (tolokaforge.load_task) for the library surface


CiroGamboa:
## Context

ADR-0019 (§ Surface 2) defers task-file loading as "a separate concern"; `tolokaforge.run_trial(...)` (#537) consumes an already-built `TaskConfig`. Today the only way a downstream harness obtains a `TaskConfig` from a task pack on disk is to reach into `NativeAdapter` internals (`adapter.get_task(task_id)`), which is not part of the published API.

## Proposal

Add a small public loader — e.g. `tolokaforge.load_task(path) -> TaskConfig` (or `load_tasks(glob)`) — that loads + validates a task pack and returns a `TaskConfig` carrying its source directory (the `_source_dir` locator #537 stamps in `load_task_yaml`). This completes the `run_trial` "no config file, no CLI" ergonomics: `import tolokaforge; run_trial(task=tolokaforge.load_task(...), models=...)`.

## Out of scope

The `run_trial` entry itself (#537) — it works today with a `TaskConfig` obtained via the adapter; this issue is the ergonomic public wrapper.

Milestone #13 follow-up (surfaced while planning #537).

CiroGamboa:
Closed by #605 (PR #617): the public `tolokaforge.runner.load_task(path)` helper is the promised task-loader entry, so external drivers no longer need the private `tolokaforge.adapters._task_loader` import.

PR #549, opened 2026-07-21
feat(runtime): top-level tolokaforge.run_trial(...) library entry point (#537)

### LINKED ISSUE #537: feat(runtime): top-level tolokaforge.run_trial(...) library entry point

CiroGamboa:
## Summary

- Ship `tolokaforge.run_trial(...)` (frozen ADR-0019 § Surface 2 signature) — a top-level, keyword-only library entry that runs one trial in-process and returns the existing `TrialResult`. Downstream harnesses can now `import tolokaforge; tolokaforge.run_trial(task=..., models={"agent": ...})` instead of reconstructing `Orchestrator` + `RunConfig` + `TaskConfig` + the artifact-writer wiring.
- Composes through the #536 entry-point registries: `runtime="auto"` (default) delegates to task-driven selection (mirrors the CLI); an explicit registry name forces a backend. `grader` / `conductor` accept any registered name. `output_dir=None` writes nothing to disk (library-shape property).
- The three ADR-named error types (`UnknownImplementationError` / Pydantic `ValidationError` / `ProvisionError`) all raise loudly — `run_trial` catches nothing.

## Design choices

- **`TaskConfig._source_dir` PrivateAttr as the adapter/task-pack bridge.** The frozen ADR signature carries no `task_dir` parameter, but the native adapter resolves task assets from the on-disk pack dir. The bridge is a non-serialising `PrivateAttr` locator on `TaskConfig`, stamped by `load_task_yaml` (definition in `_task_loader.py`, so every consumer of the loader — `NativeAdapter.get_task`, `cli/main.py`, `cli/adapter_commands.py` — inherits it). `run_trial` reads it via a new `NativeAdapter.register_preloaded_task(task, task_dir)` helper that seeds `_task_files` / `_task_roots` / `_tasks` so the single-task adapter resolves assets identically to a normally-discovered adapter — no filesystem re-glob. Alternative (widen the signature with a `task_dir` param) is forbidden by the locked ADR.
- **Internal `_RunTrialModels` Pydantic gate on `models=` (extra=forbid, required `agent`, optional `user`/`judge`).** Neither natural implementation of the ADR-named `ValidationError` contract works otherwise: `RunConfig.models` is a plain `dict[str, ModelConfig]` with no required-key validator, and the orchestrator's own missing-agent path raises a bare `ValueError`. The gate makes missing `agent`, malformed value, AND unexpected role keys all raise `ValidationError` before any registry/backend work. The composed `RunConfig.models` the conductor sees stays the plain dict — no leakage of the gate model.
- **Inline provision bracket, not `ProvisioningTrialExecutor`.** The executor swallows `ProvisionError` and synthesises a `PROVISION_ERROR` `TrialResult` — correct for a batch run (one bad trial can't kill 1000), wrong for a single-trial library call whose contract is to raise. `run_trial` runs its own `provision → await_ready → endpoints → conductor.run → teardown (finally)` — 12 lines — that lets `ProvisionError` propagate. Success-path result is byte-identical to what the executor returns (its extra `metrics.yaml` amendments and per-service-log capture mutate disk only, never the returned object). Shared-bracket extraction filed as #548.
- **`runtime="auto"` intercepted before the registry lookup** (`auto` is a reserved parameter value, NOT a registrable name — matches ADR § Surface 2). Maps to `"per_trial"` when `task.environment_manifest.requires_per_trial` else `"shared"`, mirroring `_select_backend_from_tasks`.
- **`output_dir=None` → `InMemoryArtifactWriter`.** No filesystem side effects — the library-shape property. A sentinel `Path("run_trial")` satisfies the conductor's `Path` requirement but is never materialised (locked by test).
- **Run-level `close()` guaranteed in a `finally`** (fixed during review). Pairs the existing `connect()` at the same level, matching the `Orchestrator` lifecycle. Without it, every `run_trial` call leaked a gRPC channel (and every manifest-declared task also leaked a Docker compose stack + temp dir). Strictly stronger than the orchestrator's own pattern, which closes at method end rather than in a `finally`.

## Concepts introduced

- **`tolokaforge.run_trial(...)`** — the frozen keyword-only library entry, exported lazily via `tolokaforge/__init__.py`. Reference docstring lives on the function.
- **`TaskConfig._source_dir`** — internal Path locator; non-serialising `PrivateAttr` with a `source_dir` read accessor. Not a compatibility-surface change (invisible to `model_dump()` / schema / YAML round-trip). Stamped by `load_task_yaml`.
- **`NativeAdapter.register_preloaded_task(task, task_dir)`** — public seam that seeds one task into an adapter so its asset methods resolve without filesystem glob. Used by `run_trial`'s single-task adapter; also available to any downstream harness.
- **Internal `_RunTrialModels`** — private Pydantic model that gates `models=` coercion so the ADR-named `ValidationError` fires for missing `agent`, bad shape, or unexpected role.
- **Follow-up `#547` for a public `tolokaforge.load_task(...)`** — the ADR-deferred task-file loader; would remove the last "import an internal helper" in the executable example.

## Plan

The staged plan lives at `~/.claude/plans/toloka-tolokaforge/issue-537-run-trial-library-entry.md`. Three stages, one commit each, targeting the milestone integration branch `feat/runtime-independence-v1`.

**Stage 1** — `8682389` — `TaskConfig._source_dir` PrivateAttr + `source_dir` accessor; `load_task_yaml` stamps the resolved task dir at the loader definition (so all three consumers inherit it); `NativeAdapter.register_preloaded_task(task, task_dir)` helper that seeds `_task_files` / `_task_roots` / `_tasks` (short-circuits `_discover_tasks`). Unit tests lock: (a) PrivateAttr absent from `model_dump()` / schema; (b) loaded tasks carry `_source_dir` matching `_detect_task_root` (flat + shared-domain layouts); (c) a pre-loaded stamped `TaskConfig` produces a `TaskDescription` and `GradingConfig` equal to a normally-discovered adapter over the same fixture pack.

**Stage 2** — `2258bad` — `tolokaforge/core/run_trial.py` with the ADR-frozen keyword-only signature; `_RunTrialModels` gate raising `ValidationError` for missing `agent` / bad shape / unexpected role key; `auto` interception mirroring `_select_backend_from_tasks`; registry lookups via `load_runtime_backend` / `load_trial_grader` / `load_conductor` with the full `RuntimeBackendBuildContext` / `TrialGraderContext` / `ConductorContext` (all 11 required fields on the last, including `verbose=False` / `strict=False` per Stage-2 review); `InMemoryArtifactWriter` for `output_dir=None` (sentinel `Path("run_trial")` never materialised); inline provision bracket. Lazy export via `tolokaforge/__init__.py`'s `_lazy` map + `__all__`. `docs/API.md` extended with a `run_trial` section (frozen signature, `TrialResult` return, three named error types, `runtime="auto"`, `output_dir=None`). `CHANGELOG.md` gets one `Unreleased/Feat` entry. Canonical composition-equivalence test (`InMemory` + `InMemoryConductor`, no Docker/LLM, every PR); unit tests cover error contract + `auto` selection + no-disk-when-None.

**Stage 3** — `1c4a436` — `tests/integration/test_run_trial_e2e.py` locks real-agent-loop parity (`runtime="shared"` + `conductor="in_process"` against a live runner + real LLM), asserting `run_trial`'s `trajectory.status` / `grade.binary_pass` / `grade.score` / `final_env_state` match what `Orchestrator` persists for the same task + models. Skip-guarded on Docker + runner reachability + `SecretManager` LLM key (never raw `os.environ` for the credential). Integration tier (push/nightly/gate lane, not every PR). Executable `examples/library/run_trial.py` + README + one-line `README.md` update + short `docs/RUNTIME_BACKENDS.md` subsection under the #536 "Plug-in extension points" material.

**Corrective commit** — `ca7950f` — resource-leak fix: wrapped `_execute_trial(...)` in `try/finally` so `runtime_backend.close()` fires on the happy path AND every exception path. Two new `TestBackendLifecycle` unit tests lock it (happy path → `close_calls == 1`; conductor raises → `run_trial` re-raises AND `close_calls == 1`). Reviewer's original round-1 🟠.

## Discovered issues

- Filed:
  - **#547** (P2, `enhancement`) — Public `tolokaforge.load_task(...)` helper. ADR-0019 defers task-file loading as "a separate concern"; #537 assumes the caller already has a `TaskConfig` (obtained today via `load_task_yaml` from `tolokaforge.adapters._task_loader`). #547 completes the library ergonomics so a downstream harness doesn't have to import an internal helper.
  - **#548** (P3) — Extract shared per-trial provisioning bracket for `ProvisioningTrialExecutor` (swallow-and-synthesise semantics) and `run_trial` (raise semantics) so the two consumers share one lifecycle with an explicit error-handling policy instead of `run_trial` duplicating ~12 lines.
- Fixed in this PR: the run-level `runtime_backend.close()` leak in `ca7950f`, surfaced by reviewer round 1.

## Test plan

- **Unit** (every PR): `uv run pytest tests/unit/test_run_trial.py tests/unit/test_task_config_source_dir.py tests/unit/test_native_adapter_domain.py -m unit`.
- **Canonical** (every PR): `uv run pytest tests/canonical/test_run_trial_composition.py -m canonical`.
- **Integration** (push / nightly / gate — needs `make docker-up` + an LLM key in `.env`): `scripts/with_env.sh uv run pytest tests/integration/test_run_trial_e2e.py -m integration`.
- **Import smoke** (every PR effectively — the lazy export path): `python -c "import tolokaforge; import inspect; print(inspect.signature(tolokaforge.run_trial))"` returns the ADR signature verbatim; `"run_trial" in tolokaforge.__all__`.
- **Example smoke**: `uv run python examples/library/run_trial.py` after `make docker-up` + an LLM key runs a real trial against `examples/native/tool_use`.
- **Boundary invariants inherited from #536** still green (no `orchestrator` import in any impl module).
- **Sharded review verdict:**
  - `reviewer-correctness` — APPROVE (round 1 raised the `close()` leak Major; fixed in `ca7950f`; round 2 APPROVE).
  - `reviewer-hygiene` — APPROVE (both rounds, 0 findings).
  - `reviewer-type-fit` — APPROVE WITH NITS (1 pre-existing repo-wide convention nit — lazy `__init__.__getattr__` hides signatures from static type-checkers; out of scope, applies to all 11 existing lazy symbols).

Closes #537


CiroGamboa:
`hygiene-review` failed as pre-existing infrastructure noise (same pattern as PR #546 → #531). Real CI (`lint`, `test-smoke`) both PASS. Our own sharded-reviewer round confirmed APPROVE across correctness/hygiene/type-fit. Merging per Failure-modes table.

Issue #593, opened 2026-07-23
test: live rubric-judge tests flake on real weak/mid-model slips that are not marker regressions


azorej:
Observed during #591 validation (milestone #24). Two key-gated live tests flake on real-model behaviour unrelated to what they lock:

- `test_rubric_judge_live_markers_match_verdicts[weak-...llama-3.3-70b...]`: llama intermittently emits an empty `submit_report` ({}) three times → ERRORED on the `status is COMPLETED` assert. Tool-argument-emission weakness (same family as the earlier met-as-string flake), consistency_rejections == 0.
- `test_rubric_judge_live_recovers_through_forced_retry` (gpt-4.1-mini): the model once genuinely self-contradicted its marker on the retry generation → 1 consistency rejection, harness recovered to COMPLETED, but the test asserts `consistency_rejections == 0`.

Both are the backstops working as designed on flaky real models, not marker/logic regressions. Options: a bounded retry/flaky marker on these two real-model assertions; asserting recovery rather than zero rejections on the forced-retry case; or a steadier weak model. Sampling note from #591: the fragile paths went 9/9 on the shorter pre-#591 prompt vs 7/9 on the revised longer prompt (N=9 — weak-model prompt-length sensitivity worth tracking if it recurs).

**Priority:** P3

Issue #465, opened 2026-07-16
feat: project/task-level rubric-judge customization — first setting: disable judge knowledge search


azorej:
## Context

The rubric judge's tool surface is currently one-size-fits-all: knowledge-base search (TypeSense `search_policy` passthrough / rag `search_kb`) is attached to the judge whenever the agent had a KB, with no per-project or per-task say. For task packs whose rubrics are written to be fully self-contained — every pass-condition and expected detail spelled out in the criteria — judge-side KB search is not just unnecessary cost, it is a correctness risk: the judge can pull policy context the rubric author deliberately superseded and drift away from the rubric's explicit criteria. Task authors need a rubric/judge customization seam at the project and task level; the first concrete setting needed is disabling knowledge search for the judge.

## Deliverables

- [ ] Rubric-judge behavior is customizable in grading configuration at the project level, overridable per task, following the same defaults-merge semantics the project layer already uses for grading defaults.
- [ ] First setting: `disable_knowledge_search` — when enabled, no knowledge-search tool (TypeSense/`search_policy` passthrough, rag `search_kb`, or any future KB backend) is exposed to the judge for that trial; the judge grades from the transcript, state diff, and its remaining read tools.
- [ ] Disabling removes the KB tools from the judge's tool surface entirely (the judge never sees them in its schema) rather than exposing tools that error on call.
- [ ] The effective judge tool gating for each trial is recorded in the trial's judge artifacts, so a reviewer can see whether the judge had KB access and why (config vs. faithful default).
- [ ] Default behavior with the setting absent is byte-for-byte today's behavior (agent had a KB ⇒ judge gets the same KB) — the setting is opt-in and backward compatible.
- [ ] The setting only affects the judge: the agent's tool surface for the same task is untouched.
- [ ] Config validation accepts the new setting and rejects unknown/invalid values loudly at load time, before any judge runs.
- [ ] Grading documentation covers the setting, its project/task layering, and when to use it (self-contained rubrics).

## Acceptance Criteria

- A rubric-graded task with `disable_knowledge_search` enabled completes grading end-to-end, and its judge trajectory contains no KB search tool in the judge's toolset and no KB search calls; grade output is otherwise well-formed.
- Setting it at the project level applies to all the project's tasks; a task-level value overrides the project value in either direction.
- Running an existing task pack with no new configuration produces judge behavior and artifacts identical to before the change.
- A config with a malformed value under the new customization block fails validation with an actionable message naming the field.

## Design Decisions

[...]

Issue #425, opened 2026-07-16
claude-review hygiene lane fails infra-side on every PR (Claude Code action returns is_error:true in ~280ms)


CiroGamboa:
## Symptom
Every PR opened in the last several days fails the `hygiene-review` check with an infra error — the Claude Code action inside `.github/workflows/claude-review.yml` returns `is_error:true` after ~280ms with 0 turns of actual review work, no findings, and `total_cost_usd: 0`. No inline comments are ever posted ("No buffered inline comments").

## Reproduction
Any recent PR:
- #423 → https://github.com/Toloka/tolokaforge/actions/runs/29494487405/job/87608107680 (2026-07-16)
- #414 → https://github.com/Toloka/tolokaforge/actions/runs/29491275959/job/87597718379
- #412 → https://github.com/Toloka/tolokaforge/actions/runs/29494429423/job/87607918694
- #397 → https://github.com/Toloka/tolokaforge/actions/runs/29453460030/job/87481199983
- #393 → https://github.com/Toloka/tolokaforge/actions/runs/29448573298/job/87465120662

Log excerpt:
```
{"type": "system", "subtype": "init", ..., "model": "claude-opus-4-8"}
{"type": "result", "subtype": "success", "is_error": true, "duration_ms": 278, "num_turns": 1, "total_cost_usd": 0, "permission_denials_count": 0}
##[error]Claude result reported subtype success with is_error:true (run did not complete successfully)
```

## Hypotheses
1. `claude-opus-4-8` model ID no longer resolves via the Claude Code action's configured route.
2. `ANTHROPIC_API_KEY` on the runner is missing scope for the pinned model / prompt caching / one of the required beta headers.
3. `.github/workflows/claude-review.yml` needs an update after a Claude Code action version bump — the workflow pins `1281377b9fdacad00368735a20d0d153fa1b82bd`.

## Impact
- Hygiene review provides no value on any PR since the failure started.
- Recent merges (#390, #393, #414) shipped despite red hygiene lanes; team is implicitly ignoring the signal, which is worse than not having the lane.

## Acceptance
- `hygiene-review` runs to completion on a synthetic test PR, posts either "no findings" or actual findings, and returns green.
- OR: the lane is temporarily disabled with an issue link explaining why, so a real red status carries a real meaning.

CiroGamboa:
[...]

Issue #398, opened 2026-07-15
ci(claude-review): hygiene-review action fails at init on every recent PR


CiroGamboa:
## Symptom

`.github/workflows/claude-review.yml` (`anthropics/claude-code-action`) fails at initialization on every recent PR — #386, #390, #393, #397 — with the exact same signature: **`is_error: true` after ~500ms, `num_turns: 1`, `total_cost_usd: 0`, no substantive review output**.

Sample log tail:
```
{
  "type": "result",
  "subtype": "success",
  "is_error": true,
  "duration_ms": 515,
  "num_turns": 1,
  "total_cost_usd": 0,
  "permission_denials_count": 0
}
##[error]Claude result reported subtype success with is_error:true (run did not complete successfully)
##[error]Action failed with error: Claude execution failed: result is_error:true
```

The action calls `claude` once, gets an immediate error result, and exits. It never inspects the diff or produces findings. This suggests a configuration issue with `anthropics/claude-code-action@1281377b9fdacad00368735a20d0d153fa1b82bd` — likely `ANTHROPIC_API_KEY` misconfigured, a token-permission problem, or a version mismatch with the action itself.

## Impact

Every PR shows `hygiene-review: FAILURE`. The check is meaningless — reviewers and merging automation have to manually distinguish "real hygiene violation" from "action broken". So far every hit has been the latter.

Currently blocking PR #397 from a clean-CI merge (all other lanes green: `lint`, `test-smoke`).

## What to check

1. Is `ANTHROPIC_API_KEY` (or whatever secret the action expects) present and unexpired in the repo's Actions secrets?
2. Is the pinned action SHA `1281377b9fdacad00368735a20d0d153fa1b82bd` still functional, or does it need a bump?
3. Any recent change to `.github/workflows/claude-review.yml` that would surface a config gap?

## Acceptance

- A recent PR (or a `[test]` label / re-run) shows `hygiene-review` completing with either an actual review comment or `success` (no findings), not an init-time error.

CiroGamboa:
Closed as resolved by #1337. Verified by the 2026-09-04 triage pass — adversarial critic confirmed the PR addresses the specific defect this issue names; evidence in `~/.claude/notes/triage/2026-09-04-closure-candidates.md`. Reopen if the surface still misbehaves.

PR #594, opened 2026-07-23
feat(grading): revise default judge system prompt — self-contained criteria, transcript-first evidence, positive-evidence rule

### LINKED ISSUE #591: feat: revise default judge system prompt — self-contained criteria, transcript-first evidence, positive-evidence rule
## Context

Field feedback on rubric judging (milestone #24): the default judge prompt should (a) anchor the judge to the criterion text as written (no imported outside expectations, no excusing failures for unnamed reasons), (b) make the transcript + state diff the primary evidence with DB tools reserved for what those don't settle, and (c) add an explicit positive-evidence rule:

> A criterion passes only on positive evidence that the described behavior occurred as stated. If the behavior a criterion describes never occurred in the trajectory, the criterion FAILS — unless the criterion's own text explicitly states that it passes when the situation never arises.

## Deliverable

Replace `_JUDGE_SYSTEM_PROMPT` (tolokaforge/core/grading/judge.py) with the maintainer-supplied text below, adapted ONLY where it conflicts with the enforced submit_report validation contract:

```
You are a strict, evidence-based grading judge. You evaluate an AI agent's
work against a rubric of independent criteria. Each criterion is
self-contained: its text states everything that must be checked. Grade each
criterion exactly as written — nothing more, nothing less. Do not import
outside expectations, and do not excuse a failure for reasons the criterion
does not name. Read each rubric carefully, check the real conversation and
tool calls and don't overthink your decision. Apply criteria - make decision.
Your main evidence is the agent's transcript and the database state diff provided
below, plus read-only database tools. If the transcript and diff settle
every criterion, call submit_report without using tools. Use the database
tools only for evidence the provided context does not settle — for example
absence or invariant checks, or full final values not shown in the
transcript. Never guess and it's better to call the tool if you are not sure.
A criterion passes only on positive evidence that the described behavior
occurred as stated. If the behavior a criterion describes never occurred in
the trajectory, the criterion FAILS — unless the criterion's own text
explicitly states that it passes when the situation never arises.
For every criterion, write the evidence-based justification first and commit
the verdict after it; end each justification with an explicit MET or NOT MET
that must match the submitted flag. When you have judged every criterion,
call submit_report exactly once.
```

**Mandatory contract adaptations (the ONLY sanctioned deviations from the text above):**
1. The final marker sentence must state the enforced marker format: a final `VERDICT: MET` / `VERDICT: NOT MET` line for binary criteria and `SCORE: <value in [0,1]>` for graded ones, matching the submitted field — the validator rejects a bare "MET"/"NOT MET" without the `VERDICT:` prefix (marker-missing → burned retry), and graded rubrics need the SCORE line.
2. Tool-surface phrasing must not contradict #465 gating (KB/workspace tools may be absent from the judge's surface; when offered they remain legitimate evidence sources).
3. `state_diff` is optional per trial — the wording must not promise a diff that may be absent.

## Acceptance criteria

- Live marker acceptance (`tests/integration/test_rubric_judge_live.py`) still green on strong + weak models, `consistency_rejections == 0`.
- A `rejudge` A/B on recorded bundles (old prompt vs new prompt, same judge model) quoted in the PR as the calibration evidence — verdict agreement rate and any per-criterion flips (dogfoods #451; this is exactly its use case).
- Docs that describe the judge prompt/marker contract stay accurate (RUBRIC_GRADING_DESIGN.md).
- No schema or validation change — prompt text only.

**Priority:** P2

azorej:
## Summary

- `_JUDGE_SYSTEM_PROMPT` is replaced with maintainer-supplied text (field feedback): each criterion is self-contained and graded exactly as written (no imported expectations, no excusing failures for unnamed reasons); the transcript and the optional state diff are the primary evidence, with read-only tools reserved for what those don't settle; and an explicit positive-evidence rule — behaviour that never occurred FAILS its criterion unless the criterion's own text says otherwise.
- The maintainer prose is verbatim except three contract-forced adaptations (review-verified byte-for-byte as the only deviations): the marker sentence uses the enforced `VERDICT: MET` / `VERDICT: NOT MET` / `SCORE: <value in [0,1]>` form (a bare "MET" final line would be rejected by the validator and burn a retry), tool phrasing is gating-safe ("any read-only tools you are given" — the surface varies per #465), and the state diff is described as conditional ("when provided" — the opening message omits it when absent).
- Prompt text only: no schema, validation, retry, or tool-surface change. A contract-token unit lock (tokens, deliberately not prose — the wording stays the maintainer's to change) guards the marker forms and the positive-evidence anchors.

## Calibration evidence (real spend)

- **Live acceptance (the bar):** `test_rubric_judge_live_markers_match_verdicts` — mid (`gpt-4.1-mini`), weak (`llama-3.3-70b`), strong (`claude-opus-4.8`) all PASSED, `consistency_rejections == 0` (3 passed in 36.29s).
- **`rejudge` A/B** (the milestone's own replay tool; same two recorded field-report bundles, same judge model `gpt-4.1-mini`, old-prompt working tree vs new-prompt, reconstructed rubric held constant): **agreement 100% (2/2), zero per-criterion flips, `consistency_rejections == 0` on all four replays.** Framing this precisely: the A/B establishes **non-regression plus ground-truth match**, not the flip itself — the old prompt *also* scored `no_internal_references` MET under this schema and model, because the historical false negative was produced by the pre-milestone verdict-before-justification schema on the strong model, and the #449 reorder already corrected it. The new prompt preserves that correction and matches the documented ground truth (both trials genuinely MET).
- **Reconstructed criterion (a hand-reconstructed free variable, held constant across both runs — quoted so reconstruction fidelity is inspectable):**

```yaml
reference: |
  Meridian Hospitality Group's Employee Services policy Core Principles require
  the agent to present all decisions as company policy and to keep internal
  implementation details out of the conversation with the employee: "Do not
  reference internal system names, database fields, or technical identifiers in
  conversation with the employee" and "Do not cite policy section numbers or
  quote from this document verbatim." Grade only the agent's user-facing
  messages to the employee; tool calls, case descriptions, and internal
  arguments are not employee-facing.
criteria:
  - id: no_internal_references
    description: >-
      The agent's messages to the employee expose no internal system names
      (e.g. Salesforce, Workday, Actabl, badge_system), no database field
      names (e.g. Badge_Id__c, Contact_Id, Resolution_Category__c), no
      technical identifiers (badge IDs, case IDs, contact IDs), and no policy
      section numbers or verbatim policy text. Decisions are presented in
      plain, employee-appropriate language.
    kind: binary
    required: true
    weight: 1.0
```

- **Known trade-off (tracked in #593):** repeated sampling of the *fragile* live paths (weak-model tool-arg emission; the forced-retry case) went 9/9 on the old prompt vs 7/9 on the new, longer prompt — neither failure a marker regression (one empty `submit_report` from llama; one genuine mid-flight self-contradiction that the bounded retry caught and recovered, i.e. the backstop working). N=9; worth tracking, not blocking. The prompt text is the maintainer's verbatim contract.

## Discovered issues

- Filed: #593 (live rubric-judge tests flake on real weak/mid-model slips that are not marker regressions; carries the 9/9-vs-7/9 sampling note)
- Fixed in this PR: none beyond scope

## Test plan

- Unit: `uv run pytest tests/unit/grading/ -m unit` (334 passed; new 5-way contract-token lock)
- Integration (key-gated, real spend): acceptance run quoted above
- Post-merge: grading unit suite on the integration branch

Closes #591


azorej:
hygiene-review failure is the known repo-wide infra breakage (#425 / #398) — same init-death signature as every milestone PR. Merging on the PR's own green lanes (lint ✅, test-smoke ✅).

Issue #687, opened 2026-07-29
test: no runner-side differential for numeric_string_fields folding (hash-verdict half done by #686)


azorej:
## Context

Found while revising the #675 substrate-parity plan against review. #675's differential parity test proves a `grading.yaml` key discriminates (satisfied scores above violated) on **both** grading substrates. For most keys the runner side is a pure function (`evaluate_transcript_rules`, `evaluate_jsonpath_checks`) callable at `canonical` tier with no services.

The hash family is the exception. The runner's hash evaluator is `RunnerServiceImpl._execute_hash_grading` (`runner/service.py:1539`), a service method that drives the db-service over HTTP: get stable hash → snapshot → reset → replay golden actions → restore. `compute_stable_hash` lives in the db-service (`env/json_db_service/app.py`) and `core/hash.py:258`. It cannot be exercised without the core Docker stack, and mocking the DB client to fake it would violate AGENTS.md rule 5 ("mocks hide problems, test real behavior") on the one test whose job is to detect substrate drift.

There is also no existing integration test to point the guard at. The only two integration tests that mention hash config are `tests/integration/test_docker_grading_jsonpath.py:80` (`hash_enabled: False` — deliberately off) and `tests/integration/test_runner_cleanup_trial_grpc.py:46` (`hash_enabled: True, golden_actions: []`, asserting cleanup behaviour, not a grade). Neither asserts a hash-graded verdict.

## Consequence

`state_checks.hash` (and its nested keys) and `state_checks.numeric_string_fields` land in the #675 manifest with `enforcement: FIELD_RESOLUTION_ONLY` — the guard proves the field exists and resolves on both config models, but not that both substrates actually score it. `numeric_string_fields` is the sharper gap: `docs/GRADING.md` explicitly promises it is "honored identically on both grading substrates (the core `GradingEngine`/`to_hashable` path and the runner gRPC/`compute_stable_hash` path)", and that promise is currently untested end-to-end on the runner side.

## Fix

An integration-tier test in the shape of `tests/integration/test_docker_grading_jsonpath.py` (testcontainer fixtures, real gRPC, no LLM, deterministic): `RegisterTrial` a task with `hash_enabled: True` and golden actions over a known `initial_state`, `GradeTrial`, and assert the verdict — one case matching the golden state and one diverging from it. Then a second pair differing only in whether a numeric-looking string field is listed in `numeric_string_fields`, which pins the documented folding promise on the production path.

[...]

Issue #686, opened 2026-07-29
fix(grading): state_checks.hash.weight is core-only, and the hash-vs-jsonpath combination rule differs between substrates


azorej:
## Context

Found while revising the #675 substrate-parity plan against review. Two related asymmetries in how the two grading substrates fold `state_checks.hash` together with `state_checks.jsonpaths`. Both are the class of drift #675's manifest exists to record, but fixing them changes verdicts and so does not belong inside #675.

## 1. `state_checks.hash.weight` has no runner counterpart

Core reads it: `core/grading/combine.py:122-124` → `hash_weight = hash_config.get("weight", 1.0)`, threaded into `StateChecker.grade` / `grade_tau_style`.

The runner has no weight concept anywhere on its hash path — `_execute_hash_grading` (`runner/service.py:1539-1546`) takes only `trial_id`, `trial_context`, `golden_actions`, `numeric_string_fields`, and the runner `StateChecksConfig` declares no hash-weight field. A pack that sets `state_checks.hash.weight` is honoured on the core substrate and silently ignored in production.

## 2. The combination rule itself differs

Given both a hash score and a jsonpath score, the two substrates combine them differently:

- **Core** — a weighted blend (`core/grading/state_checks.py:347` and `:555`):
  ```python
  final_score = (jsonpath_score * (1 - hash_weight)) + (hash_score * hash_weight)
  ```
- **Runner** — a product, with no weight (`runner/grading.py:838-840`, mirrored for the reported component at `runner/service.py:1242`):
  ```python
  active_components["state_checks"] = hash_score * jsonpath_score
  ```

For a task with `hash` and `jsonpaths` both active, these disagree for essentially every input. With the default `hash_weight=1.0`, core ignores the jsonpath score entirely (`1.0 * hash_score`) while the runner multiplies it in — so a task passing every hash check and half its jsonpath assertions scores `1.0` on core and `0.5` on the runner.

This is the deeper of the two problems: #1 is an ignored knob, #2 means the two substrates disagree on the composite `state_checks` score whenever both sources are configured.

## Why it is filed separately from #675

#675 records the asymmetry honestly in the key manifest (`state_checks.hash.weight` → `CORE_ONLY`, `state_checks.hash` → signal parity only, both pointing here). Fixing it means choosing one combination rule and moving verdicts for every task that configures hash and jsonpaths together, which is a deliberate scoring change, not a guard-rail bugfix.

## Fix

[...]

Issue #685, opened 2026-07-29
fix(grading): core transcript scoring averages four always-present buckets, so a disallowed-tool violation scores 0.75 instead of failing


azorej:
## Context

Discovered while planning #675 (substrate parity). Not that ticket's scope — #675 makes the two substrates *consume* the same keys; this issue is about the core substrate's *aggregation* being wrong independently of parity.

`TranscriptChecker.grade` (`tolokaforge/core/grading/transcript.py:96-134`) computes:

```python
scores = [contain_score, regex_score, turns_score, tools_score]
final_score = sum(scores) / len(scores)
```

Four buckets, always present, always weighted 1/4 each — regardless of how many rules the author actually declared.

## Two consequences

**1. A hard violation cannot fail the component.** A pack declaring `disallowed_tools: [delete_customer]` whose agent calls `delete_customer` scores `transcript_rules = 0.75`, not `0.0`:

```
CORE transcript verdict, disallowed tool called: 0.75
  | Missing required tools: write_file; Used disallowed tools: delete_customer
```

`check_tool_expectations` correctly returns `0.0` for the violation, but the three unconfigured buckets each contribute a free `1.0`.

**2. Adding a rule dilutes every other rule.** Because the denominator is a constant `4` rather than the number of declared rules, a task that declares only `max_turns` is graded 3/4 on rules it never wrote.

The runner substrate does not have this defect — `evaluate_transcript_rules` (`tolokaforge/runner/grading.py:250-286`) decomposes the config into one sub-check per declared field entry and scores `passed / total`, so an empty config is a no-op pass and a single violated rule in a single-rule config scores `0.0`.

## Why it is filed separately

Fixing it changes core-substrate verdicts for every task with `transcript_rules`, which moves canonical snapshots (`tests/canonical/snapshots/grading_paths/*.json`, `tests/canonical/test_grading_canon.py`). That is a deliberate, reviewable verdict change and does not belong inside a bugfix whose whole point is that the two substrates stop drifting silently.

## Fix

Give the core `TranscriptChecker` the same per-declared-sub-check decomposition the runner already uses: `score = passed_sub_checks / total_sub_checks`, empty config → `1.0`. That also makes `transcript_rules` a `BOTH_SCORE_PARITY` key in the #675 manifest instead of signal-parity-only.

## Acceptance

- A single-rule `transcript_rules` config whose rule is violated scores `0.0` on the core substrate.
- An empty `transcript_rules` config scores `1.0`.
[...]

Issue #679, opened 2026-07-29
feat: Trace-check validation — load-time schema checks and a pre-run gate


azorej:
## Context

A mistyped tool name, a nonexistent argument, an uncompilable regex, or an unsatisfiable count bound in a trace check produces a silently mis-graded run — or, worse, grades the *agent* down for the author's typo (this is the existing behaviour for an invalid disallow-regex, which becomes a failing sub-check rather than an authoring error). Because trace checks reference the task's own tool schemas, almost all of these are catchable before anything runs.

Task validation already exists as a CLI command, and the run pipeline already has a precedent for a pre-flight abort that inspects every selected task and refuses to start while naming the offending tasks. This ticket generalises both to cover trace checks.

## Deliverables

- [ ] Task validation rejects trace-check config with: unknown/duplicate constraint ids, dangling references, tool names absent from the task's declared tool set, argument names absent from the referenced tool's schema, argument value types that contradict the schema, operators applied to incompatible types, regexes that do not compile, and empty (match-everything) matchers
- [ ] Task validation rejects semantically unsatisfiable constraints: self-ordering, contradictory count bounds, absence-required matchers used in ordering, single-path alternatives
- [ ] The existing grading validation entry point is generalised from its current judge-only scope to validate every grading component
- [ ] A **pre-run gate** validates trace checks for every selected task and aborts the run before any trial executes, naming the offending tasks and the specific problem
- [ ] Hard schema errors always abort; advisory findings are toggleable via run config and default to on
- [ ] The tier-1 tool-name and argument-name sweep also covers `transcript_rules.tool_expectations` (`required_tools` / `disallowed_tools`), not only `trace_checks` matchers
- [ ] The sweep rejects `state_checks.hash.expected_state_hash` declared without `hash.enabled: true` — a silent no-op on **both** substrates today, so not drift, but it cannot be intentional
- [ ] Weighting is validated: negative weights, a non-gate weight set summing to zero (a division by zero), a weight declared on a gate constraint, and a path with no constraints all fail at load
- [ ] Docs: what is validated, when, and how to interpret each failure

## Acceptance Criteria

- A task with a mistyped tool name in a trace check fails `validate` with a message naming the tool and listing the task's real tools
- A task with an argument name absent from the tool's schema fails `validate`
- An uncompilable regex fails at validate time rather than grading the agent down at run time
- Starting a run with any invalid trace-check task aborts before the first trial and names every offending task
[...]

PR #699, opened 2026-07-30
fix(grading): substrate-parity contract for every grading.yaml key

### LINKED ISSUE #675: fix: tool_expectations and env_assertions are silently ignored on the runner grading path
## Context

Two `grading.yaml` keys are honoured by the core `GradingEngine` (the substrate canonical tests exercise) but never consumed by the runner, which is the **production** grading path. A task that ships `transcript_rules.tool_expectations` or `state_checks.env_assertions` today gets **zero grading signal from that key, with no warning** — the value is either dropped during task-description translation or accepted by the runner config and never evaluated.

This is a silent-failure violation of AGENTS.md rule 1, and it is exactly the class of drift the incoming trace-checks work must not repeat. Fixing it first establishes the guard rail.

## Deliverables

- [ ] `transcript_rules.tool_expectations` (`required_tools` / `disallowed_tools`) produces real grading signal on the runner path
- [ ] `state_checks.env_assertions` produces real grading signal on the runner path
- [ ] A **substrate-parity test** enumerates every supported `grading.yaml` key and fails when a key is consumed by one grading substrate but not the other
- [ ] Any `grading.yaml` key that reaches the runner and cannot be evaluated fails loud rather than no-opping
- [ ] `docs/GRADING.md` and `docs/REFERENCE.md` describe the true behaviour of both keys, per substrate

## Acceptance Criteria

- A task pack declaring `disallowed_tools` that the agent then calls **fails** on a runner-backed run (today it passes)
- A task pack declaring an `env_assertion` grades differently when that assertion is violated (today the score is identical)
- Adding a new `grading.yaml` key to only one substrate fails CI with a message naming the key and the missing substrate
- Every example pack either keeps its current score, or the change is explained in the PR

## Design Decisions

- **Implement both keys rather than delete them**, because `disallowed_tools` is minefield-shaped behaviour the trace-checks work needs regardless, and `env_assertions` is tau2 compatibility surface we advertise. Deleting them loudly is an acceptable fallback if implementation turns out to be disproportionate — silently leaving them is not.
- **Parity is enforced by a test, not by convention**, because the drift already happened silently on two separate keys. Convention demonstrably does not hold here.
- **Unevaluatable keys fail loud**, because the failure mode being fixed is precisely a config key that looks honoured and is not.

## Out of Scope

- New matching operators, ordering, or any part of `trace_checks` itself
- Reworking how the two grading substrates are structured — this ticket makes them agree, it does not merge them

## Open Questions

- Confirm the implement-rather-delete call above. If either key should be dropped instead, say so before work starts — it materially changes the ticket.

azorej:
## Summary

- **Every author-facing `grading.yaml` key is now enumerated in one machine-readable manifest** that both grading substrates and the tests consume, declaring per key what kind of key it is, which substrates evaluate it, at what enforcement tier, with which evaluator — and when only one substrate consumes it, why, plus the issue that closes the gap.
- **`transcript_rules.tool_expectations` produces real grading signal on the production runner path.** A task declaring `disallowed_tools` that the agent then calls now fails a runner-backed run; previously it passed clean at `score=1.0`.
- **`state_checks.env_assertions` and `state_checks.db_hash_check` are rejected at config load** with a migration error naming the replacement. Neither ever produced grading signal on either substrate.
- **A populated-but-unaccounted scored key now fails `GradeTrial` loud** instead of grading silently, and every deliberate skip is recorded with its reason and surfaced in `grade.reasons`.

## Why this exists

Two grading substrates evaluate a trial — the in-process core `GradingEngine` that `tests/canonical/` exercises, and the runner's gRPC `GradeTrial` path that production actually grades on. Nothing enumerated which keys each one consumes, and the YAML→runner translation is hand-written key by key with no completeness check. **There was no artifact that could be wrong**, so four author-facing keys had drifted into silent no-ops without anything noticing.

Enumerating them turned up a consistent pattern worth stating plainly: **the core substrate systematically scores more leniently than the runner.** A green canonical suite has therefore been overstating the shipped verdict, in several independent places. The follow-ups below are that finding, itemised.

## Design choices

- **The manifest carries two orthogonal axes, not one.** `KeyKind` (scored check / config input / aggregation) and `Enforcement` (canonical differential / integration differential / field resolution only). Coverage alone conflates three different kinds of key: config inputs are consumed at `RegisterTrial`, before `GradeTrial` exists, so they can never be "evaluated". Scoping both the differential test and the runtime ledger to scored checks is what keeps each honest — an earlier draft without it would have failed `GradeTrial` on **every task using `id_fields`**.
- **The exemption sets live in the test module, never beside the manifest**, each asserted as set equality against a computed set. A constant sitting next to the data it guards can be widened by the same data change it exists to catch. Widening an exemption now requires a reviewable edit to a test file in the same commit.
- **Keys whose real evaluator needs live services are tiered off canonical-tier differential enforcement rather than mocked.** The runner's hash evaluator drives the db-service over HTTP; `db_probes` opens a real connection per probe. Mocking either to make an anti-drift test pass would defeat the guard's purpose.
- **Parity fixtures are `grading.yaml` text driven through the real loaders**, not hand-built config objects. The two substrates use different names for the same author key, and the drift happened in the code that reconciles them — a hand-built fixture would mirror that aliasing by hand and pass while the real translation drops the key.
- **The runtime ledger records skips with reasons, not just evaluations.** Four legitimate skip paths exist. Treating a skip as unaccounted would error trials that grade correctly today, including degenerate ones — which would collide with #677, whose whole point is that those score badly rather than error.
- **"Populated" means non-empty, not present**, because a dump that excludes defaults cannot distinguish an explicitly-written empty list from an unset field, and an empty check has nothing to evaluate. The same boundary decides the removal in stage 3: a pack that *asks for* removed behaviour fails loud; an inert default still loads.
- **`env_assertions` was removed rather than implemented.** Both keys arrived in the initial open-source release, and `tolokaforge/tasks/` has never existed in this repository on any branch — zero commits ever touched that path — so the assertion modules its evaluator resolved were never present. It was dead on arrival, not a regression. Because the load failure hard-zeroed the *entire* `state_checks` component, a pack declaring it scored `0.0` regardless of agent behaviour, so **removal can only raise an affected pack's score**. The external task-pack corpus could not be checked from this repo; the CHANGELOG is written to teach an affected author what happened rather than to claim the corpus was verified clean.
- **Shared models are declared runner-side and re-exported core-side.** `core/models.py` already imports a dozen names *from* `runner/models.py` at top of file, so the reverse direction is a cycle no import ordering can break — and the import-boundary test would **not** catch it, because `core.models` is a permitted runner surface, making it a plain runtime failure rather than a lock violation.

## Concepts introduced

**Substrate parity** — the property that a `grading.yaml` key is consumed by both grading substrates, declared per key rather than assumed, with asymmetries recorded as classified, issue-tracked exemptions instead of living as undocumented drift.

**Key kind vs enforcement** — two orthogonal axes. *Kind* says what a key is for; *enforcement* says how strongly parity is proven for it. Separating them is what lets a key whose evaluator needs live services be honestly declared rather than either mocked or silently exempted.

**Accounted keys** — a grade-time ledger distinguishing three states: *evaluated*, *skipped for this stated reason*, and *populated but unaccounted*. Only the third is a failure. The distinction is the whole mechanism: it converts "nothing read this key" from a silent pass into a loud, named failure.

**The enumeration ceiling** — documented explicitly rather than left implied. The guard introspects declared model fields, so author keys consumed through untyped dicts are structurally outside it. A green parity suite proves each key *discriminates*; it does not prove the discrimination is *correct*.

## Plan

# Plan: substrate-parity guard rail + `tool_expectations` on the runner path

Issue: #675 (sub-issue of epic #684, milestone 28 "Deterministic trace checks")
Branch: work lands on the integration branch `feat/deterministic-trace-checks` (already created off `main`) — **not** `main`

## Context

Two `grading.yaml` keys were reported as silently ignored on the runner (production)
grading path. Reproducing them by running (not reading) confirmed both reports and
uncovered **two more instances of the same drift, plus a worse problem with one of the
two reported keys**. Every claim below was produced by executing the code, and the exact
output is quoted.

**Reproduction 1 — `transcript_rules.tool_expectations` is dropped in translation.**
`NativeAdapter` loads it into the core config and never passes it to the runner:

```
CORE   transcript_rules: {..., 'tool_expectations': {'required_tools': ['browser', 'write_file'], 'disallowed_tools': []}, ...}
RUNNER transcript_rules: {'must_contain': [], 'disallow_regex': [], 'max_turns': 15, 'required_actions': [], 'communicate_info': []}
RUNNER TranscriptRulesConfig fields: ['must_contain', 'disallow_regex', 'max_turns', 'required_actions', 'communicate_info']
```

(`tests/data/tasks/browser_basic` through `NativeAdapter.get_grading_config` vs
`NativeAdapter.to_task_description`.) The runner substrate then grades a violation as a
clean pass:

```
RUNNER transcript verdict with a called disallowed tool: passed=True score=1.0 details=[]
CORE   transcript verdict same violation: 0.75 | Missing required tools: write_file; Used disallowed tools: delete_customer
```

**Reproduction 2 — `state_checks.env_assertions` is worse than reported: it is
non-functional on *both* substrates.** The runner declares it and never evaluates it
(`env_assertions` appears nowhere in `runner/grading.py` or the grading path in
`runner/service.py`). The core substrate *attempts* it and always fails, because
`EnvironmentEvaluator.load_assertion_function` resolves assertion functions from
`tolokaforge.tasks.<domain>.assertions` / `tolokaforge/tasks/<domain>/assertions.py` —
and `tolokaforge/tasks/` **does not exist in this repository** (task packs are external,
Known Gotcha #10). Both the satisfying and the violating state hard-zero:

```
CORE SATISFIED: score=0.0 state_checks=0.0 pass=False
   reasons: Assertion failed: assert_order_cancelled({'order_id': 'O1'}) - Error evaluating assertion:
            Failed to load assertions module for domain retail: Assertion module not found:
            /workspaces/tolokaforge/tolokaforge/tasks/retail/assertions.py
CORE VIOLATED:  score=0.0 state_checks=0.0 pass=False
   reasons: <identical>
```

So `env_assertions` has never produced grading signal anywhere. **Git archaeology confirms
it was dead on arrival, not a regression**: both keys arrived in the initial open-source
release commit (`2151fff`, 2026-04-17), and `tolokaforge/tasks/` has never existed in this
repository's history — `git log --all -- 'tolokaforge/tasks'` returns zero commits and
`git log --all --diff-filter=A -- '**/assertions.py'` returns none, so no assertion module
has ever existed on any branch. There is also a key-name split in the translation:
`native.py:658` reads `assertion.get("tool_name")` into `func_name`, while the core
`StateChecksConfig` parses `func_name` — the same YAML block needs two different key names
depending on which substrate reads it.

**Two further drifted keys, not in the ticket.**

- `state_checks.db_probes` — declared on the **core** `StateChecksConfig`
  (`core/models.py:1449`) and never read by the core `GradingEngine` (`db_probes` does
  not appear in `core/grading/combine.py`). Inverse-direction drift.
- `state_checks.db_hash_check` — declared and evaluated core-side
  (`core/models.py:1448`, `environment_evaluator.py:193`), **dropped in translation** to
  the runner exactly like `tool_expectations`. It also silently passes when enabled with
  no expected hash — `tests/unit/grading/test_evaluators.py:242-250` pins
  `score == 1.0` for `db_hash_check=True` with no hash, an AGENTS.md rule 1 violation
  baked into a test.

**And a fifth, found during revision (now #686):** `state_checks.hash.weight` is read
core-side (`combine.py:122-124`) and has no runner counterpart at all — but the deeper
problem is that the two substrates use *different composition rules* for hash against
jsonpaths. Core blends (`state_checks.py:347`, `:555`:
`jsonpath_score * (1 - hash_weight) + hash_score * hash_weight`); the runner multiplies
(`grading.py:840`: `hash_score * jsonpath_score`). At the default `hash_weight=1.0` a
task passing every hash check and half its jsonpath assertions scores `1.0` core-side and
`0.5` on the runner.

**Why the drift was invisible.** The adapter's YAML→runner translation
(`native.py:636-754`) is hand-written key by key with no completeness check, and nothing
anywhere enumerates the supported key set. There is no artifact that can be wrong.

**Blast radius on in-repo packs is nil.**
`grep -rn "env_assertions\|tool_expectations\|db_hash_check" examples/` returns no matches
across all 27 `examples/**/grading.yaml` files. Only
`tests/data/tasks/{browser_basic,calc_basic}` declare `tool_expectations`. **The external
task-pack corpus is NOT covered by this measurement** — `tasks/` does not exist in this
checkout (gotcha #10), so it could not be greped from here. See Stage 3's dispatch
precondition.

## Goal

1. Every author-facing `grading.yaml` key is enumerated in **one machine-readable
   manifest** that both substrates and the tests consume, declaring per key what kind of
   key it is, which substrate evaluates it, with which evaluator, at what enforcement
   tier, and — when not both — *why not*.
2. Four tests make the manifest load-bearing: a key cannot be added to either config
   model without an entry; a scored key claiming both substrates must demonstrably move
   both substrates' scores; a key claiming the runner must survive adapter translation;
   the exemption set is frozen in the test module so widening it is a reviewable edit.
3. `transcript_rules.tool_expectations` produces real grading signal on the runner path,
   including a `disallowed_tools` violation failing a runner-backed run.
4. A populated *scored-check* key that reaches the runner with neither an evaluator nor a
   recorded skip fails `GradeTrial` loud, naming the key — never a no-op.
5. `env_assertions` and `db_hash_check` stop pretending to work.

## Non-goals

- Any part of `trace_checks` — matchers, operators, ordering (#678).
- Load-time validation of grading config against the task's declared tool set — #679 owns
  it (epic §3.9 tier 1 names it explicitly).
- Merging the two grading substrates, or unifying their *aggregation*. This plan makes
  them agree on **which keys they consume**; it does not make every key's numeric score
  identical. Where scores differ, the manifest says so out loud, with a tracking issue.
- Resurrecting `custom_checks` on the runner (epic #684 §7, a sandboxing project).
- Implementing `db_probes` on the core substrate (Stage 1 declares the architectural
  reason).
- Fixing the hash/jsonpath composition asymmetry (#686) or adding runner-side hash
  differential coverage (#687). Both are recorded in the manifest with tracking issues.

---

### Stage 1: The substrate-parity contract

The load-bearing artifact of the whole issue and the guard rail the rest of milestone 28
depends on. Lands first, with the current drift recorded as **explicit, reasoned,
issue-tracked exemptions in a frozen set** so the stage is green — later stages remove
exemptions one at a time, and each removal automatically switches on enforcement for that
key.

#### Contract

New module `tolokaforge/core/grading/key_manifest.py`. In `core/grading/*`, which is one
of the core surfaces the runner is permitted to import
(`tests/canonical/test_runner_import_boundary.py` lock 2), so the Stage 4 runtime ledger
can consume it without inverting the dependency direction.

Types per the AGENTS.md type table — in-process values that never cross a serialisation
boundary, so `@dataclass(frozen=True)` + `str, Enum`, **not** Pydantic:

```python
class KeyKind(str, Enum):
    SCORED_CHECK   # produces a component score from a trajectory / final state
    CONFIG_INPUT   # shapes how another check behaves; no score of its own
    AGGREGATION    # combines component scores into the final score

class SubstrateCoverage(str, Enum):
    BOTH_SCORE_PARITY   # both substrates consume it AND produce the same component score
    BOTH_SIGNAL_PARITY  # both consume it and both discriminate; aggregation differs
    CORE_ONLY
    RUNNER_ONLY

class Enforcement(str, Enum):
    DIFFERENTIAL_CANONICAL    # test 3 proves the differential in-process, no services
    DIFFERENTIAL_INTEGRATION  # the differential needs services; `enforcing_test` names it
    FIELD_RESOLUTION_ONLY     # only "the field exists and resolves"; requires tracking_issue

@dataclass(frozen=True)
class GradingKey:
    author_key: str                 # dotted grading.yaml path, e.g. "transcript_rules.tool_expectations"
    kind: KeyKind
    coverage: SubstrateCoverage
    enforcement: Enforcement
    core_field: str | None          # dotted *model attribute* path; None if the substrate does not declare it
    runner_field: str | None
    core_dict_key: str | None = None    # see "Path resolution" below
    runner_dict_key: str | None = None
    core_evaluator: str | None = None   # dotted path of the function that consumes it
    runner_evaluator: str | None = None
    enforcing_test: str | None = None   # required when enforcement is DIFFERENTIAL_INTEGRATION
    reason: str = ""                    # required (non-empty) unless coverage is a BOTH_* value
    tracking_issue: int | None = None   # see "Exemption classes" below

GRADING_KEYS: tuple[GradingKey, ...]
def author_keys() -> frozenset[str]: ...
def entry(author_key: str) -> GradingKey: ...            # raises KeyError, never returns None
def scored_keys_claiming_runner() -> tuple[GradingKey, ...]: ...   # what Stage 4's ledger iterates
```

**Why `kind` exists.** `SubstrateCoverage` alone says *who* consumes a key; it does not
say whether a satisfying/violating fixture is even a meaningful thing to build. It is not
for `combine.pass_threshold` (a cut, not a check), `combine.weights`, `combine.method`,
`state_checks.id_fields` or `state_checks.relaxed_validation` (both consumed at
`RegisterTrial` / load time — `native.py:668-673`, `service.py:618-635`,
`id_resolution.py:146` — and absent from `_grade_trial_async`'s component phase entirely).
Scoping the differential and the Stage 4 ledger to `kind == SCORED_CHECK` is one edit that
makes both correct.

**Path resolution, including untyped dicts.** `core_field` / `runner_field` always name a
path that resolves by `model_fields` walking, e.g. `StateChecksConfig.jsonpaths`. When an
author key lives *inside* an untyped dict field, the manifest names the dict field in
`*_field` and the key inside it in `*_dict_key`: `state_checks.hash.enabled` is
`core_field="StateChecksConfig.hash"`, `core_dict_key="enabled"`,
`runner_field="StateChecksConfig.hash_enabled"`, `runner_dict_key=None`. Test 1 resolves
`*_field` and, when `*_dict_key` is set, additionally asserts the resolved field is
dict-typed. The dict-key half is **declared data, not introspection-verified** — which is
the guard's honest ceiling, documented in `docs/GRADING.md` (see Doc updates).

**Container fields.** `GradingConfig.state_checks`, `.transcript_rules`, `.llm_judge` and
`GradingCombineConfig` are containers, not keys. Test 1 walks *into* them and claims their
leaves; the container fields themselves are listed in an explicit
`_CONTAINER_FIELDS` frozen constant in the test module and excluded. `llm_judge` is the
one exception: it is a container whose whole subtree is runner-evaluated, so it gets a
single entry and its leaves are not separately claimed — stated explicitly so two
implementers cannot resolve it differently.

**Exemption classes — the freeze covers BOTH axes.** Four frozen constants **in the test
module** (never in `key_manifest.py` — a constant living beside the data it guards could be
widened by the same data change it is supposed to catch), each asserted as set equality
against the manifest, so widening any of them requires an edit a reviewer sees in the same
commit:

*Coverage axis — who consumes the key:*

- `_ARCHITECTURAL_EXEMPTIONS` — non-`BOTH_*` keys that can never be both. Each **must**
  have `tracking_issue is None` and a non-empty `reason`.
- `_DRIFT_EXEMPTIONS` — non-`BOTH_*` keys that *should* be both and are not yet. Each
  **must** have a non-`None` `tracking_issue`.

*Enforcement axis — whether the claim is actually proven:*

- `_NON_DIFFERENTIAL_SCORED_KEYS` — scored keys that claim both substrates but are **not**
  differentially proven in-process. Asserted as:

  ```python
  assert _NON_DIFFERENTIAL_SCORED_KEYS == {
      e.author_key for e in GRADING_KEYS
      if e.kind is KeyKind.SCORED_CHECK
      and e.coverage.startswith("BOTH")
      and e.enforcement is not Enforcement.DIFFERENTIAL_CANONICAL
  }
  ```

- `_NON_TRACKED_FIELD_RESOLUTION_KEYS` — the declared `AGGREGATION` / `CONFIG_INPUT`
  entries allowed to be `FIELD_RESOLUTION_ONLY` without a tracking issue (`combine.*`,
  `id_fields`, `relaxed_validation`), so every *other* `FIELD_RESOLUTION_ONLY` entry must
  carry one.

**Why the enforcement axis needs its own freeze.** The coverage freeze alone leaves the
round-1 escape hatch intact in new clothes: an entry declared `BOTH_SIGNAL_PARITY` +
`FIELD_RESOLUTION_ONLY` + any non-`None` `tracking_issue` sits in *neither* coverage set,
is skipped by test 3, and is green. Concretely — and this is the outcome the guard rail
exists to prevent — **#678 could add `trace_checks` to the runner `GradingConfig` and as a
dead field on the core one (`extra="ignore"`, never read), declare
`BOTH_SIGNAL_PARITY` / `FIELD_RESOLUTION_ONLY` / a core follow-up issue, and CI would
pass**, because test 1 proves a field *resolves*, not that it is *evaluated*. That is
precisely the `db_probes` drift shape, reproduced on the milestone's keystone issue.
`_NON_DIFFERENTIAL_SCORED_KEYS` closes it: at Stage 1 that set is exactly the hash family
(`state_checks.hash` + nested, tracking #687), so it is cheap and self-documenting, and
adding `trace_checks` to it is a visible test-file edit that a reviewer must approve.

Together these are the answer to "`reason` being a non-empty string is not a gate": the
gate is the frozen sets, not the prose, on both axes.

#### Manifest state at the end of Stage 1

| `author_key` | kind | coverage | enforcement | note |
|---|---|---|---|---|
| `combine.method` / `.weights` / `.pass_threshold` | `AGGREGATION` | `BOTH_SCORE_PARITY` | `FIELD_RESOLUTION_ONLY` — no tracking issue needed; aggregation is not differentially testable by construction, which the test asserts rather than exempts | core nests under `combine.*`, runner flattens to `combine_method` / `weights` / `pass_threshold` |
| `state_checks.hash` (+ `.enabled`, `.expected_state_hash`, `.golden_actions`) | `SCORED_CHECK` | `BOTH_SIGNAL_PARITY` | `FIELD_RESOLUTION_ONLY`, tracking **#687** | the runner's evaluator is `_execute_hash_grading`, a service method driving db-service over HTTP; no service-free differential exists and mocking the DB client on the anti-drift test would violate AGENTS.md rule 5 |
| `state_checks.hash.weight` | `CONFIG_INPUT` | `CORE_ONLY` | `FIELD_RESOLUTION_ONLY`, tracking **#686** | `_DRIFT_EXEMPTIONS`. Core blends hash against jsonpaths by this weight; the runner multiplies them and has no weight concept |
| `state_checks.jsonpaths` | `SCORED_CHECK` | `BOTH_SIGNAL_PARITY` | `DIFFERENTIAL_CANONICAL` | |
| `state_checks.numeric_string_fields` | `CONFIG_INPUT` | `BOTH_SCORE_PARITY` | `FIELD_RESOLUTION_ONLY`, tracking **#687** | GRADING.md:93-95 promises identical behaviour on both paths; the runner side reaches it only through `_execute_hash_grading` (`service.py:1061`) |
| `state_checks.id_fields` / `.relaxed_validation` | `CONFIG_INPUT` | `BOTH_SCORE_PARITY` | `FIELD_RESOLUTION_ONLY` — no tracking issue; consumed at load / `RegisterTrial`, never in the component phase | |
| `state_checks.db_probes` | `SCORED_CHECK` | `RUNNER_ONLY` | `DIFFERENTIAL_INTEGRATION`, `enforcing_test="tests/integration/test_helpdesk_workflow_end_to_end.py"` | `_ARCHITECTURAL_EXEMPTIONS`: the probe DSN (e.g. `app-db:5432`) resolves only inside the task's docker network, which the runner container joins and the host-side core engine does not. The core config retains the field for round-trip fidelity (pinned by `test_native_adapter_canon.py:216`) and must not evaluate it. **Not canonical-tier testable**: `evaluate_db_probes` opens a real connection per probe (`runner/grading.py:696-712`) and treats a connection failure as a *failed* probe, so satisfying and violating fixtures both score `0.0` in a service-free test and `score_satisfied > score_violated` would be false. The cited test already asserts the primitive produced a `DB probes:` reasons segment with no connection/driver error (`:195`, `:201`), so this costs a citation, not a new test |
| `state_checks.env_assertions` | `SCORED_CHECK` | `CORE_ONLY` | `FIELD_RESOLUTION_ONLY` | `_DRIFT_EXEMPTIONS`, tracking #675 — removed in Stage 3 |
| `state_checks.db_hash_check` | `SCORED_CHECK` | `CORE_ONLY` | `FIELD_RESOLUTION_ONLY` | `_DRIFT_EXEMPTIONS`, tracking #675 — removed in Stage 3 |
| `transcript_rules.must_contain` / `.disallow_regex` / `.max_turns` / `.required_actions` / `.communicate_info` | `SCORED_CHECK` | `BOTH_SIGNAL_PARITY` | `DIFFERENTIAL_CANONICAL` | signal, not score: core averages four fixed buckets while the runner scores declared sub-checks — **#685** |
| `transcript_rules.tool_expectations` | `SCORED_CHECK` | `CORE_ONLY` | `FIELD_RESOLUTION_ONLY` | `_DRIFT_EXEMPTIONS`, tracking #675 — flipped in Stage 2 |
| `llm_judge` (whole subtree, single entry) | `SCORED_CHECK` | `RUNNER_ONLY` | `DIFFERENTIAL_INTEGRATION`, `enforcing_test="tests/integration/test_rubric_judge_live.py"` | `_ARCHITECTURAL_EXEMPTIONS`: the rubric judge runs runner-side on the shared `ToolCallingLoop`; `GradingEngine` deliberately leaves `components.llm_judge` unset (`combine.py:233-235`) |
| `custom_checks` | `SCORED_CHECK` | `CORE_ONLY` | `FIELD_RESOLUTION_ONLY`, tracking **#684** | `_DRIFT_EXEMPTIONS`: the runner reports `custom_checks=-1.0`; resurrecting it is a sandboxing project, out of scope per epic §7 |
| `grading_method` | `AGGREGATION` | `RUNNER_ONLY` | `FIELD_RESOLUTION_ONLY` | `_ARCHITECTURAL_EXEMPTIONS`: a runner-side dispatch selector with no `grading.yaml` counterpart. Live external producer: `external_adapters/tolokaforge-adapter-terminal-bench/.../adapter.py:246` sets `grading_method="test_execution"`. Records that this dispatch mode bypasses key-level evaluation entirely (`service.py:1014` returns before the component phase) — the declared reason Stage 4's ledger does not apply to it |

Six `SCORED_CHECK` + `BOTH_*` + `DIFFERENTIAL_CANONICAL` entries exist at Stage 1
(`jsonpaths`, `must_contain`, `disallow_regex`, `max_turns`, `required_actions`,
`communicate_info`), so the differential test has real content on landing.

#### Behaviour to lock

New `tests/canonical/test_grading_substrate_parity.py`, tier `canonical` (contract/schema
shape, no services, no LLM):

1. `test_manifest_covers_every_declared_config_field` — introspect `model_fields` on the
   core `GradingConfig` / `StateChecksConfig` / `TranscriptRulesConfig` /
   `GradingCombineConfig` and the runner equivalents, walking into containers and skipping
   `_CONTAINER_FIELDS`. Assert every leaf is claimed by exactly one manifest entry, and
   every `core_field` / `runner_field` resolves (plus the dict-typed assertion when a
   `*_dict_key` is set). **This is the mechanism that stops #678's `trace_checks` landing
   on one substrate only**: adding the field to either model fails the test naming the
   field and the model.
2. `test_exemption_sets_are_frozen_and_classified` — all four frozen constants asserted as
   set equality against the manifest (see "Exemption classes" above): the non-`BOTH_*`
   author keys equal `_ARCHITECTURAL_EXEMPTIONS | _DRIFT_EXEMPTIONS`; architectural entries
   have `tracking_issue is None` and a non-empty `reason`; drift entries have a non-`None`
   `tracking_issue`; `_NON_DIFFERENTIAL_SCORED_KEYS` equals the computed set; every
   `FIELD_RESOLUTION_ONLY` entry outside `_NON_TRACKED_FIELD_RESOLUTION_KEYS` carries a
   `tracking_issue`; every `DIFFERENTIAL_INTEGRATION` entry names an `enforcing_test` that
   exists on disk; every entry matching **test 3's predicate** has both evaluators set and a
   fixture; every declared evaluator path is importable and callable.
3. `test_both_substrates_discriminate_each_shared_scored_key` — the differential. **Test
   3's predicate, stated once and used verbatim by test 2:**
   `kind is SCORED_CHECK and coverage.startswith("BOTH") and enforcement is
   DIFFERENTIAL_CANONICAL`. All three conjuncts are required — `SCORED_CHECK` because a
   threshold or a load-time config input has no violating trajectory, `BOTH_*` because a
   single-substrate key has nothing to compare, and `DIFFERENTIAL_CANONICAL` because
   `db_probes` and the hash family need real services. Test 2 must derive its
   "has a fixture" obligation from this same predicate rather than restating a looser
   `BOTH_*` scoping, or the two tests disagree about which entries owe a fixture. For each
   matching entry, run its `(config, satisfying, violating)` fixture through **each
   substrate's real production evaluator plus its real combine** —
   `GradingEngine.grade_trajectory` core-side, and `evaluate_transcript_rules` /
   `evaluate_jsonpath_checks` + `combine_grade_components` runner-side. Two constraints
   that make the assertion non-vacuous:
   - **The fixture config declares only the key under test.** Otherwise a violating case
     that also violates a second evaluated key discriminates without the key under test
     ever being read.
   - The assertion is on **that key's own sub-check / component**, not only the aggregate
     score: `score_satisfied > score_violated` on both substrates, and for
     `BOTH_SCORE_PARITY` the two substrates' component scores are equal on both fixtures.

   No mocks, no fakes, no re-implemented aggregation — the runner functions used are the
   same objects `_grade_trial_async` calls. **The implementer must not build a helper that
   mirrors `_grade_trial_async`'s hash/jsonpath precedence**; that is why the hash family
   is `FIELD_RESOLUTION_ONLY` (#687) rather than forced into this test.
4. `test_adapter_translation_carries_every_runner_key` — the test that would have caught
   this bug at its source. A fixture pack declaring every manifest key with a
   `runner_field`, run through `NativeAdapter.to_task_description`; assert each such key is
   non-default on the resulting runner `GradingConfig`. `tool_expectations` and
   `db_hash_check` were dropped exactly here.

**Fixture construction — YAML through the adapter, not hand-built config objects.** Test 3's
`(config, satisfying, violating)` fixtures start as `grading.yaml` text and reach each
substrate through its real loader: `NativeAdapter.get_grading_config` for the core side and
`NativeAdapter.to_task_description` for the runner side. Hand-built config objects would
mirror the substrates' key aliasing by hand — `required_actions[].name` core-side versus
`.tool_name` runner-side, `jsonpaths` versus `jsonpath_checks` — and could therefore pass
while the real translation drops the key, which is the exact bug this ticket exists to fix.
Driving both substrates from one YAML string also makes test 3 and test 4 mutually
reinforcing rather than two independent hand-maintained mirrors.

Fixture pack: `tests/data/grading_parity/all_keys/` with its **own** glob, deliberately
outside the `tasks/**/task.yaml` pattern that **20+ test modules** across `tests/unit`,
`tests/canonical` and `tests/integration` enumerate — a new pack under
`tests/data/tasks/` would perturb their pack counts.

- **Compatibility:** internal only. New module, new test module, new test fixture; no
  author-facing surface change and no behaviour change to grading.
- **Deliverable:** `tolokaforge/core/grading/key_manifest.py`,
  `tests/canonical/test_grading_substrate_parity.py`,
  `tests/data/grading_parity/all_keys/`, and a "Substrate parity" section in
  `docs/GRADING.md`.
- **Validation:** dev MCP `run_tests` marker `canonical` and marker `unit` (full lanes —
  the manifest touches shared models, and the 20+ pack-enumerating modules must stay
  green); `lint_check`; `format_check`.
- **Doc updates:** `docs/GRADING.md` — new "Substrate parity" section covering the two
  substrates, the manifest as the single source of truth, the `kind` / `coverage` /
  `enforcement` vocabulary, the current exemption table, and **the guard's enumeration
  ceiling stated plainly**: `model_fields` introspection cannot see author keys consumed
  through untyped dicts, and all four core models are `extra="ignore"`, so
  `state_checks.hash.*`, the `state_checks.jsonpaths[*]` operator vocabulary, and
  `custom_checks.*` internals are structurally outside the guard; a green parity suite
  proves each key *discriminates*, not that the discrimination is *correct*. **Rewrite**
  the two existing ad-hoc parity claims so they reference the section instead of restating
  the promise locally: the `numeric_string_fields` sentence (GRADING.md:93-95) and the
  `id_fields` belt-and-suspenders paragraph (GRADING.md:148-151). `tests/README.md` — one
  line for the parity guard and what a failure means.

### Stage 2: `tool_expectations` produces real grading signal on the runner path

- **Contract:**
  - New `class ToolExpectations(BaseModel)` — `required_tools: list[str] = []`,
    `disallowed_tools: list[str] = []`, `model_config = {"extra": "forbid"}`. **Declared in
    `tolokaforge/runner/models.py` and re-exported from `tolokaforge/core/models.py`**
    alongside the existing re-export block at `core/models.py:32-43`, so there is one
    definition and no cycle.

    The direction is forced by the existing dependency, and getting it backwards is an
    ImportError, not a lock violation: `core/models.py:32-43` already imports twelve names
    from `runner.models` at **top of file** (`Criterion`, `CriterionResult`,
    `EnvironmentManifest`, `EnvironmentPatch`, `JudgeCustomization`, `LLMJudgeConfig`,
    `ResetSpec`, `Rubric`, `ServiceIsolation`, `ServiceNetworkAccess`, `ServiceSpec`,
    `StackPatch`), while `runner/models.py` imports only `core.deprecations` and
    `core.netpolicy_constants` — never `core.models`. Declaring `ToolExpectations` core-side
    and importing it runner-side would make `core.models → runner.models → core.models` a
    cycle that import ordering cannot break, because `core/models.py`'s import of
    `runner.models` is top-of-file and no class can be defined ahead of it. Note that
    `test_runner_import_boundary.py` lock 2 **permits** `core.models` as a runner import
    surface, so the boundary test would not catch this — it fails at runtime on a
    partially-initialised module.

    `LLMJudgeConfig` (`core/models.py:37`) is the exact analogue and the convention to
    follow: a nested grading-config block shared by both substrates, declared runner-side
    because the runner config is the `extra="forbid"` wire surface, re-exported core-side so
    there is a single definition.

    Declared on the runner `TranscriptRulesConfig` as
    `tool_expectations: ToolExpectations | None = None`.
  - The core `TranscriptRulesConfig.tool_expectations` is retyped from
    `dict[str, list[str]] | None` to that same re-exported model. `extra="forbid"` on the
    nested block inside the `extra="ignore"` parent is deliberate: today a `required_toolz`
    typo is silently dropped on both substrates.
  - **The retype breaks `core/grading/combine.py:214-224`**, which calls
    `.get("required_tools")` / `.get("disallowed_tools")` on the dict. Converting those to
    attribute access is part of this stage's deliverable, not an incidental fix.
  - `NativeAdapter.to_task_description` populates the runner field from
    `grading.yaml`'s `transcript_rules.tool_expectations`.
  - `evaluate_transcript_rules` decomposes it into **one sub-check per declared tool**,
    matching how it already decomposes `must_contain` / `disallow_regex`:
    - `rule_type="required_tool"`, one per `required_tools` entry — passes iff a call to
      that tool with `status == "success"` appears in `tool_history`. Same "a failed call
      did not happen" rule `_check_required_action` already applies.
    - `rule_type="disallowed_tool"`, one per `disallowed_tools` entry — passes iff **no**
      call to that tool appears in `tool_history` **at any status**. An attempted
      forbidden call is a violation even when it errored; that is the point of a minefield
      check.
  - **No undeclared-tool check, and no new parameter on `evaluate_transcript_rules`.**
    Round-1 review was right that "a named tool absent from `agent_tools ∪ user_tools` is
    a failing sub-check" needs a tool set the function does not have, and that adding it
    runner-side only would create a fresh asymmetry under a `BOTH_SIGNAL_PARITY`
    declaration. It is also **already owned by #679**, which validates "tool names absent
    from the task's declared tool set" at *load* time — strictly better than a grade-time
    failing sub-check, and cheaper. Dropped here rather than half-built. The resulting
    known limitation — `required_tools: [typo]` is indistinguishable from "the agent never
    called it", and `disallowed_tools: [typo]` passes trivially — is documented in
    `docs/GRADING.md` with #679 named as the fix. **#679's tier-1 validation sweep is being
    extended to cover `transcript_rules.tool_expectations` specifically**, so this deferral
    has a named landing place rather than being an open-ended promise.
  - Manifest entry flips to `BOTH_SIGNAL_PARITY`, `enforcement=DIFFERENTIAL_CANONICAL`,
    `tracking_issue=None`, both evaluators named, and its author key is removed from
    `_DRIFT_EXEMPTIONS`. Stage 1 tests 2, 3 and 4 begin enforcing the key automatically.
    Signal, not score, parity: the core substrate's four-bucket average is #685.
- **Behaviour to lock:**
  - `unit` (extend `tests/unit/grading/test_grading_correctness.py`, which already drives
    `evaluate_transcript_rules`): required tool present → pass; required tool absent →
    named failing sub-check; required tool called but `status="error"` → fail; disallowed
    tool never called → pass; disallowed tool called with `status="success"` → fail;
    disallowed tool called with `status="error"` → fail; an empty `ToolExpectations`
    contributes no sub-checks (so it cannot dilute the fraction).
  - `canonical`: Stage 1 tests 3 and 4 now cover the key. `test_native_adapter_canon.py`
    gains an explicit `to_task_description` assertion that `browser_basic`'s
    `tool_expectations` reaches the runner config. **No `grading_config.json` snapshot
    changes are expected** — both fixture packs declare `required_tools` *and*
    `disallowed_tools` explicitly, so the typed model dumps identically. If a snapshot
    does move, stop and report it rather than regenerating.
  - `integration`, new `tests/integration/test_docker_grading_tool_expectations.py`,
    modelled on `tests/integration/test_docker_grading_jsonpath.py` (`requires_docker`, no
    LLM, no API keys, deterministic): `RegisterTrial` a task declaring one **builtin**
    agent tool that needs no env service — pick from the tools the runner can dispatch
    without db-service or rag-service, per `tests/unit/test_runner_builtin_dispatch.py` —
    with `disallowed_tools: [that tool]`; `ExecuteTool` it over real gRPC so the runner
    records it in `tool_call_history`; then `GradeTrial` → assert `binary_pass is False`
    and the reasons name the tool. **This is the ticket's headline acceptance criterion
    and the only test that proves the `service.py` wiring rather than the evaluator in
    isolation.** Add the mirror case (declared but not called → pass) so the test
    discriminates.
- **Compatibility — a real migration, name it:** the runner `TranscriptRulesConfig` is a
  gRPC wire surface with `extra="forbid"`, so a new engine emitting `tool_expectations`
  against an **older runner image** fails `RegisterTrial` validation. Same version-lock
  class as `id_fields`. Old engine + new runner stays safe. Migration: a CHANGELOG entry
  naming the key and the version lock, plus a "Runner-engine version lock" line in the new
  GRADING.md `tool_expectations` section worded like the existing one at GRADING.md:148-151.
- **Deliverable:** `tool_expectations` graded on the production path; a task pack that
  declares `disallowed_tools` and calls one fails a runner-backed run.
- **Validation:** `run_tests` markers `unit` + `canonical`; then the new integration test.
  **The runner image must be rebuilt (`make docker-build-core`) before the integration
  test** — the new model field ships inside the image. A stage report that omits the
  integration run has not validated this stage.
- **Doc updates:** `docs/GRADING.md` — a `transcript_rules` section documenting
  `tool_expectations` semantics per substrate (sub-check decomposition, error-status
  handling, the undeclared-tool limitation with #679 named, version lock), written as if
  this is the only behaviour there has ever been. `docs/REFERENCE.md:139-141` and
  `docs/CONFIG.md:321-323` — the existing `grading.yaml` examples stay, gaining the
  per-substrate note. `docs/TASK_DESCRIPTION_SCHEMA.md:240-246` — the
  `TranscriptRulesConfig` block gains the field. `CHANGELOG.md`.

### Stage 3: `env_assertions` and `db_hash_check` stop pretending to work

**Dispatch precondition — this stage must not be dispatched until both are true:**

1. The human has arbitrated the implement-vs-remove fork (still open; see Risks). Plan of
   record below is loud removal; the alternative is fully specified after Stage 4.
2. **The external task-pack corpus has been greped for `env_assertions` and
   `db_hash_check`, and the result quoted in the stage report.** The Context measurement
   covers in-repo `examples/**` only — `tasks/` does not exist in this checkout, so the
   corpus that Core Rule 5 actually protects was not measured. A hit changes the
   migration from "delete a key nothing uses" to "delete a key with live users", which
   needs a real migration path and probably a different decision. Whoever runs it should
   grep the pack repositories the run configs point `task_packs` at.

- **Contract:**
  - `state_checks.env_assertions` and `state_checks.db_hash_check` are **rejected at
    config load** with a migration error naming the replacement — `jsonpaths` for
    per-record state assertions, `hash: {enabled: true}` for whole-state comparison,
    `db_probes` for substrate SQL. Implemented as a `@model_validator(mode="before")` on
    the core `StateChecksConfig`, in the same shape as
    `LLMJudgeConfig._reject_legacy_rubric_shape` (`runner/models.py:538-575`) — the repo's
    existing precedent for a loud removal.
  - Deleted in this stage, with no shims and no `_legacy_*` aliases: `EnvAssertion` (core
    and runner models), `StateChecksConfig.env_assertions` and `.db_hash_check` on both
    models, `tolokaforge/core/evaluators/environment_evaluator.py` in full, the
    `use_tau2_evaluator` branch in `combine.py:91-107` (dead once both keys are gone), the
    `env_assertions` translation block in `native.py:652-663`, and
    `TestEnvironmentEvaluator` in `tests/unit/grading/test_evaluators.py`.
  - `rg 'env_assertions|db_hash_check|EnvAssertion|EnvironmentEvaluator'` returns hits only
    inside the migration-error text and the CHANGELOG. **Running that `rg` is part of the
    stage, not a nicety.**
  - Manifest loses both entries and both author keys leave `_DRIFT_EXEMPTIONS`; Stage 1
    tests 1 and 2 prove the manifest still matches the (now smaller) models.
- **Behaviour to lock:**
  - `unit`: a `grading.yaml` declaring `env_assertions` raises at load with `jsonpaths`
    named in the message; same for `db_hash_check` with `hash` named. One test per key,
    asserting on the remediation text, not just the exception type.
  - `canonical`: `grading_config.json` snapshots for all five snapshot packs lose the two
    keys — regenerate with `update_canonical_snapshots` and enumerate the diff in the
    stage report. **This is the only stage expected to move snapshots.**
- **Compatibility — a task-pack compatibility surface (AGENTS.md Core Rule 5), so the
  migration is explicit:** zero in-repo `examples/**` packs declare either key (verified
  across 27 `grading.yaml` files), so **no example pack's score changes**; the external
  corpus is covered by dispatch precondition 2. Nothing regresses even for a pack that
  does declare `env_assertions`: the key hard-zeroes the entire `state_checks` component
  today (`environment_evaluator.py:125` → ImportError → `passed=False` →
  `assertion_score=0.0`), so removing it can only raise such a pack's score. This is not a
  behaviour the key ever had and then lost — it arrived broken in the initial open-source
  release (`2151fff`) and `tolokaforge/tasks/` has never existed on any branch, so **no
  external pack can ever have had a working `env_assertions`** either; a corpus hit under
  precondition 2 would find a pack being silently zeroed, not a pack being served.
  Migration =
  the loud load-time error naming the replacement + a CHANGELOG entry listing both keys,
  why they are gone (never produced signal on either substrate), and what replaces each.
- **Deliverable:** neither key can be declared; the dead evaluator and the
  `func_name`/`tool_name` translation split are gone.
- **Validation:** `run_tests` markers `unit` + `canonical`; `validate_tasks` over
  `examples/**` to prove no example pack trips the new error; `lint_check`; the `rg` sweep;
  the external-corpus grep from precondition 2.
- **Doc updates:** `docs/TASK_DESCRIPTION_SCHEMA.md` — delete the `env_assertions` field
  (line 237) and its worked example (lines 621-625); the block reads as if the new schema
  is the only schema, with no "previously" note. `docs/GRADING.md` — the parity exemption
  table shrinks by two rows. `CHANGELOG.md`.

### Stage 4: a populated-but-unaccounted-for scored key fails `GradeTrial` loud

Deliverable 4 of the ticket, and the belt-and-braces that makes this class of drift
unshippable even if someone bypasses the tests.

- **Contract:**
  - `_grade_trial_async` accumulates an explicit `accounted_keys: dict[str, str]` mapping
    author key → `"evaluated"` or `"skipped: <reason>"`, **populated at the points where
    each evaluator is invoked or deliberately skipped** — not by re-deriving which
    evaluators "should have" run, which would be a second copy of the dispatch logic and
    could itself drift (compare #422, the same duplication problem in the neighbouring
    `state_checks` precedence).
  - **Record per author key, not per evaluator call.** `evaluate_transcript_rules` is one
    function covering six author keys, so marking all six `"evaluated"` because the function
    ran would record a *populated but never decomposed* key as evaluated — a false negative
    in exactly the drift class this stage exists to catch, and the shape a future
    `TranscriptRulesConfig` key would take. The function already branches per field
    (`grading.py:252-275`), so each branch records its own author key as it produces
    sub-checks. Same rule for `evaluate_jsonpath_checks`. This direction produces no false
    positives either way, so it is correctness-of-coverage rather than a safety fix.
  - **The ledger covers `kind == SCORED_CHECK` only.** `CONFIG_INPUT` and `AGGREGATION`
    keys are excluded by construction: `id_fields` and `relaxed_validation` are consumed
    at `RegisterTrial` before `GradeTrial` exists (`service.py:618-635`),
    `numeric_string_fields` reaches only `_execute_hash_grading` (`service.py:1061`) so a
    jsonpath-graded task legitimately never "evaluates" it, and `combine.*` is the
    aggregation itself. Without this scoping the ledger would return `success=False` for
    **every `id_fields` task** — the whole #600/#603 line of work.
  - **The three legitimate skip paths record a skip, not nothing.** `transcript_rules` is
    skipped when there are no messages *and* no tool history
    (`service.py:1131-1147` `else:` branch) and `llm_judge` when `llm_messages` is empty
    (`service.py:1203-1204`). Both branches record
    `"skipped: no transcript messages or tool history"` / `"skipped: no transcript
    messages"` into `accounted_keys` for every author key under that component, and the
    reason is surfaced in `grade.reasons` so the skip becomes visible instead of silent.
    A degenerate trial must **score badly, not error the RPC** — erroring it would collide
    head-on with #677.

    The **third** skip site is the `hash_enabled` gate at `service.py:1049-1052`, and it is
    the subtle one: `native.py:679-680` gates `hash_enabled` on `hash.enabled` but populates
    `expected_hash` from `hash.expected_state_hash` **unconditionally**. So
    `hash: {expected_state_hash: "…"}` with no `enabled: true` produces a runner config
    where `expected_hash` is populated and truthy while the evaluator never runs — a
    populated, runner-claiming `SCORED_CHECK` key with nothing accounted, which without this
    bullet would error the RPC on a config that grades fine today. The `else:` side of that
    gate records `"skipped: hash grading not enabled"` for **every** hash-family author key
    (`state_checks.hash` and its nested leaves) so the family is accounted atomically rather
    than leaf by leaf.

    Latent, not live: no in-repo pack declares `expected_state_hash` as grading input (the
    only hits are recorded output bundles under `tests/data/projects/tau_retail_mini/`), and
    `native.py` is the only adapter populating runner state checks. But the Stage 4
    compatibility claim below is only true *because* this site is enumerated, so it is a
    contract bullet, not a footnote.
  - `grading_method == "test_execution"` returns at `service.py:1014` before the component
    phase, so the ledger does not apply to that dispatch mode. This is a **declared**
    exemption recorded on the manifest's `grading_method` entry, not an accident of
    control flow.
  - After the component phase, compute the scored author keys the request's grading config
    actually populated and subtract `accounted_keys`. **Populated means non-empty, not
    present**: `model_dump(exclude_defaults=True)` cannot distinguish
    `disallowed_tools: []` written explicitly from unset, and an empty check has nothing
    to evaluate — so a key counts as populated only when its value is truthy after the
    dump.
  - A non-empty remainder returns `GradeTrialResponse(success=False, error=...)` naming
    each unaccounted key and, from its manifest entry, the expected `runner_evaluator` or
    its declared absence. **Never a score**, matching how the errored judge is excluded
    rather than folded in as `0.0`.
  - A key the manifest declares `CORE_ONLY` that nonetheless arrives populated is the same
    loud failure, quoting that entry's `reason`.
- **Behaviour to lock:**
  - `unit`: a runner `GradingConfig` populating a scored manifest key whose entry names no
    runner evaluator yields `success=False` with the key in the error text; **a config
    shaped like the in-tree `native_widgets_id_fields` pack (`numeric_string_fields` +
    `id_fields` + `relaxed_validation` + `jsonpath_checks` + `must_contain`) yields a
    grade, not an error** — the explicit regression test for the false-positive class
    round-1 review found; a degenerate trial with `transcript_rules` and no messages and
    no tool history yields a grade whose reasons carry the skip note; **a config carrying
    `expected_hash` with `hash_enabled=False` yields a grade whose reasons carry
    `skipped: hash grading not enabled`, not an error** — the third-skip-site regression
    test; a fully-evaluated config yields the same score it produced before this stage.
  - `canonical`: the ledger's runner-field → author-key mapping resolves for every
    manifest entry claiming the runner, so a future entry with a malformed `runner_field`
    fails here rather than at grade time in production.
- **Compatibility:** a deliberate behaviour change on the gRPC surface — a task that
  previously graded while silently ignoring a *scored* key now fails the RPC loudly. That
  is the ticket's point. No in-tree pack is affected, and the claim rests on three named
  exclusions rather than on absence of evidence: Stage 3 removed the only two keys that
  could otherwise trigger it, `CONFIG_INPUT` / `AGGREGATION` keys are out of scope by
  `kind`, and all three skip sites (`transcript_rules` no-input, `llm_judge` no-messages,
  `hash_enabled` false) record an explicit skip. CHANGELOG entry.
- **Deliverable:** an unevaluatable scored grading key can no longer no-op in production.
- **Validation:** `run_tests` markers `unit` + `canonical`; re-run the Stage 2 integration
  test after `make docker-build-core` to prove a normal graded trial is unaffected.
- **Doc updates:** `docs/GRPC_PROTOCOL.md` — `GradeTrial` error semantics gain the
  unaccounted-key failure, and the two skip paths. `docs/GRADING.md` — the parity section
  gains the runtime-ledger paragraph including its `SCORED_CHECK` scoping, so the doc
  describes both the static guard and the runtime guard as the current design.

---

## Discovered issues

**Fix in this PR:**

- `state_checks.db_hash_check` dropped in adapter translation, and its
  `tests/unit/grading/test_evaluators.py:242-250` test pinning `score == 1.0` for
  "enabled with no expected hash" — a silent pass baked into a test. Removed in Stage 3.
- `native.py:658` reads `assertion.get("tool_name")` into `EnvAssertion.func_name` while
  the core model parses `func_name` — one YAML block, two key names. Disappears with
  Stage 3.
- `core/grading/combine.py:214-224` dict access on `tool_expectations`, converted with the
  Stage 2 retype.
- `state_checks.db_probes` (core-declared, never core-evaluated), `custom_checks`
  (core-only, `-1.0` runner stub), `llm_judge` (runner-only by design), and
  `grading_method` (runner-only dispatch selector) become declared, classified, reasoned
  exemptions in a frozen set rather than undocumented asymmetries.

**Filed as issues:**

- **#685** — core `TranscriptChecker.grade` averages four always-present buckets
  (`core/grading/transcript.py:131-132`), so a `disallowed_tools` violation scores `0.75`
  instead of failing, and declaring one more rule dilutes every other rule. Milestone 28.
  Not fixed here: it moves core-substrate verdicts and canonical snapshots for every task
  with `transcript_rules`. Closing it upgrades `transcript_rules` from
  `BOTH_SIGNAL_PARITY` to `BOTH_SCORE_PARITY`.
- **#686** — `state_checks.hash.weight` is core-only, and more seriously the two
  substrates use different composition rules for hash against jsonpaths (core blends by
  `hash_weight`, `state_checks.py:347`/`:555`; the runner multiplies, `grading.py:840`), so
  they disagree on the composite `state_checks` score whenever both sources are
  configured. Milestone 28. Tracked by the manifest's `state_checks.hash.weight` entry.
- **#687** — no runner-side hash-grading verdict test exists, so the parity guard can only
  field-resolve the hash family (`state_checks.hash*`, `numeric_string_fields`) rather than
  prove the differential; the runner's evaluator is a service method driving db-service
  over HTTP, and mocking it on the anti-drift test would violate AGENTS.md rule 5.
  Milestone 28. Tracked by those manifest entries' `FIELD_RESOLUTION_ONLY` enforcement.

**Noted, not filed:**

- #422 (already open) is the same duplication pattern in the neighbouring `state_checks`
  precedence logic. Stage 4's ledger deliberately does not add a third copy.
- **`hash.expected_state_hash` without `enabled: true` is a silent no-op on *both*
  substrates**, found while verifying Stage 4's third skip site. Core gates reading it on
  `hash_config.get("enabled", False)` (`combine.py:114-117`); the runner receives it
  populated but gates the evaluator on `hash_enabled` (`service.py:1052`). Not drift — the
  two substrates agree — so it is outside this ticket's remit and needs no manifest entry.
  It is squarely a #679 tier-1 load-time check ("declaring `expected_state_hash` without
  `enabled: true` cannot be intentional"), and worth adding to that sweep alongside
  `tool_expectations`. Not filed separately to avoid duplicating #679.
- `combine_grade_components`' `actually_configured` heuristic
  (`runner/grading.py:865-881`) returns `0.0, False` with a **warning log** when grading
  was configured but nothing was evaluated. Stage 4's ledger supersedes it for the
  unaccounted-key case; the heuristic itself is left alone.

---

## Risks / open questions

- **Stage 3 needs two things before dispatch, and the fork is NOT decided.** (a) The human
  arbitrates implement-vs-remove. The ticket's open question was resolved as "implement
  both", but that resolution explicitly authorised the escape hatch if discovery showed
  implementation was disproportionate for one key. It did, for `env_assertions`, with the
  evidence in Context and the four obstacles above. Plan of record is loud removal;
  flipping is a one-stage swap. (b) The external pack corpus must be greped
  (precondition 2) — this repo cannot answer it, and a hit would change the answer to (a).
  Stages 1, 2, and 4 are unaffected either way.
- **Stage 2 requires a runner image rebuild** before its integration test can pass. A
  stage implementer that runs only `unit` + `canonical` will report green while the
  headline acceptance criterion is unverified.
- **Stage 2 introduces an engine↔runner-image version lock** (`extra="forbid"` on the
  runner `TranscriptRulesConfig`), same class as `id_fields`. Documented, not avoided — the
  alternative is loosening a wire contract that exists to catch exactly this.
- **The guard's ceiling is real and must be documented, not glossed.** Three routes still
  reach one substrate only: a declared exemption (green by design — mitigated, not closed,
  by the frozen exemption set); a key with no typed model field, i.e. anything inside
  `state_checks.hash`, the `jsonpaths[*]` operator vocabulary, or `custom_checks`, which
  `model_fields` cannot see; and a vacuous `BOTH_*` fixture (mitigated by the
  single-key-per-fixture rule). **If #678 grows matcher operators inside untyped dicts, the
  guard will not see them** — `trace_checks` must be a typed model for the guard to hold,
  and that requirement belongs in #678's contract.
- **A green parity suite proves each key discriminates on both substrates, not that the
  discrimination is correct.** Per-key correctness stays with each key's own tests. Stated
  in the GRADING.md section so nobody reads green as a correctness proof.
- **The parity fixture pack must live outside `tests/data/tasks/`.** 20+ test modules
  across `tests/unit`, `tests/canonical` and `tests/integration` enumerate that directory
  with `tasks/**/task.yaml`; adding a pack there perturbs their counts.
- Pre-existing, not ours: an untracked, un-gitignored `judge_mismatch_artifacts.tar.gz`
  (75K) in the repo root, violating AGENTS.md root cleanliness. Do not stage it, do not
  delete it, and never `git add -A` on this branch.

## Test plan

Post-merge, in rough order of what would fail first if something is wrong:

- `canonical` lane — **726 passed**, five locks in `tests/canonical/test_grading_substrate_parity.py`. Adding a field to either substrate's grading config without a manifest entry fails lock 1; declaring a key covered-on-both while exempting it from differential testing fails the frozen-set equality; a ledger key with no recording site fails lock 5.
- `unit` lane — **3571 passed**. Eight failures in `tests/unit/test_persistent_shell_local.py` are pre-existing and environment-sensitive (bracketed-paste escape codes), verified identical on a clean worktree at the base commit.
- **Integration, needs a rebuilt runner image**: `make docker-build-core`, then `tests/integration/test_docker_grading_tool_expectations.py` — 3 passed. This is the headline criterion: a pack declaring `disallowed_tools` that the agent then calls fails a runner-backed run over real gRPC. A *stale* image does not skip; it fails at `RegisterTrial` with an extra-forbid error naming the field.
- **Version-skew check before deploying**: `state_checks` now requires engine and runner image from the same release in *both* directions. An engine older than this release emits `env_assertions` for every pack with a non-empty `state_checks:` block, so an old engine against a new image is rejected at `RegisterTrial` for *every* such trial — not only for packs that used the key.
- `tolokaforge validate` over `examples/**` — 27 valid, 2 invalid. The 2 are pre-existing `examples/terminal_bench/**` `TaskConfig` failures on missing fields, identical on `main`, unrelated to grading.
- No canonical snapshot moved except in the removal commit, where exactly five `grading_config.json` files lost exactly the two removed keys — 10 deletions, 0 insertions, 0 value changes.

## Review notes

Three sharded reviewers ran in parallel over two fix rounds. Final verdicts: correctness **APPROVE**, hygiene **APPROVE WITH NITS**, type-fit **APPROVE**.

The two findings worth knowing about, because both were guards that did not guard:

- **A lock that could not fail.** The first attempt at the accounting-site guard claimed six keys via a set comprehended from the manifest — the same source it was asserting against. It was proven green-in-CI / red-in-production with an injected manifest entry. Now claimed via the evaluator's own per-site constants, and proven red both by dropping a constant and by reproducing the original injection.
- **A key reported as evaluated that nothing reads.** The ledger folded `state_checks.hash.expected_state_hash` into the hash family's outcome and stamped it evaluated, while the manifest declares no runner path reads it. Atomicity is right for members the evaluator consumes; folding in a declared-dead key is masking. It now carries a standing skip whose reason reaches the author.

Closes #675


azorej:
## CI status and the basis for merging

Recording this so the merge decision is auditable rather than implicit.

### What ran and passed

- **`lint`** — pass
- **`test-smoke`** — pass (3m6s). This is the substantive PR lane: `tests/unit/` + `tests/canonical/` with coverage, the `tools/` suites, public-example validation, and a Docker task-pack mount smoke.

### What did not run, and why none of it is about this branch

- **`test-full`** — skipped by design; its condition is `github.event_name != 'pull_request'`, so it is a push-event lane.
- **`test-gate`** — skipped, and it skips for **everyone, always**. Its condition requires a `ready-to-merge` label that does not exist in this repository, and it is `skipped` in **40 of the last 40** `ci.yml` runs. It is the only PR lane that builds images and runs the integration tier, so no PR has ever had integration tests run in CI. Filed as #700.
- **`hygiene-review`** — failed, but **systemically**: four consecutive failures across two unrelated branches over two days, with the comment "Claude encountered an error after 0s" and no analysis performed. Reproduced on `feat/rate-limit-probe-mode` before this branch existed. Filed as #701.

### How the integration gap was covered instead

The tier `test-gate` would have run was verified locally against a purpose-rebuilt runner image (`make docker-build-core` after the final source edit, digest confirmed):

- `tests/integration/test_docker_grading_tool_expectations.py` — **3 passed**. This is the ticket's headline criterion: a pack declaring `disallowed_tools` that the agent then calls fails a runner-backed run over real gRPC, where it previously passed clean at `score=1.0`.
- `tests/integration/test_docker_grading_jsonpath.py` — **2 passed**, exercising the same `_grade_trial_async` path the new ledger sits in.

That verification was reproduced independently on two separately-built images during the run.

### Review basis

Three sharded reviewers over two fix rounds. Final verdicts: correctness **APPROVE**, hygiene **APPROVE WITH NITS**, type-fit **APPROVE**. The two most consequential findings were both guards that did not guard — a lock that compared a set against itself and so could never fail, and a ledger entry reporting a key as evaluated that the manifest declares nothing reads. Both are fixed and both were proven red.

### Merging on this basis

Every lane that fires is green; the two that do not fire, do not fire for any PR, and both are now filed as P1. This merges into the integration branch `feat/deterministic-trace-checks`, not `main` — the consolidation PR to `main` remains the human review gate, so this is reversible.

azorej:
## Integration verification of the manifest's own citations

With provider budget authorised I ran the tests this branch cites as parity evidence — which nothing had done, because the check validating those citations only asserts the file **exists**.

**Run:** `test_docker_grading.py`, `test_helpdesk_workflow_end_to_end.py`, `test_lot_ops_end_to_end.py`, `test_rubric_judge_live.py` — 16m36s, real provider calls. Result: **13 passed, 1 failed, 2 errors, 0 skipped.**

### The `db_probes` citation is sound
`test_helpdesk_workflow_end_to_end.py` — cited as the `enforcing_test` for `state_checks.db_probes` — **passes**. That exemption's evidence is real.

### The `llm_judge` citation is partly dead — filed as #702
`test_rubric_judge_live.py::test_rubric_judge_runs_inside_slim_runner_image` fails: its inline container script imports `run_rubric_judge`, a symbol removed when the judge was lifted to a Protocol (#521). The only remaining references are inside that test's own string literal.

**Pre-existing and present on `main`** — the same three references exist there — and this branch never touched `judge.py` (`git diff` against the base is empty for that file). Not branch-caused.

The interesting part is why nothing caught it: lock 2 asserts every cited `enforcing_test` **exists on disk**. It does. It contains a test that cannot pass. So the manifest's evidence for one of its largest exemptions is partly decorative, and the guard built to keep exemptions honest cannot see it — **the same defect class this branch exists to eliminate, one level up.** That is the third instance of the pattern found during this work, after a lock that compared a set against itself and a ledger key reported as evaluated that nothing reads.

Strengthening lock 2 to verify a citation *passes* rather than *exists* is the durable fix and is captured in #702. It is deliberately not in this PR: it would mean running integration tests from a canonical-tier lock, which is a design decision rather than a correction.

### The two errors are environmental, not code
- `test_docker_containers_healthy` — 120 s health-endpoint timeout on a cold cache, the #697 class reproducing in a fresh environment.
- `test_rubric_judge_live_agent_prompt_gating_omits_policy` — Docker 403 on network teardown, "network has active endpoints". A cleanup race.

### Effect on the merge basis
Strengthens it. The integration tier that `test-gate` would have run (#700) has now been exercised well beyond this branch's own surface, and every failure traces to a pre-existing defect or the environment. Merging.

PR #703, opened 2026-07-30
fix(docker): auto host ports for rag/mock-web; persist rag HF cache on volume

azorej:
## Problem

Hit in real benchmark runs from `tolokaforge-tools` (2026-07-30):

1. **`full_stack()` hardcodes host ports 8080 (mock-web) and 8001 (rag-service)** with no run-config plumbing to override them. Any unrelated host process on those ports (IDEs commonly hold 8080) kills the whole run with `PortAllocationError` — after up to an hour of image building. `db_port`/`runner_port` already default to `"auto"`.
2. **rag-service re-downloads its sentence-transformers model on every container recreation** — the HF cache defaults to `~/.cache` *inside* the container, so the `rag_data` volume never caches it despite the code comment claiming "downloads on first run, then cached". Cold starts regularly exceeded the 120s health timeout, failing runs.

## Changes

- `rag_port` / `mock_web_port` default to `"auto"` (consistent with db/runner). Nothing outside the stack consumes these host ports — in-stack consumers use the `mock-web` / `rag-service` network aliases.
- **`{port:<container_port>}` placeholder in HTTP probe URLs**, resolved by `EngineStack._resolve_health_probe` after port allocation. Without this, auto ports force `health_probe=None` and the deferred-probe fallback (port-8000 convention, fixed 30s) silently drops rag-service's long cold-start timeout.
- `HF_HOME=/env/rag/hf_cache` on the existing `rag_data` volume, making the model cache actually persistent; probe timeout 120→300s to cover the one remaining first-ever download on slow networks.

## Validation

- New unit tests for placeholder resolution (`tests/unit/test_stack_health_probe_placeholder.py`); targeted unit + canonical suites pass (80 tests), ruff clean.
- Both fixes were run in anger as local venv patches during a day of benchmark runs in `tolokaforge-tools` (see toloka-partners/tolokaforge-tools#56) — mock-web came up on an auto port with 8080 occupied, and rag-service went healthy in ~2 min cold / seconds warm with the volume-backed cache.

### DIFF
--- a/tests/unit/test_stack_health_probe_placeholder.py
+++ b/tests/unit/test_stack_health_probe_placeholder.py
@@ -7,8 +7,12 @@ port-8000 fallback — e.g. rag-service's long cold-start timeout, which a
 30s default probe would break on first-run model downloads.
 """
 
+import pytest
+
 from tolokaforge.docker.health import HealthProbe
 from tolokaforge.docker.stack import EngineStack
+
+pytestmark = pytest.mark.unit
 
 
 def test_placeholder_resolved_from_port_map():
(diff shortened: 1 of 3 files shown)


azorej:
All three points addressed in 1e56db7:

- **🔴 Marker blocker** — `pytestmark = pytest.mark.unit` added. You called the root cause exactly: my "80 tests" validation invoked the file by explicit path, which bypasses `-m unit` deselection. Verified fixed: the 4 tests are now selected and pass under `pytest tests/unit/ -q -m unit`.
- **🟡 Silent probe drop** — now raises `ValueError` on an unresolved placeholder (agreed: a placeholder always ships with a matching `PortConfig`, so no mapping is a programming error, and dropping the probe reintroduces the exact silent-skip failure mode this PR targets). Test renamed to `test_placeholder_with_unresolved_port_raises` with `pytest.raises`.
- **🟡 getattr nit** — switched to `isinstance(probe, HttpHealthProbe)`, which also narrows the type for the `model_copy` call.

Full local `-m unit` run: 3692 passed, 8 failed — the 8 are all `test_persistent_shell_local.py` and fail identically on clean `main` in my environment (local-shell-dependent), unrelated to this change. Canonical suites + ruff clean.

Issue #674, opened 2026-07-29
Remove dead custom-checks convenience API (run_custom_checks / result_to_score)


CiroGamboa:
## Context

While finishing the custom-checks runner wire (#669), two symbols in `tolokaforge/core/grading/check_runner.py` are effectively dead:

- `run_custom_checks(...)` — a module-level convenience wrapper. Its only references are the re-exports in `tolokaforge/core/grading/__init__.py` (`__all__`) and `check_runner.py` `__all__`. No production or test caller.
- `CheckRunner.result_to_score(...)` — reachable only from inside `run_custom_checks` and a couple of unit tests; the production grade path uses `CheckResultSet` directly.

Neither is on the documented author API (authors write decorated functions in `checks.py`; they never call these). They are internal helpers that happen to be exported.

## Proposal

Delete `run_custom_checks` and its export, and fold `result_to_score` into the caller or drop it, once #669 confirms the live grade path builds results without them. Because the symbols are in a public `__all__`, treat the removal as a compatibility-surface change: CHANGELOG entry + a note in `docs/custom_checks.md`. Small; kept out of #669 to keep that PR focused.

Relates to #669, #217.

azorej:
Scope note from planning #895.

#895 introduces a shared `custom_checks` sentence for `Grade.reasons`, emitted by both grading substrates. `CheckRunner.result_to_score` renders the same `CheckResultSet` into `Custom checks: {passed}/{total} passed, score: {x}` (`check_runner.py:600`) — the same prefix — with score semantics that contradict both substrates (`0.5` on a non-fatal error, `1.0` for a suite that declared nothing, where production leaves the component unscored at `-1.0`). Measured at `b3ca7fa1`: zero production callers, and `run_custom_checks` does **not** call it (it returns `runner.run(...)` directly at `check_runner.py:722`), so it comes out independently.

#895 therefore removes `result_to_score` and its three test cells (`test_custom_checks.py::TestResultToScore`, `test_custom_checks_runner.py::test_result_to_score`), with the CHANGELOG + `docs/custom_checks.md` treatment this issue specifies.

**Remaining scope here:** `run_custom_checks` and its export. Note that it is not inert the way `result_to_score` was — it carries #706's dual-shape tool-call decoding and is locked by the 152-line `tests/unit/grading/test_run_custom_checks_wire.py`, so removing it means deciding where that behaviour lives (or that it goes too). That is why it was kept out of #895.

Issue #673, opened 2026-07-29
Harden custom-checks execution isolation (thread runner is not a sandbox)


CiroGamboa:
## Context

`CheckRunner` (`tolokaforge/core/grading/check_runner.py`) executes author-provided `checks.py` via a `ThreadPoolExecutor` in the *grader's own process*. Once #669 wires this into the runner container, task-author Python runs in-process inside the runner alongside the DB/RAG clients.

Two real gaps this leaves:

1. **Timeout does not kill runaway checks.** `future.result(timeout=...)` returns to the caller on timeout, but the underlying Python thread cannot be force-killed — an infinite loop in `checks.py` leaks a live thread for the runner's lifetime.
2. **No isolation boundary.** A check runs with the runner process's full ambient authority: it can open the runner's internal service sockets (db-service/rag-service via internal DNS), read arbitrary container files, and mutate `sys.modules`/`sys.path` (the loader already manipulates both).

The container-level `no_internet` NetworkPolicy (ADR-0018) blocks *external* egress, and #669's first customer is a trusted, deterministic, network-free pack — so this is acceptable for the initial seam. But for untrusted or third-party check code the in-process thread model is not a sandbox.

## Proposal

Evaluate a real isolation backend for `CheckRunner` (subprocess with a killable process boundary + seccomp/namespace network deny, or equivalent), selected behind the existing executor seam so trusted packs keep the fast in-process path. Out of scope for #669 (finishing the wire); filed so the hardening is tracked rather than assumed.

Relates to #669.

Issue #583, opened 2026-07-22
single-tool adapters: register-for-execution-without-advertising, or an adapter-owned execution/grading seam


azorej:
## Use case

A single-**`bash`**-tool adapter (`bash_frozen_mcp_core`, productionizing POC C2 — see toloka-partners/tolokaforge-tools#42 and the accepted design doc `docs/design-bash-only-adapter.md`) needs the agent to be offered **exactly one tool, `bash`**, while the frozen domain tools still execute with full behavioral + state-hash parity and per-invocation audit.

On tolokaforge 0.9.3 the engine forces a lose-lose choice for such adapters, because *registering a tool so it can run* is inseparable from *advertising it to the agent* and from *how grading finds it*:

1. **The docker conductor advertises the entire registered tool list to the agent.** `core/conductor.py` builds the agent's tool list from the full `RegisterTrial` response and ignores `num_agent_tools`; there is **no per-tool "hidden"/"advertise" flag** on `ToolSchema` (`runner/models.py`). So any domain tool registered so the bridge can call it via `ExecuteTool` is also shown to the model — breaking the single-tool requirement.
2. **Docker-mode grading is runner-owned and needs the tools registered too.** `RunnerRPCTrialGrader` → `GradeTrial` replays golden actions resolved only from `trial_context.agent_tools` (`runner/service.py`); the adapter's own `grade()` is never called. So even hiding tools from the agent wouldn't let grading work without them registered.

## Current workaround (v1, adapter-side)

`bash_frozen_mcp_core` registers **only** `bash` on the runner and runs the frozen domain tools in a per-trial **sidecar** that executes them against the same `db-service` via `SyncDBServiceProxy` (identical to `MCPAsyncToolWrapper`), enforces the `tools.agent.enabled` allow-list, writes the audit log, and serves a `/grade` endpoint that reproduces the runner's hash-grading algorithm using db-service's own `get_stable_hash`. State unforgeability is structural (the bash container sits on an `internal` network and can reach only the sidecar). Verified: scripted golden-replay state hash is byte-identical to `frozen_mcp_core` on `tau_manufacturing/MAN-34` + two `ots_11` tasks.

This works but **duplicates the runner's tool-exec and hash-grading in an adapter sidecar**, and needs an out-of-band `trial_id` (reconstructed from the `tbench_<trial_id>` compose project name via the wrapper-injected `T_BENCH_TASK_DOCKER_CLIENT_CONTAINER_NAME`).

[...]

Issue #400, opened 2026-07-15
feat(examples): multi_service_lot_ops — postgres + one API + real state-mutation grading


CiroGamboa:
Part of the \`Multi-container example depth\` milestone.

## Context

Every current multi-container example is a 2-step lookup (fetch + write); \`state_checks\` inspect the agent's own submission file, never the substrate. This pack introduces the "real state-mutation grading against substrate" pattern that every future non-trivial example needs.

## Scenario

Shop-floor operator asks the agent to perform a state mutation — open a corrective-action record against a lot, allocate lots to a production order, or close a completed order. Internal-user scenario; no CRM/customer layer.

## Stack

- \`postgres:16\`
- One small FastAPI service (bind-mounted \`main.py\`, no build) exposing ~20 CRUD endpoints
- \`tolokaforge-runner:local\` + \`tolokaforge-db-service:local\`

Under 60 lines of compose. All services \`isolation: shared\` — this pack is the "shared multi-container with real state" reference.

## Grading

Three-way combined (currently unexemplified in any pack):
- \`state_checks\` — query postgres via a grading fixture; verify the correct row exists with correct fields after the agent's action.
- \`transcript_rules\` — require the specific tool-call pattern (GET before mutation).
- \`llm_judge\` — small rubric on completion-report quality.

## Ship condition

- \`scripts/with_env.sh uv run tolokaforge run --config examples/native/multi_service_lot_ops/run_config.yaml\` exits 0, produces \`binary_pass: true\`.
- \`state_checks\` component verifies substrate state (not the agent's written file).
- Integration test at \`tests/integration/test_lot_ops_end_to_end.py\` locks the behaviour.

## Priority

Medium. Introductory / cheapest pack in the milestone; foundational template.

## Gap coverage

- State-mutation grading against substrate.
- \`state_checks + transcript_rules + llm_judge\` three-way combined weighting.
- "Small stack" template (postgres + one API).

## Cost

Haiku ~\$0.05-0.20/run; Sonnet ~\$0.30-1.20/run.

CiroGamboa:
Shipped in PR #423 (squash-merged as 1c6f46e33f51c8d7a09f98f58c8bd6244cc302ca into feat/multi-container-example-depth).

- Grading primitive `state_checks.db_probes` (runner + core) with asyncpg driver + Dockerfile guard.
- `examples/native/multi_service_lot_ops/` — 4-service FastAPI + postgres pack with read-only grader-role db_probe.
[...]

PR #704, opened 2026-07-30
feat(grading): finish runner-side custom_checks as a Pattern-A extension

### LINKED ISSUE #217: [refactor] Grading: remove or quarantine the dead in-process grading engine and its dead plumbing
## Summary

A full deep-dive into the grading subsystem (July 2026, part of the Project-layer design work, see #189) found that production grading runs **entirely runner-side** (`conductor.py:641` → `trial_grader.py:103` → `runner/service.py:924 GradeTrial` → `runner/grading.py` + `core/grading/judge.py`). The parallel **in-process grading engine is dead in production**, along with a set of config fields and conversion artifacts that nothing reads at runtime.

## Dead components (no production caller)

- `core/grading/combine.py::GradingEngine` — only entry is `BaseAdapter.grade` (`adapters/base.py:470-511`), which nothing in the pipeline invokes (the only production `.grade()` call is `conductor.py:641` → `trial_grader`).
- `core/grading/state_checks.py` tau-bench paths — `grade_tau_style` (`:414`), `check_hash`/`to_hashable`/`consistent_hash` (`:20-34,241`), self-described "tau-bench compatible".
- `core/evaluators/environment_evaluator.py` — tau2-specific (`env_assertions`, `db_hash_check`; imports `tolokaforge.tasks.{domain}.assertions` at `:76`); reachable only via `GradingEngine`.
- Custom-checks stack — `@check`/`@init` registry (`checks_interface.py:399,415`) + `check_runner.py`; wired only into `GradingEngine`. Runner-side, `custom_checks` is a stub: `service.py:1205` hardcodes `custom_checks=-1.0  # Not implemented yet`, while the field remains schema-visible.
- `core/models.py::GradingConfig`/`StateChecksConfig` (`jsonpaths`/`hash`/`env_assertions`/`db_hash_check` variant) — consumed only by the dead engine; the runner uses the separate `runner/models.py::GradingConfig`.

## Dead config plumbing (populated, never read)

- `StateChecksConfig.expected_hash` (`runner/models.py:283`), YAML `expected_state_hash`, proto `precomputed_expected_hash` (`runner.proto:180`) — populated by `adapters/native.py:587`, consumed nowhere. Production computes the golden hash **live**: `_execute_hash_grading` (`service.py:1476`) resets the trial DB and replays `golden_actions` through the real agent tools (`:1522-1653`).
- `fixtures/golden_actions.json` and `fixtures/unstable_fields.json` — written by `adapters/bundle_writer.py:88-89` at conversion, read by no harness or adapter code at runtime (golden actions actually ride in `grading.yaml` → `TaskDescription`; unstable fields ride in `initial_state.unstable_fields`, registered into the db-service at `service.py:522`).

## Why act

- Contributors keep "fixing" the dead engine (see #83's drift history: core grader patched in PR #66 while production behavior was runner-side all along).
- The dead stack is where the mcp_core/tau coupling lives (#22) — deleting or quarantining it shrinks that problem.
- Two divergent `GradingConfig` models are a standing drift hazard. NOTE the packaging constraint from #83: the runner is standalone (zero `tolokaforge.core` imports), so unification means a shared dependency-free module, not an import.

## Proposed direction

1. Decide per component: delete, or explicitly quarantine as the parts bin for a future pluggable `StateCheckProvider` registry (the natural seam is the hardcoded dispatch at `service.py:1026-1073`).
2. Remove dead `expected_hash`/`precomputed_expected_hash` plumbing (or implement the precomputed path deliberately).
3. Stop writing dead fixture files in `bundle_writer`, or document them as provenance artifacts.
4. Either implement `custom_checks` runner-side (first provider on the registry?) or drop it from the schema until it exists.

Related: #83, #22, #131, #189.

### LINKED ISSUE #406: Runner-side custom checks (checks.py) are unimplemented despite a full public API
## Gap

`tolokaforge/core/grading/checks_interface.py` ships a complete, documented public API for custom Python grading checks (`@init`/`@check`, `CheckContext`, `CustomChecksConfig`), and the in-process `GradingEngine._run_custom_checks` (`tolokaforge/core/grading/combine.py`) implements it. But the **production grading path** — the runner's `GradeTrial` RPC (`tolokaforge/runner/service.py`) — hard-codes:

```python
custom_checks=-1.0,  # Not implemented yet
```

and the runner-side `GradingConfig` (`tolokaforge/runner/models.py`) has no `custom_checks` field at all. Since `tolokaforge run` grades exclusively via `RunnerRPCTrialGrader` → runner RPC (`tolokaforge/core/orchestrator.py`), a `custom_checks` block in a real task's `grading.yaml` is silently ignored end-to-end.

## Impact

A documented author-facing feature does not work in the only production grading path. Authors can write `custom_checks:` in `grading.yaml` and get no error and no effect.

## Options

1. Implement runner-side custom checks: add `custom_checks` to the runner `GradingConfig`, deliver `checks.py` to the runner (via `tool_artifacts` or a new grading-artifact channel), and execute via the existing `CheckRunner` inside `GradeTrial`.
2. If not implementing: make the native adapter reject a `custom_checks` block with a clear "not supported in the runner path" error (fail loud), rather than silently dropping it.

## Context

Discovered while planning #400. That pack does **not** need custom checks — substrate grading there is handled by a new declarative `state_checks.db_probes` primitive. Filing so the half-built seam is tracked and not mistaken for working.

### LINKED ISSUE #669: feat: finish runner-side custom checks as a general adapter/task extension
## Context

tolokaforge ships a documented public API for author-provided Python grading checks (`@init`/`@check` decorators, `CheckContext` carrying the trial transcript + state, `CustomChecksConfig` schema, subprocess sandbox runner), but the runner-side wire has been marked `# Not implemented yet` since the initial open-source release. Post-multi-container, the runner RPC is the only live grading path — so `custom_checks:` blocks in `grading.yaml` are silently dropped end-to-end. Two open issues (#406, #217) have asked to either finish or delete this seam.

The gap it fills is real and widening: benchmark adapters increasingly need deterministic Python scoring that combines transcript context with final state, using arithmetic no declarative primitive expresses. The recent `state_checks.db_probes` shipment (#400) shows the platform preferring typed primitives when the shape is regular; some shapes aren't. This ticket closes the gap by formalizing the existing custom-checks stack per ADR-0011 and finishing the runner-side wire.

## Deliverables

- [ ] ADR filed for the custom-checks extension (Pattern A: Protocol + prod impl + `InMemory*` fixture + canonical contract test)
- [ ] Runner-side wire: `custom_checks` blocks in `grading.yaml` reach a runner-side `CheckRunner`; results flow through the composite grade
- [ ] `checks.py` delivery mechanism to the runner container documented and shipped
- [ ] Fail-loud validation: unsupported `CheckContext.interface_version` rejects at grader startup, not silently at trial end
- [ ] `docs/GRADING.md` documents custom_checks as a first-class dimension alongside state_checks / transcript_rules / llm_judge
- [ ] `docs/custom_checks.md` becomes real: schema, decorator API, sandbox invariants, tier guidance for tests
- [ ] First-customer fixture task pack exercising the seam end-to-end
- [ ] Doctrine documented: checks may not initiate network (see Design Decisions)

## Acceptance Criteria

- A task pack declaring `custom_checks.enabled: true` with a `checks.py` file gets its checks executed by the runner; per-check results appear in the trial's grade with the declared weights applied
- A pack whose `checks.py` declares an unsupported `interface_version` is rejected with a clear message at grader startup (not silently at trial end)
- Canonical contract test locks the `Check` Protocol shape against future drift
- The first-customer fixture pack runs green end-to-end via `tolokaforge run`
- Closes #406 and closes #217

## Design Decisions

- **Pattern A shape (ADR-0011)**: matches every other typed-extension seam in tolokaforge (`TrialGrader` per ADR-0014, `Judge` per ADR-0020). New seam gets a Protocol + prod impl + `InMemory*` fixture + canonical contract test.
- **Runner-side execution, not host-side**: docker-mode grading is the only live path post-multi-container; the transcript already reaches the runner in its native shape. Host-side execution would fork a second grading path that's harder to keep coherent.
- **Doctrine: checks may not initiate network**: preserves the `no_internet` container invariant (ADR-0018, tightened by #581). Enforce in `CheckContext` construction if practical; document as authoring rule otherwise. Use cases needing network (LLM-shaped scoring) go through `LLMJudge`, which has the network-exception path solved.
- **Fail-loud on unsupported `interface_version`**: `1.0` becomes a real contract instead of an implicit one. Future evolution stays disciplined.
- **First customer is a deterministic, network-free case**: sequencing choice — establishing the seam with a task pack that uses transcript + workdir state + arithmetic scoring proves the general shape without co-mingling the network-doctrine question. Adapters needing LLM-shaped scoring already have `LLMJudge`; this seam is explicitly for the deterministic-Python gap.

## Out of Scope

- Non-Python check languages (runner is Python today; language-agnostic extension is ADR-0022 territory)
- Full grader-swap seam for adapters that need to replace the entire grader (tracked in #583)
- Any adapter-specific check implementations (task pack authors write these)
- Retroactive migration of existing packs — none use `custom_checks` today, so this is a green-field seam

Closes #406
Closes #217

CiroGamboa:
## Summary

tolokaforge's grading pipeline has always documented four dimensions — state checks, transcript rules, LLM judge, and custom Python checks. Three were live. The fourth was declared in schema, tested in isolation, and silently dropped end-to-end in the only production grading path since the initial open-source release (verbatim `custom_checks=-1.0  # Not implemented yet` in the runner service). A pack declaring `custom_checks:` got no error and no effect.

This PR closes that gap AND formalises the check executor as a Pattern A extension surface (ADR-0012). The seam is now a place where the platform grows.

## The problem this solves

Three of the grading dimensions are declarative primitives — JSONPath assertions on state, regex/keyword rules on transcripts, LLM-graded rubrics. They cover a wide set of graders. They don't cover *arbitrary Python over both the transcript and the final state* — arithmetic scoring, per-domain deterministic verifiers, tool-call sequence analysis. That shape needs code, not YAML.

Authors hitting that shape had three bad options:

- **Force it into `LLMJudge`.** Wrong tool: non-deterministic, token-expensive, wrong for arithmetic scoring.
- **Monkeypatch `RunnerRPCTrialGrader` at install time.** Forks the grading pipeline; adapter-owned but a one-off in-repo pattern.
- **Ship a new `TrialGrader` class per adapter.** Highest coupling, most sprawl, and the exact opposite of the platform's declarative-primitives direction (see the recent `db_probes` primitive shipment in #400 as the counter-example).

None of those actually filled the gap: a first-class *deterministic-Python check* dimension. `custom_checks:` was designed to be that dimension — it just never got wired to the docker path.

## The solution

Finish the wire, and formalise the executor as a Pattern A seam. The surface stays honest under real conditions:

- The check executes runner-side, where the trial's state and transcript live post-multi-container.
- Delivery reuses the existing `tool_artifacts` transport — no new mechanism.
- Interface-version validation moves to trial startup so unsupported packs reject before burning a trial.
- A composite-grade guard prevents a `custom_checks`-only pack from silently passing when the executor produces no score.
- Concurrent module loads across trials no longer cross-contaminate.
- Executor exceptions route through an audit-trail wire entry, not the whole-trial `success=False` bucket.

A deterministic, network-free reference pack proves the whole pipeline end-to-end on real Docker + real gRPC.

## What this opens up

The value of this PR isn't the wire fix — it's what the seam enables.

**For task-pack authors.** A real, honest extension point for deterministic Python grading. Ship a `checks.py` alongside `task.yaml`; the runner picks it up. Grading logic stays in the task pack — not in adapters, not in framework code. This is AGENTS.md Rule 2 made concrete (harness generic, task-specific logic in task packs).

**For adapters.** The pattern "my benchmark needs per-domain grading logic" no longer requires a new `TrialGrader` class, a monkeypatch of the built-in grader, or an abuse of `LLMJudge`. The adapter contributes *nothing* to the grading pipeline — it ships packs with `checks.py` and the framework does the rest. Coupling drops; adapter cohesion goes up. New benchmarks land as task packs, not as framework code.

**For platform evolution.** The `CheckExecutor` Protocol (ADR-0012) is where new executor variants plug in without touching the pipeline. Concrete candidates already visible:

- **`SubprocessCheckExecutor`** — real sandbox with enforceable timeouts, per-check process isolation, no shared filesystem/network with the runner. Closes the doctrine gap #673 tracks.
- **`ReplayCheckExecutor`** — offline replay against captured grading transcripts for regression testing and audit.
- **`RemoteCheckExecutor`** — dispatch to an isolated service if a benchmark family wants its scoring separated.
- **Language-agnostic executors** — WASM or shell-contract variants that break the "runner-must-be-Python" coupling (ADR-0022 territory).

None of these need to touch `RunnerService`, the combiner, or the trial protocol. They plug in behind the same Protocol boundary.

**For the four-dimension model itself.** `custom_checks` now lives on the correct side of the multi-container split (M8, #511) — the runner is the source of truth for the trial's transcript and state, and this dimension now runs there. When patterns of custom checks stabilise into common shapes, they graduate into typed primitives (the way `db_probes` did in #400). Custom checks *incubate*; primitives *graduate*. The pipeline gets a clean elevator between the two.

**Complementary, not competing, with #583** (adapter-owned execution/grading seam). That's for cases needing full grader replacement — separate concern, larger granularity. This seam sits at check-level, well below.

## What is NOT in scope

- **True per-check sandboxing** (#673). The in-process thread executor is not a sandbox; doctrine is enforced at the runner container's `no_internet` boundary (ADR-0018) for the trusted first-customer scope.
- **Deletion of the in-process `GradingEngine`.** #217 asked "finish or delete"; this PR closes it in the *finish* direction. Dead convenience helpers separated to #674.
- **Non-Python check languages** — ADR-0022 territory.
- **Full grader-swap seam** — #583.

## Verification

- **Canonical**: Protocol contract, combiner routing, silent-pass regression, host-side per-check mapping, RegisterTrial validation, cleanup-on-rejection, reserved-name collision.
- **Unit**: shared context-builder shape (parametrised), concurrent module load, executor-error branches (both `CheckResultSet.error` and executor-raise paths), adapter delivery + config propagation.
- **Integration** (`requires_docker`): full end-to-end grading of the reference pack on real Docker + real gRPC (positive + negative cases).

Full staged plan preserved at `~/.claude/plans/toloka-tolokaforge/issue-669-custom-checks.md`; commits map to stages.

Closes #669
Closes #406
Closes #217

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project are documented in this file.
 
+## Unreleased
+
+### Feat
+
+- **grading**: runner-side `custom_checks` execution — a pack declaring `custom_checks.enabled: true` with a `checks.py` now has its `@init` + `@check` functions executed inside the runner container over the trial's `CheckContext` (initial/final state + transcript + task metadata). Per-check verdicts ride the wire as `CustomCheckResult` entries (`Grade.custom_checks_details`), the aggregate score fills `Grade.components.custom_checks`, and the `combine.weights.custom_checks` weight is applied to the final score alongside `state_checks` / `transcript_rules` / `llm_judge`. An unsupported `interface_version` or a broken `checks.py` module now rejects the trial at `RegisterTrial` (before the agent loop) instead of at trial end. `checks.py` is bundled into `TaskDescription.tool_artifacts` for packs with or without an MCP server. The `CheckExecutor` Protocol seam ([ADR-0012](docs/adr/0012-custom-checks-extension.md)) lifts the executor to ADR-0011 Pattern A with `CheckRunner` as the in-process production impl and `InMemoryCheckExecutor` as the test fixture; no public class was renamed. See [docs/custom_checks.md](docs/custom_checks.md) and [`examples/native/custom_checks/`](examples/native/custom_checks/) — the runnable reference pack that reconciles a customer ledger arithmetically. Closes #669, #406, #217.
+
 ## v0.13.0 (2026-07-30)
 
 ### Feat
--- a/docs/FUTURE_DEVELOPMENT.md
+++ b/docs/FUTURE_DEVELOPMENT.md
@@ -172,7 +172,7 @@ Test each configured provider:
 | Transcript rules | calc_custom_checks | `transcript_checks.required_phrases` pattern matching |
 | State assertions | tool_use_public_example_01 | `state_checks.assertions` JSONPath checks |
 | LLM judge | (create test task) | `judge` config calls LLM and returns structured score |
-| Custom checks | order_modify_with_checks | `custom_checks.script` executes Python grading logic |
+| Custom checks | reconcile_ledger (`examples/native/custom_checks/`) | `custom_checks.enabled: true` + `custom_checks.file` runs the pack's `checks.py`; the aggregate score fills `Grade.components.custom_checks` and per-check `CustomCheckResult`s ride the wire |
 | Combined grading | (create test task) | Multiple grading methods compose correctly |
 
 ### 12.2 Unstable fields verification
--- a/docs/GRADING.md
+++ b/docs/GRADING.md
@@ -1,10 +1,14 @@
 # Grading System
 
-Tolokaforge evaluates agent performance across three dimensions:
+Tolokaforge evaluates agent performance across four dimensions:
 
 1. **State Checks** - Final environment state verification (hash-based or JSONPath)
 2. **Transcript Rules** - Process constraints (required phrases, tool usage, turn limits)
 3. **LLM Judge** - Per-criterion rubric grading by a read-only agentic judge
+4. **Custom Checks** - Author-written Python `@check` functions for the
+   deterministic-Python gap the other three don't express (arithmetic
+   over final state, transcript patterns tied to computed values). See
+   [custom_checks.md](custom_checks.md).
 
 Scores are weighted and combined into a final score. See [REFERENCE.md](REFERENCE.md) for `grading.yaml` schema.
 
@@ -424,6 +428,41 @@ scored as it was). See [OUTPUT_FORMAT.md](OUTPUT_FORMAT.md).
 
 ---
 
+## Custom Checks
+
+The `custom_checks` component runs author-written Python `@check`
+functions from a pack's `checks.py`. It's the deterministic-Python gap
+the other three components don't express: arithmetic over final DB
+rows, invariants that span multiple tables, transcript patterns tied to
+computed values. Each `@check` returns `CheckPassed` / `CheckFailed` /
+`CheckSkipped`; per-check results ride the wire as `CustomCheckResult`
+entries and the aggregate `CheckResultSet.aggregate_score` fills the
+`custom_checks` component.
+
+```yaml
+custom_checks:
+  enabled: true
+  file: "checks.py"
+  interface_version: "1.0"
+  timeout_seconds: 30
+  weight: 1.0
+  fail_on_error: true
+```
+
+The full authoring API (`@init`, `@check`, `CheckContext`), the network
+doctrine (checks may not initiate network — the runner container's
+`no_internet` policy enforces at the container boundary; #673 tracks
+per-check sandboxing), and the delivery mechanics (`checks.py` bundled
+into `TaskDescription.tool_artifacts`) live in
+[custom_checks.md](custom_checks.md). The seam itself is
+[ADR-0012](adr/0012-custom-checks-extension.md).
+[`examples/native/custom_checks/`](../examples/native/custom_checks/)
+is the runnable reference — a ledger-reconciliation task that verifies
+`balance == opening + sum(credits) - sum(debits)` and combines that
+with `state_checks` under `combine.weights.custom_checks`.
+
+---
+
 ## pass@k Metrics
 
 Estimates probability that at least 1 of k attempts succeeds.
@@ -535,8 +574,11 @@ probe passes only for the one the after-hours policy permits.
 Final score formula:
 
 ```
-final_score = (state_score * W_state + transcript_score * W_transcript + judge_score * W_judge)
-              / (W_state + W_transcript + W_judge)
+final_score = (state_score       * W_state
+             + transcript_score  * W_transcript
+             + judge_score       * W_judge
+             + custom_score      * W_custom)
+              / (W_state + W_transcript + W_judge + W_custom)
 
 binary_pass = (final_score >= pass_threshold) AND (no required rubric criterion gated)
 ```
@@ -544,7 +586,9 @@ binary_pass = (final_score >= pass_threshold) AND (no required rubric criterion 
 A component that was not evaluated is **excluded** from both the numerator and
 the denominator — this includes an `llm_judge` component whose judge ERRORED
 (see [LLM Judge](#llm-judge-rubric-grading)): a broken judge is never folded in
-as a `0.0`.
+as a `0.0`. A `custom_checks`-only pack whose score comes back absent still
+fails loud (empty active set with a configured `custom_checks` weight ⇒
+`(0.0, False)`, not a silent `(1.0, True)`).
 
 ### Weighting Strategies
 
@@ -560,6 +604,13 @@ combine:
 combine:
   weights: { state_checks: 0.6, transcript_rules: 0.3, llm_judge: 0.1 }
   pass_threshold: 0.75
+```
+
+**Outcome + deterministic-Python check:**
+```yaml
+combine:
+  weights: { state_checks: 0.4, custom_checks: 0.6 }
+  pass_threshold: 0.8
 ```
 
 ### Inheriting `combine` from the project
@@ -641,5 +692,6 @@ state_checks:
 ## See Also
 
 - [REFERENCE.md](REFERENCE.md) - Configuration schemas
-- [CUSTOM_CHECKS.md](CUSTOM_CHECKS.md) - Custom Python validation
+- [custom_checks.md](custom_checks.md) - Custom Python validation
+- [ADR-0012](adr/0012-custom-checks-extension.md) - `CheckExecutor` Protocol seam
 - [TASKS.md](TASKS.md) - Task authoring guide with difficulty design patterns
--- a/docs/GRADING_VERIFICATION.md
+++ b/docs/GRADING_VERIFICATION.md
@@ -244,13 +244,7 @@ Output files are written by [`tolokaforge/core/output_writer.py`](../tolokaforge
 
 ---
 
-## C) What's Missing / Placeholder
-
-### Custom Checks placeholder
-
-```python
-custom_checks=-1.0,  # Not implemented yet
-```
+## C) Optional Components
 
 ### Transcript Rules
 
@@ -266,20 +260,33 @@ custom_checks=-1.0,  # Not implemented yet
 - `max_turns` - Verify conversation under turn limit
 
 **Current behavior:**
-- `transcript_rules` field in `GradeComponents` is usually `null`
-- Rules are evaluated if configured in `grading.yaml`
+- `transcript_rules` field in `GradeComponents` is `null` when the pack ships
+  no `transcript_rules` block.
+- Rules are evaluated when configured in `grading.yaml`.
 
 ### Custom Checks
 
-**Status: Implemented but not used in Docker architecture**
+**Status: Live in the runner grading path.**
 
 **Implementation at:**
-- [`tolokaforge/core/grading/check_runner.py`](../tolokaforge/core/grading/check_runner.py) - `CheckRunner`, `run_custom_checks()`
-- [`tolokaforge/core/grading/checks_interface.py`](../tolokaforge/core/grading/checks_interface.py) - Interface definitions
+- [`tolokaforge/core/grading/check_runner.py`](../tolokaforge/core/grading/check_runner.py) — `CheckRunner` (in-process production impl of the `CheckExecutor` Protocol).
+- [`tolokaforge/core/grading/checks_interface.py`](../tolokaforge/core/grading/checks_interface.py) — authoring API (`@init`, `@check`, `CheckContext`, `CustomChecksConfig`).
+- [`tolokaforge/core/grading/checks_helpers.py`](../tolokaforge/core/grading/checks_helpers.py) — `build_check_context` shared by both grading paths.
+- [`tolokaforge/runner/service.py`](../tolokaforge/runner/service.py) — `RunnerService.GradeTrial` runs the executor when `grading.custom_checks.enabled`.
 
 **Current behavior:**
-- `custom_checks` field in `GradeComponents` is always `null`
-- Runner service sets `custom_checks=-1.0` (not implemented)
+- `RegisterTrial` validates `custom_checks.file` at startup — an
+  unsupported `interface_version` or a broken module rejects the trial
+  before the agent loop runs.
+- `GradeTrial` runs each `@check`, fills `GradeComponents.custom_checks`
+  with the aggregate score, and rides per-check `CustomCheckResult`s on
+  `Grade.custom_checks` (the host maps them into
+  `Grade.custom_checks_details`).
+
+See [GRADING.md § Custom Checks](GRADING.md#custom-checks) for the mechanism,
+[custom_checks.md](custom_checks.md) for the authoring API + network doctrine,
+and [ADR-0012](adr/0012-custom-checks-extension.md) for the `CheckExecutor`
+seam.
 
 ---
 
--- a/docs/GRADING_VERIFICATION.md
+++ b/docs/GRADING_VERIFICATION.md
@@ -10,7 +10,7 @@ The grading system has been verified to work correctly for:
 - ✅ Results saved correctly to grade.yaml and trajectory.yaml
 - ⚠️ LLM judge (placeholder - not implemented)
 - ⚠️ Transcript rules (implemented but rarely used)
-- ⚠️ Custom checks (implemented but rarely used)
+- ✅ Custom checks (live in the runner grading path — see §C)
 
 ---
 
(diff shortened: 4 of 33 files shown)


azorej:
## Integration note from milestone 28 (#684) — not a review of the design

Context: I am driving milestone 28 on the integration branch `feat/deterministic-trace-checks`. #675 merged there (PR #699) and adds two things to the grading path that **do not exist on `main`**, so this PR could not have known about them. Flagging what happens when they meet, plus one detail where I think your fix stops one step short.

The design argument here is sound and I am not contesting it — "custom checks incubate, primitives graduate" is a good framing, and finishing a dimension that was `custom_checks=-1.0  # Not implemented yet` since the initial release is plainly the right call.

### Three things this PR will trip once the two branches meet

#675 added a **grading-key manifest** (`tolokaforge/core/grading/key_manifest.py`) that enumerates every author-facing `grading.yaml` key with which substrate evaluates it, and a **grade-time accounted-keys ledger** (`tolokaforge/runner/grading_ledger.py`). Both are guarded by canonical locks.

1. **A canonical lock will go red on the new runner field.** Lock 1 walks `model_fields` on both substrates' grading configs and asserts every leaf is claimed by exactly one manifest entry. `custom_checks: dict[str, Any] | None` on the runner `GradingConfig` would be an unclaimed leaf, so the test fails naming the field and the model. That is the guard doing its job — it exists precisely so a grading key cannot land on one substrate without being declared.

2. **The manifest entry for `custom_checks` becomes false in all three clauses.** It currently reads `coverage=CORE_ONLY`, `runner_field=None`, with the reason *"the runner reports `custom_checks=-1.0`; resurrecting it is a sandboxing project, out of scope per epic §7"*. After this PR the runner does evaluate it. The entry needs new coverage, a `runner_field`, a named `runner_evaluator`, removal from the drift-exemption set, and its tracking issue cleared.

3. **A second lock will go red for a missing ledger recording site.** `custom_checks` is `kind=SCORED_CHECK`. Once it has a `runner_field` it enters the ledger, and lock 5 asserts every ledger key with a `runner_field` is claimed by an evaluate-or-skip site in `_grade_trial_async`. Without one, **every task populating `custom_checks` would get `success=False` from `GradeTrial`** rather than a grade.

**None of this is work for you.** We own the manifest and the ledger, they are not on your base, and the reconciliation is one manifest entry plus one recording site. I will absorb it on our side once this lands on `main` — I am flagging it so the red locks are expected rather than alarming if you ever rebase onto our branch, and so nobody "fixes" a lock by loosening it.

### One place I think the fix stops short

Your composite-grade guard is the same defect we independently recorded against #678, and I think your diagnosis is exactly right — the empty-active-components branch returning a silent `(1.0, True)` is an AGENTS.md Rule 1 violation. But the fix adds a fourth hand-written clause:

```python
if "custom_checks" in weights and grading_config.get("custom_checks") is not None:
    actually_configured.add("custom_checks")
```

That closes it for `custom_checks` and leaves the shape that made it possible. The branch enumerates component names by hand, so **a fifth component is invisible to it again** — and milestone 28 is adding exactly that (`trace_checks`, #678). We would then add a fifth `if`, and the next person a sixth.

Replacing the chain with a lookup over the known component names would retire the class instead of adding to it. Cheap here, cheaper than the second rediscovery. Entirely your call on whether it belongs in this PR or a follow-up — I have noted it on #678 either way so we do not both add a clause and neither of us fixes the shape.

### Two other overlaps, both minor

- **`docs/GRADING.md`** — #675 added a "Substrate Parity" section describing the manifest, its three axes, and its enumeration ceiling. Your +57 lines to the same file will conflict textually; nothing semantic.
- **`docs/GRADING_VERIFICATION.md`** — #675's review corrected its "Supported rule types" list, and #698 tracks the larger question of that file duplicating what the manifest now answers. Your edits there may overlap.

Happy to do the reconciliation from our side in either merge order — just let me know if you would rather land on `main` first (which seems simplest, since you target `main` and close three issues) or want our branch to go first.

azorej:
## Smoke failure diagnosed — one missing snapshot regeneration

`test-smoke`: **1 failed, 4463 passed** — `tests/canonical/test_terminal_bench_adapter_canon.py::TestTerminalBenchAdapterCanon::test_task_description`.

The actual and golden dicts differ in **exactly one key**. The adapter now produces:

```
"grading": { ..., "custom_checks": None }
```

and the committed golden has no `custom_checks` at all. Confirmed against `main`: that snapshot's `grading` block carries seven keys (`combine_method`, `grading_method`, `llm_judge`, `pass_threshold`, `state_checks`, `transcript_rules`, `weights`) and `custom_checks` is not among them.

**Cause:** adding `custom_checks: dict[str, Any] | None = None` to the runner `GradingConfig` changes `TaskDescription.grading`, so every serialised task description gains the key — including the terminal-bench golden. The PR touches no snapshot files.

**Fix:** regenerate `tests/canonical/snapshots/tbench_echo_hello/task_description.json`. The diff should be one added line, `"custom_checks": null`. Nothing semantic, and no other snapshot appears affected — the native-adapter goldens already carried a `custom_checks` key on the core config.

Easy one to miss: that snapshot belongs to the **external** terminal-bench adapter, so it does not come to mind while editing grading code. Worth a `rg -l custom_checks tests/canonical/snapshots/` before the next push in case a second golden is affected in your working tree.

---

One observation, offered as an argument for the guard rail rather than a criticism of this PR. Your CI caught this as *"a snapshot moved."* The parity manifest on our milestone branch would have caught the same change at its source as *"a field was added to a grading config model without declaring which substrate evaluates it"* — which names the obligation rather than a downstream symptom, and points at the two follow-on requirements (the manifest entry and the ledger recording site) instead of leaving them to be discovered later.

That is exactly the reconciliation I flagged earlier: the same field that moved this snapshot is the one that will trip lock 1 when the branches meet. Still ours to do, not yours.

azorej:
Pushed the fix to this branch as `f9f9812` — one regenerated golden, one added line. Shout if you would rather I had left it to you.

**What I did**, in case you want to verify rather than take it:

1. Reproduced on your branch: `1 failed, 11 passed`, with `grading` the only differing item — the diagnosis above confirmed rather than inferred.
2. Regenerated via the repo mechanism (`pytest tests/canonical/test_terminal_bench_adapter_canon.py --update-canon`), not by hand-editing the JSON.
3. Diff is exactly `+ "custom_checks": null,` — `1 file changed, 1 insertion(+)`. Nothing else moved.
4. Full canonical tier on your branch: **740 passed, 0 failed.**

**It was the only stale golden, and I checked rather than assumed.** I walked every `task_description.json` snapshot on `main`: this is the only one whose `grading` block lacked the key. The reason is worth knowing for next time — `task_description.json` serialises the **runner** grading config and only the terminal-bench adapter has such a snapshot, while the native-adapter goldens snapshot `grading_config.json` (the core config), which already carried `custom_checks`. So there is no second cycle waiting.

Two things I noticed while in your CI that are not yours to fix:

- **`hygiene-review` passed here** (2m52s, real analysis). It failed on four consecutive PRs across two branches when I filed #701, so either that is fixed or it is intermittent — I will update #701 with this counter-evidence rather than leave it describing a permanent outage.
- **`test-gate` skipped again**, as in all 40+ runs I sampled — #700. That is the only PR lane that builds images and runs the integration tier, so **your `requires_docker` integration tests are not running in CI either.** Given this PR ships a reference pack whose whole point is proving the pipeline on real Docker + real gRPC, that gap seems worth knowing about before merge: the tier you wrote that evidence for is not being exercised by the merge gate.

azorej (approved):

azorej:
## Reconciled with the milestone-28 parity guard rail — plus one blocker in this PR's own reference pack

Reconciliation landed on `feat/676-event-record` (commit `c3d8bce`) ahead of the
integration branch for milestone #28. Recording what it found, since two items are about
this PR rather than about the guard rail.

### `custom_checks` now claims both substrates

The parity manifest declares it `coverage=BOTH_SCORE_PARITY`,
`enforcement=DIFFERENTIAL_CANONICAL`, `runner_field="GradingConfig.custom_checks"`,
`runner_evaluator="…RunnerServiceImpl._grade_custom_checks"`, with no exemption and no
`reason` — it needs neither.

Decided by measurement, not by reading the diff: this PR's `combine.py` change is an
**extraction, not a removal**, so core still evaluates. Both substrates route through
`CheckRunner.run` over a `CheckContext` from the same `build_check_context` and return
`result.aggregate_score`. Probed on one pack: satisfying **1.0 / 1.0**, violating
**0.0 / 0.0**. Because the executor is in-process, canonical-tier enforcement was
available, so no integration citation was needed. `custom_checks` left the drift
exemptions and the non-differential set was not widened.

### Blocker: the positive case of this PR's reference pack cannot pass

`tests/integration/test_custom_checks_e2e.py` needed a mechanical fix for an unrelated
`call_id` kwarg. After that it fails on a **pre-existing defect this PR inherits rather
than introduces**, which I deliberately did not paper over:

The builtin `db_update` posts to `{db_url}/update` — the **non-trial-scoped legacy route**
— which services the shared `DEFAULT_TRIAL_ID = "__default__"` trial. That trial is never
seeded from the pack's `initial_state`, so a JSON-Patch `replace` on a path that does not
exist returns 400. The agent writes to `__default__` while grading reads
`/trials/{trial_id}/state`; the two never meet.

Filed as **#710** with the full route table. Scope is wider than this PR — all four
builtin JSON-DB tools bypass trial isolation, so concurrent trials share mutable state —
and it also covers the blanket `except Exception → HTTPException(400)` in that handler,
which is what makes a setup failure present as an agent-visible validation error. MCP task
tools are unaffected; they bind `trial_id` through `DBServiceProxy` correctly.

The practical consequence for this PR: **`examples/native/custom_checks` demonstrates the
config surface but cannot be run end-to-end as a positive case until #710 lands.** Worth
knowing before it is pointed at as the reference.

I tested and rejected a JSONPath-syntax hypothesis for the 400 and reverted that
speculation rather than ship an unverified guess.

### Also fixed here

Four canonical tests this PR added were routed through a shared
`tests/utils/runner_requests.register_request` helper, and the four duplicated
`custom_checks` enabled-predicates were consolidated into a `custom_checks_enabled()`
TypeGuard in `core/grading/checks_helpers.py`. Push back if you would rather have either
split out.

Tree state after reconciliation: canonical **765 passed / 0 failed**, unit **3809 passed**
with only the pre-existing `test_persistent_shell_local.py` set red, lint and formatters
clean, no snapshot moved.

PR #718, opened 2026-07-31
feat(llm): route LLM calls through a gateway (LiteLLM proxy), env-configured

balazs-toloka:
## Summary

Adds optional support for routing LLM calls through a **gateway** instead of
calling providers directly. A [LiteLLM proxy](https://docs.litellm.ai/docs/simple_proxy)
is the reference target; any gateway presenting the same surface works. Such a
gateway holds the upstream credentials, enforces per-team budgets, and
attributes spend.

Setting `LLM_PROXY_BASE_URL` enables it. Model names, run configs, presets,
certificates, and pricing are untouched. With the variable unset — the default —
behaviour is byte-identical to today.

Nothing in the engine knows about any particular gateway product or
organisation. Deployment specifics, including whatever attribution headers a
given gateway demands, are supplied entirely as configuration.

### What "the same surface" means

Deliberately **not** "the gateway is OpenAI-compatible" — that overclaims. This
layer only overrides the base URL; **litellm decides the path, the auth header,
and the body shape, per provider.** So the real contract is: for each routed
provider, the gateway must serve the route that litellm's transport for that
provider targets. That is why routing is an allow-list rather than a blanket
redirect, and why the allow-list is a claim about the *installed litellm* — which
is pinned by a canonical test rather than left to drift on a dependency bump.

## Environment contract

| Variable | Meaning |
|---|---|
| `LLM_PROXY_BASE_URL` | Gateway base URL. **Setting this enables the transport**; everything else is optional. |
| `LLM_PROXY_API_KEY` | Credential presented to the gateway. Omit only for gateways that authenticate by network position — litellm then falls through to its provider-env lookup and forwards the *provider's* key to the gateway host instead. |
| `LLM_PROXY_HEADERS` | JSON object of static headers added to every request. Wins over the engine's own provider headers on a name collision. |
| `LLM_PROXY_REQUEST_ID_HEADER` | Header *name* that receives a fresh UUID4 per request. A static env var cannot express "new value per call". |
| `LLM_PROXY_PROVIDERS` | Provider allow-list, replacing the default. See the routing table below before widening it. |

All five resolve through `SecretManager`, so `.env`, the process environment,
and the runner container's `TOLOKAFORGE_SECRETS_JSON` behave identically.

## Which providers can be routed

**Setting `api_base` does not make litellm speak OpenAI to that URL — it makes
litellm speak that provider's native protocol to that URL.** Captured against
litellm 1.87.0:

| provider | request litellm sends to the gateway |
|---|---|
| `openrouter/…` | `POST {base}/chat/completions`, bearer auth |
| `openai/…` | `POST {base}/chat/completions`, bearer auth |
| `anthropic/…` | `POST {base}/v1/messages`, `x-api-key` |
| `gemini/…` | `POST {base}/models/<m>:generateContent`, `x-goog-api-key` |

Only the first two are an OpenAI-envelope request, so routing is an **allow-list**
defaulting to `{openrouter, openai}`, pinned against the installed litellm by
`tests/canonical/test_llm_gateway_envelope_contract.py`. Other providers can be
named explicitly, meaning "my gateway also serves that provider's native route" —
true of a LiteLLM proxy's `/v1/messages` passthrough, false of a plain
OpenAI-compatible gateway.

Widening the default to *every* provider would require forcing the OpenAI
envelope (`custom_llm_provider="openai"` on routed calls, as the Nova path
already does) rather than only swapping the base URL. That collides with the
`custom_llm_provider` the OpenRouter branch sets and is more than a transport
swap, so it is deliberately left out of this PR.

`mock` and `nova` are in `UNROUTABLE_PROVIDERS` and rejected even when named
explicitly. `mock` never reaches the wire. `nova` depends on
`_call_with_key_rotation` rewriting its bare model name into `openai/<name>`
next to its own hardcoded base URL; a gateway replaces the base URL but not the
rewrite, so litellm would raise `BadRequestError` before sending anything.

## Design: transport swap only

`_build_kwargs` sets `api_base`, `api_key`, and `extra_headers`. The litellm
model string keeps its `<provider>/<name>` shape. That invariant is deliberate —
the obvious alternative (a new `provider: litellm_proxy` with a re-prefixed
model name) breaks two subsystems **silently**:

1. **Presets.** `match:` globs resolve off `ModelConfig.name` and the
   `providers:` overlay off `ModelConfig.provider`. Renaming the provider drops
   the `providers.openrouter` overlay, so `reasoning_via_extra_body` flips to
   `False` and the reasoning wire format changes — while the reported
   `effective_preset` stays the same. Exact-glob presets stop matching too,
   re-opening the documented `recursive_ref` and `dict_map` regressions.
2. **Pricing.** `normalize_model_name` strips exactly one leading
   `openrouter/`, then returns any remaining slash-bearing name verbatim. A
   second prefix guarantees a pricing-table miss, degrading `cost_source` to
   `"unknown"` and tripping `Capability.COST_USD_POPULATED`, which is
   `required` on every certificate.

Because the provider string is untouched, provider-specific request shaping
still applies on top of the gateway: OpenRouter's `HTTP-Referer` / `X-Title`
headers and its `extra_body.provider` upstream pinning both survive.

## Key rotation

Rotation stays bound to the provider key chain, and whether it applies depends
on one thing: is a gateway key pinned?

- **`LLM_PROXY_API_KEY` set** — rotation is skipped. `_rotate_key` republishes
  `OPENROUTER_API_KEY` into the environment, but the pinned `api_key` kwarg wins
  in litellm, so rotating would resend byte-identical requests and then report
  an exhausted key chain that was never in play. A gateway quota rejection
  raises an error naming the gateway URL instead.
- **`LLM_PROXY_API_KEY` unset** — rotation works and is left alone, because
  litellm reads the provider env var that `_rotate_key` rewrites. Suppressing it
  would abort a trial with unused keys still in the chain.

The guard mirrors exactly the condition under which `_build_kwargs` pins the
key, so the two cannot drift.

## Fail-fast

A malformed value raises `ProxyConfigError` at the first `LLMClient`
construction rather than running a whole evaluation with unattributed spend.
Setting any companion `LLM_PROXY_*` variable while `LLM_PROXY_BASE_URL` is empty
also raises, so a typo in the base-URL name cannot silently fall back to direct
provider access. A configured-but-out-of-scope provider logs a warning rather
than dropping to the direct path silently.

## Testing

`tests/unit/llm/test_llm_proxy.py` — 42 tests covering env-contract resolution,
validation failures, provider scoping, per-request header freshness, kwargs
application, the rotation split, and the non-proxy path.

The load-bearing invariants each have an explicit regression test, since those
are what a future refactor would break: the model string staying intact,
OpenRouter header coexistence, `extra_body.provider` preservation, and
`_configure_openrouter_base_url` / `_configure_nova_base_url` still running when
no gateway is configured (previously uncovered by any test).

`tests/canonical/test_llm_gateway_envelope_contract.py` — 5 tests pinning the
allow-list against the installed litellm, so a dependency bump that changes a
provider's transport fails there instead of posting to a route the gateway does
not serve. Verified to have teeth: adding `anthropic` to the allow-list fails
two of them.

```
tests/unit/llm/test_llm_proxy.py                       42 passed
tests/canonical/test_llm_gateway_envelope_contract.py   5 passed
tests/unit/llm/ + test_llm_providers + test_model_client + secrets   716 passed
ruff check / ruff format --check                       clean
mypy tolokaforge/core/llm/proxy.py                     clean
```

Full `tests/unit` + `tests/canonical` on the author's machine:

```
baseline (main):   43 failed, 4210 passed, 3 errors
this branch:       43 failed, 4257 passed, 3 errors
```

The 43 failures and 3 errors are pre-existing in that working copy and
unrelated (an unsynced venv leaves plugin entry points unregistered and
`click_repl` missing). Identical before and after; the delta is exactly the 47
new tests.

## Review notes

This branch went through two rounds of adversarial review, including wire-level
capture against a local HTTP server and mutation testing of the new tests (each
fix reverted in source to confirm the corresponding test fails). Two findings
were accepted as out of scope: proxy config is validated per-`LLMClient` rather
than at CLI startup (a startup hook belongs in a change that touches the CLI),
and the gateway base URL is logged at INFO (operator-supplied configuration,
comparable to what `_load_api_keys` already logs).

## Docs

- `docs/LLM_LAYER.md` — new `proxy` section: the surface contract, env table, the
  routing table above, the transport-swap rationale, and the rotation split.
  Added to the module map.
- `docs/CONFIG.md` — pointer from § Environment Variables.
- `.env.example` — commented block for the five variables.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,7 +23,7 @@ OPENROUTER_API_KEY=your-openrouter-api-key-here
 # VERTEXAI_PROJECT=your-gcp-project-id
 # VERTEXAI_LOCATION=us-central1
 
-# OpenAI-compatible gateway (LiteLLM proxy, Portkey, internal API management).
+# LLM gateway (a LiteLLM proxy or a gateway presenting the same surface).
 # Set the base URL to route calls through the gateway instead of calling
 # providers directly; model names and configs stay unchanged.
 # See docs/LLM_LAYER.md § proxy.
--- a/.env.example
+++ b/.env.example
@@ -22,3 +22,18 @@ OPENROUTER_API_KEY=your-openrouter-api-key-here
 # GOOGLE_APPLICATION_CREDENTIALS=/path/to/service-account.json
 # VERTEXAI_PROJECT=your-gcp-project-id
 # VERTEXAI_LOCATION=us-central1
+
+# OpenAI-compatible gateway (LiteLLM proxy, Portkey, internal API management).
+# Set the base URL to route calls through the gateway instead of calling
+# providers directly; model names and configs stay unchanged.
+# See docs/LLM_LAYER.md § proxy.
+# LLM_PROXY_BASE_URL=https://gateway.example.com
+# LLM_PROXY_API_KEY=your-gateway-key-here
+# Static attribution headers your gateway requires, as a JSON object:
+# LLM_PROXY_HEADERS={"X-Team-Id":"research","X-Cost-Center":"42"}
+# Header name to fill with a fresh UUID per request:
+# LLM_PROXY_REQUEST_ID_HEADER=X-Request-Id
+# Which providers to route. Default is openrouter,openai — the only ones whose
+# litellm transport sends an OpenAI-envelope request. Widen this only if your
+# gateway also serves the other provider's native route (see docs).
+# LLM_PROXY_PROVIDERS=openrouter,openai
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -391,10 +391,10 @@ Common keys:
 - `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`
 - `OLLAMA_API_BASE`
 
-Routing every call through an OpenAI-compatible gateway instead of calling
-providers directly (`LLM_PROXY_BASE_URL`, `LLM_PROXY_API_KEY`,
+Routing calls through an LLM gateway (a LiteLLM proxy or equivalent) instead of
+calling providers directly (`LLM_PROXY_BASE_URL`, `LLM_PROXY_API_KEY`,
 `LLM_PROXY_HEADERS`, `LLM_PROXY_REQUEST_ID_HEADER`, `LLM_PROXY_PROVIDERS`) is
-documented in [`docs/LLM_LAYER.md` § proxy](LLM_LAYER.md#proxy--routing-every-call-through-an-openai-compatible-gateway).
+documented in [`docs/LLM_LAYER.md` § proxy](LLM_LAYER.md#proxy--routing-calls-through-an-llm-gateway).
 
 ## Output Structure
 
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -391,6 +391,11 @@ Common keys:
 - `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`
 - `OLLAMA_API_BASE`
 
+Routing every call through an OpenAI-compatible gateway instead of calling
+providers directly (`LLM_PROXY_BASE_URL`, `LLM_PROXY_API_KEY`,
+`LLM_PROXY_HEADERS`, `LLM_PROXY_REQUEST_ID_HEADER`, `LLM_PROXY_PROVIDERS`) is
+documented in [`docs/LLM_LAYER.md` § proxy](LLM_LAYER.md#proxy--routing-every-call-through-an-openai-compatible-gateway).
+
 ## Output Structure
 
 ```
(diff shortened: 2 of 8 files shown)


CiroGamboa (approved):
Overall LGTM — the canonical envelope-contract test pinning the litellm assumption is exceptional, and the fail-fast on orphan companion vars closes the typo-in-base-URL failure mode cleanly. Two small suggestions, both non-blocking:

**1. `ProxyConfig.headers` is mutable despite `frozen=True`.** The dataclass freezes the field binding, not the dict contents, so external code can `proxy.headers["X"] = "y"` and pollute every future request through that instance. `request_headers()` already copies before returning, so the practical blast radius is small — but the freeze reads as a promise the type doesn't actually keep. Cheapest fix is `headers: Mapping[str, str]` with `types.MappingProxyType` wrapping the parsed dict in the resolver, or store as a `tuple[tuple[str, str], ...]` and materialize per call.

**2. `LLM_PROXY_BASE_URL` isn't scheme-validated.** A typo like `LLM_PROXY_BASE_URL=gateway.example.com` (no scheme) passes `resolve_proxy_config` and fails later at the first request. The module already goes out of its way to fail-fast on every other malformation (orphan companions, non-object headers, unroutable providers) — a quick `urlparse(base_url).scheme in ("http", "https")` check in `resolve_proxy_config` would keep this last case consistent with the rest of the contract.

Neither blocks merge.

Issue #721, opened 2026-07-31
Judge replay silently excludes grade-less bundles, which #676 made routine


azorej:
## Summary

Judge replay discovers trial bundles by rglobbing for `grade.yaml` and requiring both `grade.yaml`
and `task.yaml` (`tolokaforge/core/grading/replay.py:628`, `:663-685`). #676 made **grade-less
bundles routine** — an infrastructure abort now produces no grade at all, by design — so an aborted
trial is not yielded by discovery **at all**, silently.

## Why this is a real gap rather than a doc nit

`docs/JUDGE_REPLAY.md:71` promises that a bundle whose `grade.yaml` is missing is "reported as a
**named per-trial failure** — the batch continues, and no eligible trial is ever silently skipped."

Before #676 that branch was effectively vacuous: every trial directory had a `grade.yaml`, including
the provision-failure path. #676 made the missing-`grade.yaml` case common, and discovery filters on
exactly that file — so **the named-failure branch is now dead for its most common cause**, while the
doc still advertises it.

Under the letter of the promise, excluding an aborted trial is defensible: eligibility is defined by
`judge_status` in `grade.yaml`, and a trial that never had a judge stage was never eligible. The harm
is not a broken guarantee, it is lost legibility:

- `rejudge --dry-run` reports an eligible count that **silently shrinks on a throttled run**
- an operator comparing two replay batches over the same suite cannot distinguish *"fewer trials
  were judge-eligible"* from *"the provider throttled us"*

That is the same defect shape as the metrics half of #676 — a number that moves for a reason nothing
in the output names — one layer over, in the replay tooling.

## Recommended fix

Discover on **`task.yaml`** (always written) rather than on `grade.yaml`, and classify a grade-less
bundle as a distinct **non-failure** status — e.g. `SKIPPED_NO_GRADE`, reason *"the trial produced no
grade (infrastructure abort); nothing to re-judge"* — counted in the report alongside the replayed
and failed counts.

This keeps `FAILED` meaning *"should have replayed and could not"*, which is what makes it
actionable, while restoring the spirit of the line-71 promise: nothing disappears without being
named.

It changes `discover_trial_bundles`' contract and the report shape, which is why #676 documented the
current behaviour (`docs/JUDGE_REPLAY.md:39`, `:71`) instead of changing it inside a review-fix round.

## Deliverables

- [ ] A grade-less trial directory is discovered and reported under a named non-failure status
[...]

Issue #720, opened 2026-07-31
An ungradeable trial vanishes from total_trials and writes no bundle


azorej:
## Summary

When `GradeTrial` refuses to grade a trial, the attempt **disappears from `total_trials`** and no
trial bundle is written. The failure is loud in the log, the retry path and `run_state.json` — but
the aggregate JSONs silently under-count, with nothing in them saying so.

Found completing review round 1 of #676, which fixed the *worse* half of this: the host used to
convert a refusal into a fabricated `Grade(score=0.0)` that landed in `success_rate`, `avg_score`
and `pass@k` as an agent failure. `RunnerRPCTrialGrader.grade` now raises `GradingFailedError`
instead (`tolokaforge/core/trial_grader.py:205`). That removed the fabricated verdict and exposed
this second, quieter problem underneath.

## Mechanism, verified

`Conductor.run` does not catch `GradingFailedError`, so:

1. `_write_artifacts` never runs — **no trial bundle** (`trajectory.yaml`, `metrics.yaml`).
2. `trial_result = future.result()` raises **before** `self.results.append(trajectory)` — in both
   worker loops, `tolokaforge/core/orchestrator.py:1734-1736` and `:2092-2094`. Confirmed by
   reading both call sites: the `append` is the statement after the `future.result()`.
3. Because nothing reaches `self.results`, the trajectory reaches **no metric at all**.
4. The orchestrator marks the attempt retryable, so it retries; a deterministic cause (a duplicate
   `call_id`) fails every retry, then `run_state.mark_failed` + a `trial_failed` event.

## Measured

One task, 2 trials: one `agent_done` scoring `1.0`, one whose grading fails.

| | today (raise) | if it returned `None` | the deleted `0.0` |
|---|---|---|---|
| `total_trials` | **1** | 2 | 2 |
| `measured_trials` | 1 | 2 | 2 |
| `scored_trials` | 1 | 1 | 2 |
| `harness_errors` | 0 | 0 | 0 |
| `success_rate` | 1.0 | 0.5 | 0.5 |
| `avg_score` | 1.0 | 1.0 | 0.5 |
| `pass@1` | 1.0 | 0.5 | 0.5 |

## Why it matters

#676's plan deliberately kept `total_trials` meaning **"every attempt"** — it is the one number
that was supposed to survive the whole infrastructure-abort change unaltered, so a reader can
always recover "how many times did we try". Dropping an attempt breaks that contract silently, and
[...]

Issue #717, opened 2026-07-31
A failed call's recorded result text differs between substrates, so a result matcher is not portable


azorej:
## Summary

The two grading substrates record **different text** for the same failed tool call, so a
`result` predicate written against one substrate does not hold on the other. `status` agrees in
every case; only the human-readable payload diverges.

Found by #676 Stage 6 while building the cross-substrate timeline parity guard. Measured, by
driving the real core loop and the real runner `ExecuteTool` handler over the same scripted
trial:

| case | core substrate records | runner substrate records |
|---|---|---|
| tool raised | `{"error": "already refunded"}` (the tool's own `ToolResult.error`) | `Tool error: RuntimeError: {"error": "already refunded"}` |
| tool raised, wrapped | `Tool execution failed: <e>` (`tolokaforge/tools/registry.py:406`) | `Tool error: <Type>: <msg>` (`tolokaforge/runner/service.py:1172`) |
| unknown tool | `Tool 'nope' not found` (`registry.py:317`) | `Tool 'nope' not found in agent tools` |

## Why it matters

This is the **same drift shape #676 exists to remove, one field over.** #676 unified the event
timeline so that grading sees one shape from both substrates — and G5 already settled that the
*record* is authoritative over the message. But the record's own `result` text is not portable,
so the unification is complete for `status`, `arguments`, `call_id`, `tool_name`, `executor` and
`latency_seconds`, and **incomplete for `result`.**

The concrete consequence is for **#678**: the first time a task author writes a trace check
matching on the result text of a *failed* call, that check passes on one substrate and fails on
the other. Since #678's whole premise is that deterministic checks are more reproducible than an
LLM judge, a substrate-dependent matcher is worse than the thing it replaces.

## Why Stage 6 did not fix it

Converging the text moves core-substrate verdicts and canonical grading snapshots, which is out
of scope for a foundation stage whose contract is "change plumbing, not verdicts". Stage 6
instead restricted its parity fixture to successful calls and **documented the limitation** in
`tolokaforge/core/grading/trace_timeline.py`'s module docstring and in `docs/GRADING.md`, rather
than weakening the compared field set to hide it. That was the right call, and it is why this
needs its own issue.

## Deliverables

- [ ] A failed call's recorded `result` is byte-identical across substrates for the same
      underlying failure, or the difference is confined to a field the matcher vocabulary does
      not expose
- [ ] The rejection texts converge too (`not found` vs `not found in agent tools`)
[...]

Issue #715, opened 2026-07-31
Resume path decides re-runs by substring-matching trajectory.yaml, ignoring the typed reason beside it


azorej:
## Summary

The resume path decides whether to re-run a trial by **text-matching a recorded
`trajectory.yaml`**, even though the typed evidence it needs is now persisted right beside the
text it greps.

`tolokaforge/core/resume.py:220` `_has_infrastructure_error(task_id, trial_index)` loads
`trajectory.yaml` and looks for "429 / status=error" shapes; `:279-283` turns that into the
re-run decision. `tolokaforge/core/runner.py::_is_rate_limit_error` is the same substring
approach on the live path.

## Why this is now worth changing

Before #676 Stage 5 there was no better source, so text matching was the only option. Stage 5
changed that:

- `TerminationReason.RATE_LIMIT` is now set from **typed** evidence only —
  `is_typed_rate_limit_exception` walks `__cause__` for `isinstance(exc, openai.RateLimitError)`
  or `status_code == 429` (`tolokaforge/core/llm/client.py`). A 429-shaped *message* with no
  typed exception deliberately no longer produces that reason.
- `classify_trial_outcome` (`core/failure_attribution.py`) turns `(status, termination_reason)`
  into `MEASURED` / `HARNESS_ERROR` / `INFRASTRUCTURE_ABORT`, and both inputs are persisted in
  `trajectory.yaml`.

So the resume path can ask the same question the metrics layer asks, from the same data, and
get an answer consistent with it. Today the two can **disagree about the same trial**: metrics
classify it `MEASURED` (no typed 429) while resume sees "429" in a message and re-runs it as
infrastructure flake — or the reverse.

## Not currently a scoring bug

Neither function sets a termination reason, so **neither can gate exclusion from the rate
denominator**. This is not invisible inflation, and it is correctly outside Stage 5's excluded-
reason manifest. The cost is different: re-run decisions are made on weaker evidence than the
harness already has, so a trial the agent genuinely failed with an error message quoting a rate
limit can be retried as though the provider killed it — quietly buying extra attempts for
exactly the failures that most look like flakiness.

## Deliverables

- [ ] `_has_infrastructure_error` derives from the persisted `termination_reason` /
      `classify_trial_outcome`, not from substring matching
- [ ] `core/runner.py::_is_rate_limit_error` reuses `is_typed_rate_limit_exception`
- [ ] Retryability and countability stay **separate predicates over one classification** —
[...]

PR #722, opened 2026-07-31
feat(grading): complete the grade-time trajectory event record, and one shared timeline both substrates read

### LINKED ISSUE #676: feat: Complete the grade-time trajectory event record
## Context

Deterministic trace checks need a single ordered view of what happened in a trial: tool calls, tool results, and messages on one timeline. Today grading receives two disjoint views — a tool-call history and a message list — with **no shared identifier**, so a tool call cannot be reliably linked to the result it produced. Argument-level and result-level matching are unsound without that link, and the harness's own termination reason is not available to grading at all.

This is the foundation for [trace-check matchers](#) and blocks every constraint that spans a call and its result, or a call and a message.

## Deliverables

- [ ] Every recorded tool call carries a stable call identifier and a monotonic sequence position
- [ ] The transcript crossing the wire to grading preserves the same call identifier on assistant tool calls, so a call and its result message are linkable by id rather than by position
- [ ] User-simulator tool calls are represented on the timeline with the same fidelity as agent tool calls, or their absence is explicit and documented
- [ ] The trial's termination reason is available at grade time
- [ ] A shared timeline builder produces one ordered, interleaved event sequence (tool call, tool result, assistant message, user message) with per-event position and turn index, and is used by both grading substrates
- [ ] Protocol documentation describing what the grading transcript contains is accurate (it currently claims tool results are excluded; they are not)

## Acceptance Criteria

- Given a trial where the same tool is called twice with identical arguments, grading can tell the two calls apart and attach the correct result to each
- The unified timeline round-trips a recorded run: replaying it yields the same event order the trial actually executed
- Termination reason appears in grading inputs and is distinguishable between a normal agent finish and an infrastructure abort (timeout, rate limit)
- Both grading substrates build their timeline from the same code path, verified by test

## Design Decisions

- **Link by identifier, not by position**, because positional joins are ambiguous exactly where it matters most — repeated identical calls, which idempotency checks exist to catch.
- **One shared timeline builder used by both substrates**, because two independent implementations is how the drift fixed in the parity ticket happened.
- **Termination reason is exposed to grading but not matchable by task authors in this ticket.** Task grading must not depend on infrastructure flakiness; deciding whether authors may branch on it is a separate call.
- **Tool output is carried untruncated**, because result matching on a truncated string silently produces wrong verdicts. On a failed call the recorded output is the harness's error text rather than tool output, and that distinction must be documented.

## Out of Scope

- Any matcher or constraint evaluation — this ticket produces the data, not the checks
- Changing what the LLM judge receives

### LINKED ISSUE #689: fix: infrastructure aborts are graded as task failures before `GradeTrial` is ever called
## Context

Discovered while planning #676, which sets out to make the trial's termination reason available at grade time. It turns out the reasons that matter most **never reach grade time at all**.

`RunnerRPCTrialGrader.grade` (`tolokaforge/core/trial_grader.py:110-141`) short-circuits before the `GradeTrial` RPC on two conditions:

- `trajectory.status in (TrialStatus.ERROR, TrialStatus.TIMEOUT)` → `Grade(binary_pass=False, score=0.0, components=GradeComponents(state_checks=0.0), reasons=f"Trial failed with status: {status}")`
- `trajectory.termination_reason == TerminationReason.STUCK_DETECTED` → same shape, `reasons="Agent got stuck (...)"`

`classify_loop_error` (`core/loop.py:157-188`) assigns `TrialStatus.ERROR` to **all** of `API_TIMEOUT`, `RATE_LIMIT`, `API_ERROR`, and `ERROR`; `PROVISION_ERROR` is also `ERROR`, and the episode timeout is `TrialStatus.TIMEOUT`.

Measured directly (in-process, `RunnerRPCTrialGrader` with a stubbed runtime backend):

```
status     termination_reason GradeTrial called?   host-side grade
completed  agent_done         True                 (runner grades)
completed  user_stop          True                 (runner grades)
completed  max_turns          True                 (runner grades)
completed  stuck_detected     False                'Agent got stuck (repeated actions without progress)' score=0.0
error      rate_limit         False                'Trial failed with status: error' score=0.0
error      api_timeout        False                'Trial failed with status: error' score=0.0
error      api_error          False                'Trial failed with status: error' score=0.0
error      error              False                'Trial failed with status: error' score=0.0
error      provision_error    False                'Trial failed with status: error' score=0.0
timeout    timeout            False                'Trial failed with status: timeout' score=0.0
```

## The problem

A trial killed by a provider rate limit, an API timeout, or a substrate provisioning failure is recorded as a **task failure with score 0.0** — indistinguishable in the aggregate from an agent that tried and got the task wrong. That inflates measured task difficulty by however much provider flakiness the run experienced, and the per-trial `Grade` carries no signal that the cause was infrastructure.

The reason is also structurally unfixable from the grading side: only `agent_done`, `user_stop` and `max_turns` ever reach `GradeTrial`, so a `termination_reason` field on `GradeTrialRequest` (#676) makes the three *reachable* reasons legible and cannot touch this class at all.

`trajectory.termination_reason` and `core/failure_attribution.py` do carry the honest cause into the output bundle — the loss happens specifically in the `Grade`, which is what pass-rate and score aggregation read.

## Deliverables

- [ ] An infrastructure-aborted trial is distinguishable from a task failure in the `Grade` / aggregation, not only in `trajectory.yaml`
- [ ] The decision is explicit about whether such a trial is excluded from pass-rate denominators, retried, or scored — and the choice is documented in `docs/GRADING.md` / `docs/ANALYTICS.md`

## Acceptance Criteria

- Given a trial that terminated with `rate_limit`, aggregation does not report it as a failed attempt at the task.
- Given a trial that terminated with `agent_done` and got the task wrong, nothing changes.

## Notes

Adjacent to #677 (degenerate trials can pass) — that ticket is about a trial that *ran* and did nothing; this one is about a trial that never got to run. Keep the two verdicts distinguishable.

### LINKED ISSUE #705: fix(core): the agent's done-marker check can never match, so AGENT_DONE is unreachable on the core substrate
## Context

Found during #676 Stage 2. `TrialRunner._is_done` lowercases the text it searches but not the marker it searches for:

```python
done_markers = ["###STOP###"]
text_lower = text.lower()
return any(marker in text_lower for marker in done_markers)
```

`text_lower` contains no uppercase characters, and the marker is uppercase, so the test can never succeed. Verified by inspection and by measurement — `_is_done("###STOP###")`, `_is_done("done ###STOP###")` and `_is_done("I am finished. ###STOP###")` all return `False`.

**Consequence: `TerminationReason.AGENT_DONE` is unreachable on the core substrate.** A trial can only end via the user simulator's own `###STOP###` handling or by exhausting `max_turns`. An agent that correctly signals completion is recorded as having run out of turns.

## Why this matters beyond the one-line fix

Two pieces of in-flight work depend on distinguishing a deliberate finish from an exhausted budget:

- **#676 Stage 4** puts the termination reason on the grading wire specifically so grading can tell them apart. Umbrella scenario 8 ("termination discipline" — did the agent actually report back, or did it get cut off?) is unevaluatable while every core-substrate trial reports `max_turns`.
- **#677** adds a lower bound on agent activity, and its whole premise is separating "the agent did nothing" from "the agent finished".

It also means recorded runs on this substrate carry a termination reason that misdescribes what happened, so any analysis that groups by termination reason has been reading a constant.

## Deliverables

- [ ] The done-marker check matches the marker the prompt actually instructs the agent to emit, with case handling that is correct rather than accidental
- [ ] A behaviour-locking test that fails if the comparison is broken again — case-mismatch bugs are invisible to a test that only checks the happy path with the exact literal
- [ ] The reachability of every `TerminationReason` on both substrates is either tested or documented, so a member that no code path can produce is caught rather than assumed
- [ ] Whether historical recorded runs should be re-interpreted is decided and stated — their termination reasons are wrong, not merely stale

## Acceptance Criteria

- An agent emitting the done marker terminates with `AGENT_DONE`, on both substrates
- The marker check is case-insensitive in a way a reader can verify at a glance, or deliberately case-sensitive and matching the prompt exactly
- No `TerminationReason` member is unreachable without that being recorded

## Design Decisions

- **Fix the comparison, do not just lowercase the marker.** A bug that comes from lowercasing one side of a comparison is a symptom of comparing in two steps; normalising both sides in one place is what stops it recurring.
- **The reachability question is the durable part.** This is the same defect class #676 and #675 keep surfacing: an enum member, a config key, or a code path that looks supported and cannot fire. `#676` Stage 2 added a canonical lock asserting every `ToolExecutionStatus` member is reachable from some recording path; the same treatment for `TerminationReason` would have caught this.

## Out of Scope

- Changing the done marker itself or the prompt that instructs the agent to emit it
- The user simulator's separate `###STOP###` handling, which works

## Dependencies

- **Should be resolved before #676 Stage 4**, which is otherwise building a wire field to carry a distinction the core substrate cannot make.
- Related: #677.

### LINKED ISSUE #716: docs/OUTPUT_FORMAT.md's tool_usage example does not match what the writer emits
## Summary

`docs/OUTPUT_FORMAT.md:378` documents `metrics.yaml`'s `tool_usage` block with a shape the
writer does not produce — the example shows per-entry `tool` / `count` / `success` / `fail`
keys that disagree with what the code emits.

A previous CHANGELOG entry records this disagreement as corrected, but **the example block
itself was missed**, so the doc still teaches the wrong shape while the changelog claims it was
fixed. Anyone parsing `metrics.yaml` from the documentation writes code against a structure
that never appears.

Pre-existing; found while updating `docs/OUTPUT_FORMAT.md` for #676 Stage 5 and left alone to
keep that stage's diff scoped to the keys it added.

## Deliverables

- [ ] The `tool_usage` example matches what the writer actually emits, verified by generating a
      real `metrics.yaml` and diffing the block rather than by reading the code
- [ ] A grep for other example blocks in `docs/OUTPUT_FORMAT.md` that drifted the same way —
      if one example was missed by a correction, its siblings are suspect

## Why bother

AGENTS.md Rule 8 keeps docs current, and an output-format document is consumed by people
writing parsers. A wrong example is worse than a missing one: it looks authoritative, and the
changelog actively asserts it is right.

azorej:
Closes #676. Closes #705. Closes #689. Closes #716.

Targets the integration branch `feat/deterministic-trace-checks`, not `main`.

## What this is

The grade-time trajectory record was lossy in ways that made deterministic trajectory grading
impossible to build on. This lands the record, and lands one shared timeline both grading
substrates read. It is the foundation #678–#682 consume — **#678's contract is already relayed on
that issue with its five build-affecting consequences.**

18 commits, 105 files, +8150/−1606. Seven plan stages plus a concurrent-PR reconciliation plus one
review round.

## The problem, concretely

- A tool call could be linked to its result only **by position** — ambiguous exactly where it
  matters, since a provider can return two calls to one tool with byte-identical arguments in a
  single response.
- **Four incompatible per-executor tool-log shapes**, with no shared clock, concatenated rather
  than interleaved. Cross-executor order was destroyed by construction.
- Rejected calls recorded nothing while the transcript gained a `role: tool` error, so a trial full
  of rejected calls reported zero failed calls.
- Latency came from a field the *tool* was expected to set and none does, so `total_duration_s` was
  `0.0` for every successful call.
- `TIMEOUT` was **unrecordable on the production path** — the client collapsed the wire status to a
  bool one line before recording.
- A rate-limited trial was scored `0.0` and counted as a task the model failed.

## Stages

| Commit | Stage |
|---|---|
| `7cdc4f3` | 1 — the tool-call id crosses the wire, end to end |
| `1bbb0b1` | 2 — one recorded-tool-call type, recorded once, in true order |
| `98e5f2d` | 3 — the transcript wire encoder and its inverse, in one module, with the id |
| `a8d2102` | 4 — the termination reason at grade time (**closes #705**) |
| `c3d8bce` | — reconciling #704's concurrent `custom_checks` landing |
| `a3bde1f` | 5 — an infrastructure abort is not a task failure (**#689 absorbed**) |
| `7141f0f` | 6 — one shared trial event timeline, joined by call id |
| `868ab6c` `591d77e` `db91b04` `af40d2b` | 7 — both substrates grade off the timeline |
| `0e75816`…`4edd0d3` | review round 1 (7 commits) |

## Key design choices

| Decision | Rationale |
|---|---|
| One `RecordedToolCall`, recorded once, by the **caller** rather than each executor | The caller is the only party that knows both the provider `call_id` and the executor identity. Moving the site there fixed three independent defects that were all bypasses of the old append |
| Arguments recorded **raw** on both substrates; redaction moves to artifact-write time (#694) | Recorded calls are the grader's *input*. Rewriting a `token`-named value made a trial fail an argument match it should have passed — on one substrate only |
| Version gates fail at `RegisterTrial`; unrecognised enum values are refused, not coerced | A per-call protocol rejection reaches the agent as an ordinary tool error: it retries until the budget is gone, the trial reports `completed`, and the skew is visible only inside the transcript |
| An encoder ships with its inverse, in one module, with a byte-identical round-trip property | The encoder previously lived alone with **no decoder anywhere**, which is why the dropped id went unnoticed — nothing could compare the two directions |
| Reachability is asserted by **driving real paths and asserting over what was observed**, never a hand-written table | A table is written from the same belief as the code, so it agrees with the bug. #705 survived exactly this way |
| An infrastructure abort produces **no grade** and leaves the rate denominator | The errored-judge rule one level up: a component that was not evaluated is excluded from numerator and denominator rather than folded in as a zero |
| Unknown/future termination reasons default to **counted**, not excluded | Asymmetric harm: silent exclusion is invisible inflation; wrongly counting deflates visibly and boundedly |
| One flat `TraceEvent`, not a four-way union | #678's matchers are conjunctions of field predicates and want uniform field access |
| The timeline builder takes `Sequence[Message]`, one code path | Two normalising branches inside the builder would move the drift *inside* the thing built to prevent it |

## Compatibility — read this before deploying

- **`ExecuteToolRequest.call_id` is required** and **`RegisterTrialRequest.engine_protocol_version`
  locks the engine to the runner image.** An engine older than the image is refused at
  registration, naming the skew. **`make docker-build-core` is required.**
- **Metrics semantics changed.** `success_rate` / `avg_score` / `pass@k` are now over
  **measured** trials. `total_trials` keeps meaning "attempted". **Anyone comparing new numbers
  against previously published ones is comparing different denominators.** Both schema stamps are
  bumped (`AGGREGATE_SCHEMA_VERSION`, `TRIAL_BUNDLE_SCHEMA_VERSION`) so stale files are
  machine-detectable rather than a changelog sentence.
- Previously recorded runs **are** reinterpretable without new data — the classifier's inputs are
  already persisted and aggregation is recomputed from trajectories.
- One deliberate verdict change: a trial whose only tool call is an unexecuted terminating-turn
  declaration, on a timeline with **no records at all**, now fails `disallowed_tools` by name
  instead of passing silently. In the CHANGELOG.

## Review

Three sharded reviewers, all **REQUEST CHANGES**, all addressed in round 1.

The blocker worth naming: **every loud grading failure was re-scored as `0.0` one call frame up.**
`GradeTrial` correctly returned `success=False` with no `Grade`; the host then manufactured
`Grade(binary_pass=False, score=0.0)`. Three documents in this very diff promised the opposite, and
**CI pinned both halves** — the new test pinned the runner side, a pre-existing test pinned the host
side to `score == 0.0`. No test joined the two layers, which is why it survived. The host now raises
`GradingFailedError`, the pre-existing test was rewritten rather than weakened, and a new
`TestDuplicateCallIdPublishesNoScore` drives both layers together over a real duplicate `call_id`.

Also fixed: `avg_score_micro` imputed a score to every measured-but-ungraded trial (one harness
error either halved the run's headline or inflated it — measured `0.5` when the only score in the
run was `1.0`); a records-less timeline **silently passed every `disallowed_tools` check**, which is
reachable through published API on any recorded bundle and violates this branch's own G6b; and
`decode_transcript_wire` raised bare `KeyError` past its documented `ValueError` contract.

**Known gap, filed not hidden: #720.** With the fabricated `0.0` gone, an ungradeable trial now
*vanishes* from `total_trials` and writes no bundle, because the raise leaves `Conductor.run`
uncaught and `future.result()` precedes `self.results.append`. The failure is loud (log, retries,
`run_state.json`, terminal event) and no score is fabricated — but the aggregates under-count
silently. Resolving it means deciding how an ungradeable trial *classifies*, which changes what a
published number means and deserves the scrutiny Stage 5's table got. **P1, on the milestone.**
Also **#721** — judge replay silently excludes grade-less bundles, which this PR made routine.

## Verification

- canonical **775 passed / 0 failed** — and **no grading snapshot moved**, which is how we know the
  refactor changed no verdict. The one snapshot that did move is `run_aggregate`, expected and
  enumerated.
- unit **4010 passed / 8 failed** — the 8 are exclusively the pre-existing environment-sensitive
  `tests/unit/test_persistent_shell_local.py` set, red on `main` too.
- integration **6/6** (`test_docker_grading{,_jsonpath,_tool_expectations}.py`) on a runner image
  rebuilt from this tree. Plus `test_runner_call_id_grpc.py` + `test_runner_cleanup_trial_grpc.py` → 9 passed.
- `ruff check` clean; both formatters clean on every touched file.

Every guarantee added here has a test that fails when the guarantee is broken, proven by
patch-and-restore with the failure quoted. That discipline caught the most instructive result on the
branch: the cross-substrate equality assertion, **alone**, passed when the builder was patched to
emit nothing — two empty timelines compare equal. Its three companion tests are what make it bite.

## Discovered and filed along the way

#710 · #711 · #712 · #713 · #714 · #715 · #716 (closed here) · #717 · #720 · #721


CiroGamboa (approved):
LGTM on correctness. Verified all four Compatibility claims against the actual code — proto changes (`call_id` + `engine_protocol_version`), metrics semantics + schema-version bumps, aggregation reinterpretability from persisted classifier inputs, and the records-less-timeline `disallowed_tools` verdict change all check out cleanly. The deliberate security-redaction change on the core substrate is loud and motivated (parity with the runner substrate; tracks the correct future move at #694). Follow-ups #720/#721 are honest scope disclosures, not silent regressions — landing this PR is still the right move; the previous state was silent corruption.

Not a request for changes; discussion questions below that this PR opens up architecturally. Worth naming so the modularity story stays honest as we build on top.

---

## Architectural questions this PR opens up

Four things get *tighter* here: proto version-lock, canonical `TrialTimeline`, no-fabricated-zeros, closed metrics categories. Each is defensible correctness work — but each also shifts what the "pluggable everything" story can look like downstream. Worth naming which are deliberate contractions.

### 1. Extension story for `TraceEvent` types

`TrialTimeline` becomes framework-owned. That's the right call for `trace_checks` to reason declaratively over a fixed vocabulary. But it means an adapter that wants to emit a domain-specific event type has three options: (a) shoehorn into an existing shape, (b) petition the framework, (c) route the concern through `custom_checks` where arbitrary structure is legal.

I read the platform stance from #684 §7 as: **(c) is the answer** — `trace_checks` is the closed declarative surface; `custom_checks` remains the escape hatch for anything else. If that's the intent, worth codifying it (in ADR-0011's cross-references, or in a future ADR alongside trace_checks) so adapter authors don't try to extend the timeline schema and get pushback later.

### 2. "Grade may be `None`" is a real semantic shift

The `GradingFailedError` propagation change kills the "every trial has a grade" invariant. #720 is filed as P1 for one aspect (ungradeable trials vanish from `total_trials`), but the broader implication is: anything downstream that consumes trial results — rejudge, aggregation, judge-replay, dashboards, third-party analytics — has to handle the "trial had no grade" case explicitly. This PR schema-bumps the aggregation-side files so stale ones are machine-detectable, which handles it at the data layer. The question is whether **all downstream code paths that assumed a grade exists have been surveyed**. If a broader sweep is planned, worth naming as follow-up. If not — worth doing before the change propagates to other consumers.

### 3. Third-party engine/runner story

`engine_protocol_version` at RegisterTrial is a hard version-lock. Right call for correctness inside the monorepo (engine + runner ship together, version numbers move in lockstep). Constraining if tolokaforge grows a plugin ecosystem where third parties ship their own engine or runner images — they need to implement the version-check contract and coordinate bumps.

Not urgent. But if the ADR-0022 vision (runtime independence, language-agnostic runners) matures, the version-lock might want a formal negotiation protocol rather than a hard reject. Worth naming as a future architectural conversation on the epic.

### 4. Classification-category extensibility

`infrastructure_aborts`, `harness_errors`, `outcomes_by_reason` — are these closed enums the framework owns, or open sets adapters can extend?

If closed: an adapter that surfaces a novel failure mode has no first-class category for it; goes into a bucket that doesn't match its semantics. If open: the metrics.py enum needs an extension point.

Same shape question as #1 — probably intentional to keep it closed (declarative matchers over a fixed vocabulary), but worth stating so it's not accidentally load-bearing.

---

**Two docs suggestions** for the Compatibility section (optional, not blocking):

- Rebuild-order note: worth spelling out that the runner image must be rebuilt *before* any engine at the new version rolls, since a new engine talking to an old runner image passes ignore-unknown-field defaults then fails later on missing `call_id`.
- Metric-movement direction: rates trend *upward* on runs that hit provider flakiness (rate-limited trials no longer drag averages down). The CHANGELOG has this with concrete numbers; a one-liner in the compat section would help operators comparing dashboards.

None of the above blocks merge. Approve.

azorej:
Thanks — these are the right four questions, and two of them turned into work. Answering each, with what I measured rather than what I assume.

## 2. "Grade may be `None`" — I ran the survey rather than promising one

This was the one concretely checkable item, so I did it before replying. **Every consumer of `trajectory.grade` in the engine already guards for `None`:**

| Surface | Sites | State |
|---|---|---|
| `trial_executor.py` | `:205`, `:238` | guarded |
| `orchestrator.py` | `:1790`, `:1791`, `:1800`, `:1810` | guarded |
| `failure_attribution.py` | `:101`, `:161`, `:197`, `:209` | guarded (`:103/:168/:175/:198/:211` deref inside those blocks) |
| `metrics.py` | `:246`, `:295`, `:320`, `:354` | guarded |
| `artifacts.py` `:456`, `output_writer.py` `:257` | | guarded |
| `shared_stack_runtime.py` `:542` | | guards `response.grade` |
| `dx/cli/main.py` `:1137` | | guarded by `if grade:` — a missing `grade.yaml` skips the section |
| `core/grading/replay.py` | — | no `.grade` deref at all |
| `scripts/`, `tools/`, `external_adapters/` | — | no hits |

There is **no unguarded `trajectory.grade.<attr>` anywhere.** That is partly because the `None` case is not new — Stage 5 introduced it for infrastructure aborts and the guards went in then; `GradingFailedError` widened the *cause*, not the shape.

Two known gaps, both already filed, and the survey **shrank one of them**: `TrialCard.binary_pass` is *already* `bool | None` (`live_panel.py:358`), so #714 is only two Protocol signatures and one `else False` at `orchestrator.py:1798` — not the model widening the issue implied. I've updated #714 with the measurement. The other is #721 (replay discovery). Beyond those I found nothing needing a sweep, so I'm not filing a survey follow-up — the survey is this table.

## 1 + 4. Closed vocabularies — deliberate, and you're right that it isn't written down

Both are the same question and the answer to both is **closed**, with `custom_checks` as the sanctioned escape hatch. Your read of #684 §7 is correct.

The reasoning for the event vocabulary is #684 §3.10's: no arithmetic, no loops, no derived values — *"the fence is what keeps this convertible."* Convertibility to other harness formats was the original reason to build a DSL instead of resurrecting Python checks, and since #704 the escape hatch is a **live** surface rather than a hypothetical, which is what makes the fence affordable.

The classification is closed on **stricter** grounds, which is worth separating out: an adapter-supplied category would let an adapter decide *what leaves the rate denominator* — i.e. move a published benchmark number from outside the framework. Exclusion is deliberately earnable only by typed evidence (`EXCLUDED_TYPED_REASONS`, enforced by `test_every_excluded_reason_is_earned_by_typed_evidence`), and an open set would be a hole straight through that guard. The extension point that *does* exist is `outcomes_by_reason`, which records every reason observed with the class it was counted as — so a novel failure mode stays visible and recomputable from the aggregate, it just doesn't get to define its own denominator.

Your framing — *"worth stating so it's not accidentally load-bearing"* — is the argument, and #684 is currently the only place it lives. Filed as **#725** for an ADR, with pointers from `docs/GRADING.md` and `trace_timeline.py` so an adapter author hits it where they're already reading. Five issues are about to build on this vocabulary; better declared than discovered.

## 3. Third-party engine/runner version lock

Agreed on all counts, including that it isn't urgent. A hard reject is right while engine and runner ship from one tree, and a negotiation protocol is the right shape if ADR-0022's runtime-independence vision matures. Noted on the epic rather than filed, since there's nothing actionable until then.

## Both docs suggestions landed — `79bba25`

**Rebuild order.** You were right, and the existing text implied the hazard without stating it. `docs/RUNNER.md` said the bound was one-sided; it didn't say the unprotected direction is the *quieter* one. Verified the gate is `request.engine_protocol_version < ENGINE_PROTOCOL_VERSION`, so:

- old engine + new image → refused at `RegisterTrial`, loudly
- new engine + old image → **registers fine**, because proto3 drops the unknown field, and then `call_id` is silently ignored too, so calls are recorded without the id grading joins on

Now explicit, with the rule: **rebuild the image before rolling the engine.**

**Metric direction.** Added to `docs/ANALYTICS.md` with the point that matters for a dashboard reader — every rate is *weakly higher*, the gap is exactly the abort count, and a dashboard that appears to improve on the day this lands has not improved; it stopped counting trials the provider killed as trials the model failed.

## Also fixed since your review — `000b0a5`

Both nits from the automated hygiene pass: `resolve_tool_output()` now sits beside `resolve_tool_status` so the failed-call record text has one home (and it's where #717's substrate convergence belongs), and three `dict[str, any]` annotations became `Any`. I deliberately did **not** fold in the neighbouring `f"Error: {...}"` expressions — that's the *message* text the agent reads, and collapsing it into the *record* text would put a human-facing prefix into grading input, which is the mistake Stage 2 removed when it deleted redaction from the recording site.

`lint` is green now too, and the cause was worth naming: `pyproject.toml` pinned `black==26.3.1` while `.pre-commit-config.yaml` pinned `25.11.0`, under a comment saying they must match. The two versions reformatted each other's output, so no commit could satisfy both — which is why `lint` had been red on `main` since before this milestone, including the v0.13.0 release commit. Fixed in `1aecb96`; the 17-file reformat is verified semantically empty (byte-identical ASTs, 17/17).

bberkes-toloka:
One gap in the same class this PR just closed, on the sibling functions.

`check_tool_expectations` got the `records_present` gate (`tolokaforge/core/grading/transcript.py:108`). `check_must_contain` (`:40`) and `check_disallowed_regex` (`:61`) did not, and both read `_searchable_text(timeline)`, which sources tool text exclusively from `TOOL_RESULT` events, i.e. exclusively from records.

On a records-less timeline the searchable string therefore loses every tool result. That state is the one this PR already identifies as re-grading a recorded bundle (`BaseAdapter.grade` -> `GradingEngine.grade_trajectory`, `tool_log` never persisted). The text is not actually absent from the bundle: `write_trajectory` persists every `role: tool` message with its full content and its `tool_call_id`, and `_KIND_BY_ROLE` in `trace_timeline.py` maps only `ASSISTANT` and `USER`, so the timeline drops evidence that is sitting on disk.

Both directions are wrong, and one of them is wrong silently:

* `disallow_regex`: a violation whose text lives in a tool result passes unconditionally on re-grade. This is the `disallowed_tools` failure mode, and it is the thing `TrialTimeline`'s own docstring rules out ("never a silent pass").
* `must_contain`: a phrase that lives in a tool result fails on re-grade.

The CHANGELOG does name the adjacent narrowing ("a failed call's searchable text is now the executing layer's record rather than the differently-worded `role: tool` message"), so the area was considered rather than missed. That reasoning holds only while records are present. Without them, "differently worded" becomes "absent", which is a different verdict rather than a different wording.

Nothing pins it either way: every new test touching the phrase rules builds a timeline that carries records. `test_a_tool_result_is_searchable` pins the records-present half, and the records-less half is unpinned, which is why the gap survived the same review round that produced the `check_tool_expectations` gate.

Two shapes of fix, not equivalent:

1. Mirror the `check_tool_expectations` gate: name the state and fail the sub-check rather than search a string the missing view was supposed to fill. Removes the silent pass, keeps the replayed bundle unevaluatable.
2. Let the timeline fall back to `role: tool` messages when `records_present` is false, joined by `tool_call_id`, leaving `status` / `executor` / `latency_seconds` as `None`. The evidence exists, so this makes the phrase rules evaluate correctly on a replayed bundle, and leaves `records_present` gating only the checks that genuinely require a record.

Live grading is unaffected either way, so this does not touch the substrate-parity result on a normal run. It changes what a re-graded bundle reports.

PR #723, opened 2026-07-31
feat(automation): report gateway availability and accept a route directive

balazs-toloka:
> **Stacked on #718** (`feat/configurable-llm-proxy-transport`). Base is that branch, not
> `main`, because this uses the gateway transport it adds. Review #718 first; this diff is
> only the automation layer.

## Summary

Teaches the Slack-triggered integration flow about the LLM gateway, in two opt-in ways:

1. **Reports availability.** When the gateway secrets are configured, the poller says, per
   resolved model, whether that model is *also* reachable through the gateway.
2. **Accepts a route directive.** `integrate Grok 4.5 via litellm` (also `through the
   gateway`, `using the proxy`, `via openrouter`, `via OR`). The choice travels in
   `plan.json` and reaches `integrate-model.yml` as a new `route` input, which adds
   `LLM_PROXY_*` to the candidate's `.env` so its probes run over the gateway.

**OpenRouter remains the default**, and the reply now names the effective route explicitly.

## Why availability is advisory, not automatic

A gateway route can be backed by a *different upstream* for the same model name. Verified
during #718: `provider: openrouter` + `anthropic/claude-opus-4.7` reached a gateway's
catch-all `anthropic/*` route, which was **Bedrock**, not OpenRouter. It failed loudly only
because that Bedrock model rejects `temperature=0.0`; a closer-matching route would have
silently evaluated a different serving path.

That makes the route a leaderboard-comparability decision, not a transport detail. The
automaton surfaces the option; a human picks it.

## Two report strengths, deliberately distinguished

| Reply says | Means |
|---|---|
| `also on the gateway as <route>` | an explicit catalog entry — someone configured this model |
| `probably reachable … (matched a passthrough)` | only a `<prefix>/*` wildcard covers it; a live call is the real proof |
| `not on the gateway` | the catalog was read and does not cover it |

Presenting a wildcard match as a configured model would invite a human to pick a route that
then 404s at eval time.

## Example reply

```
<@U1> here is what I could resolve from your request:

:white_check_mark: *Grok 4.5*: starting integration as `x-ai/grok-4.5` via *openrouter*
    ◦ probably reachable on the gateway as `openrouter/x-ai/grok-4.5` (matched a passthrough, not an explicit entry)
:white_check_mark: *Claude Opus 4.7*: starting integration as `anthropic/claude-opus-4.7` via *openrouter*
    ◦ also on the gateway as `openrouter/anthropic/claude-opus-4.7`

_Default route is *openrouter*. To use the gateway instead, re-request with `via litellm`._
```

## Details worth a reviewer's eye

- **`parse_route` returns `None`, not the default**, when nothing was asked. Callers need to
  tell "chose OpenRouter" from "did not ask" — only the former justifies overriding a
  gateway-only model.
- **The directive is stripped before the phrases are split.** Without that, `Grok 4.5 via
  litellm` becomes the model phrase and resolves to nothing, i.e. a valid request would come
  back as an unknown model. Pinned by a test.
- **A model literally named `or`/`openrouter` is not mistaken for a directive** — the regex
  requires a `via`/`through`/`over`/`using` lead-in. Also pinned.
- **Fetch failures are swallowed.** A notification path must never break the poll, so an
  unreachable gateway yields `unknown`. A sentinel distinguishes "not fetched yet" from
  "fetched, no info", so one unreachable gateway is not re-timed-out per request in a poll.
- **The base URL is a secret, not a variable**, because the hostname is typically internal
  and this repo is public.
- **`integrate-model.yml` requires both gateway secrets or neither.** A base URL without a
  key would make litellm fall through to the provider env lookup and forward the *OpenRouter*
  key to the gateway host.

## Degradation

No gateway configured → the lookup returns `unknown`, nothing is reported, `route` stays
`openrouter`, and the flow is byte-identical to today. Asking for `via litellm` without the
secrets logs a workflow warning and probes over OpenRouter rather than failing the run. The
plan reader defaults `route` too, so a plan written by an older poller still dispatches.

## Testing

```
tools/automation/tests/test_gateway_catalog.py    26 passed
tools/automation full suite                      157 passed
ruff check / ruff format --check                 clean
both workflows                                   re-validated as YAML
```

26 new tests: lookup strengths, quiet fetch degradation, the route directive (including the
strip and the false-positive guard), reply shape with and without availability, and 4 pinning
the `plan.json` row since that is the contract with the workflow.

The reply path was also exercised end-to-end in-process against four realistic requests
(unstated / `via litellm` / `via openrouter` / two models + `through the gateway`).

## Not in scope

The gateway route changes the candidate's *probe* transport. It does not yet add a
`pricing.json` entry for gateway-routed model strings, so `cost_source` can be `unknown` on
that path — noted in #718 and harmless for the probes, which do not gate on cost.

### DIFF
--- a/.github/workflows/integrate-model.yml
+++ b/.github/workflows/integrate-model.yml
@@ -40,6 +40,13 @@ on:
         description: "what triggered it (e.g. slack); informational only"
         required: false
         default: manual
+      route:
+        description: >-
+          transport for the candidate's probes: "openrouter" (default) or "litellm" to go
+          through the LLM gateway. The gateway route needs the ARENA_AUTOMATION_LLM_PROXY_*
+          secrets; without them the run falls back to OpenRouter rather than failing.
+        required: false
+        default: openrouter
 
 permissions:
   contents: write
@@ -187,7 +194,24 @@ jobs:
         # metacharacters can never be command-substituted.
         env:
           ARENA_KEY: ${{ secrets.ARENA_AUTOMATION_OPENROUTER_API_KEY }}
-        run: printf '%s\n' "OPENROUTER_API_KEY=$ARENA_KEY" > .env
+          ROUTE: ${{ inputs.route || 'openrouter' }}
+          PROXY_BASE_URL: ${{ secrets.ARENA_AUTOMATION_LLM_PROXY_BASE_URL }}
+          PROXY_API_KEY: ${{ secrets.ARENA_AUTOMATION_LLM_PROXY_API_KEY }}
+        # OPENROUTER_API_KEY is always written: it is the fallback, and the probes' own pricing
+        # lookups use it even on the gateway route. LLM_PROXY_* is added only when the gateway
+        # route was requested AND configured -- writing a base URL without a key would silently
+        # forward the OpenRouter key to the gateway host (see docs/LLM_LAYER.md § proxy).
+        run: |
+          printf '%s\n' "OPENROUTER_API_KEY=$ARENA_KEY" > .env
+          if [ "$ROUTE" = "litellm" ]; then
+            if [ -n "$PROXY_BASE_URL" ] && [ -n "$PROXY_API_KEY" ]; then
+              printf '%s\n' "LLM_PROXY_BASE_URL=$PROXY_BASE_URL" >> .env
+              printf '%s\n' "LLM_PROXY_API_KEY=$PROXY_API_KEY" >> .env
+              echo "candidate probes will run through the LLM gateway"
+            else
+              echo "::warning title=Gateway route unavailable::route=litellm was requested but ARENA_AUTOMATION_LLM_PROXY_BASE_URL / _API_KEY are not set; probing over OpenRouter instead"
+            fi
+          fi
 
       - name: Resolve candidate model_id slug
         id: slug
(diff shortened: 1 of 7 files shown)


CiroGamboa (dismissed):
LGTM. The wildcard-vs-exact strength split and the "None means didn't ask" convention on `parse_route` are the two subtle calls I would have argued for — good that they're already pinned by tests. Falling back to OpenRouter with a `::warning::` when the gateway is asked-for-but-unconfigured is the right degradation.

One small suggestion, non-blocking:

**`_UNFETCHED = []` is a mutable module-level sentinel** (`poller.py:51`). The identity check works today, but the pattern is fragile: anything that ever `.append()`s to a variable while it still holds the sentinel identity would silently poison every future `run()` — `gateway_models is _UNFETCHED` would still be True, the fetch would be skipped, and the mutated list would be treated as a real fetched catalog. Nothing in this diff does that, so this is about closing the failure mode rather than fixing a live bug. `_UNFETCHED = object()` with `gateway_models: list[str] | None | object = _UNFETCHED` removes it entirely at zero cost.

Bonus micro-nit if you're touching this anyway: `fetch_gateway_catalog`'s bare `except → return None` is the right policy for the poll loop, but a single `slack._log(f"gateway catalog fetch failed: {exc!r}")` on the except would give operators a breadcrumb when availability is unexpectedly "unknown" across the board. Skip if not worth the churn.

CiroGamboa (approved):

PR #738, opened 2026-07-31
integrate: azure_ai/cohere-command-a-plus-05-2026

balazs-toloka:
Integrates **Cohere Command A+** (`azure_ai/cohere-command-a-plus-05-2026`) for the arena leaderboard.

The model is served only by the deployment's LiteLLM gateway, not by OpenRouter (its public catalogue carries only `cohere/command-a` and `command-r*`), so this runs over the gateway transport added in #718. Bootstrapped by hand rather than through the Slack request path, to exercise that route end to end.

The live capability run required for `AGENTS.md` rule 1 is quoted below and is green. See "Remaining before this goes on the board" for what is still open, none of which is a code change to this PR.

## What changed (8 files, +457/-3)

| File | Why |
|---|---|
| `core/llm/params_policy.py` | New `supports_tool_choice_auto` axis, default `true`. |
| `core/llm/client.py` | `_build_kwargs` omits `tool_choice` for the value `auto` only when the axis is off. |
| `core/data/model_presets.yaml` | New `cohere_command_a_plus` preset. |
| `core/data/pricing.json` | Azure GlobalStandard retail price for the model. |
| `tests/integration/llm/registry.py` | `ModelCertificate`. |
| `tests/unit/llm/test_params_policy_tool_choice.py` | 15 tests pinning the axis and the whole preset body. |
| `docs/LLM_LAYER.md` | The `GenerationParams` flag table was two axes out of date. |
| `.env.example` | Documents the probes' dedicated opt-in variable. |

## The one engine change, and why it is not a workaround

Cohere's Chat API accepts only `REQUIRED` and `NONE` for `tool_choice`. There is no `AUTO` ([docs](https://docs.cohere.com/reference/chat)), and its documented behaviour when the parameter is **omitted** is "the model is free to choose whether to use the specified tools or not". That is exactly what `auto` means, and `auto` is the only value this codebase ever sends. So omitting that one value reproduces the intended semantics in the provider's own vocabulary; it is permanent and correct, and measurements stay comparable with every other board model.

The axis is deliberately named after the **value**, not the parameter: an explicit `REQUIRED`/`NONE` still goes through untouched. Silently discarding a caller's explicit forcing would be a behaviour change rather than a no-op. (This is a correction made during the PR: the first version was called `supports_tool_choice` and dropped the parameter unconditionally.)

Evidence, live A/B on 2026-08-01: identical client, prompt and tools. Without the parameter the model returns a correct `get_weather({'city': 'Budapest', 'unit': 'c'})`; with `tool_choice="auto"` the call is rejected by litellm before it leaves the process (`UnsupportedParamsError: azure_ai does not support parameters: ['tool_choice']`).

Three proxy-side "fixes" were considered and **rejected**, and the preset comment says so, so nobody retries them: `allowed_openai_params: ['tool_choice']` and a `model_info: supports_tool_choice: true` override would both push the non-existent `auto` through to Cohere, trading a clear client-side error for a provider-side one; `drop_params` gateway-wide would silently discard unsupported parameters for every model and every consumer of that gateway.

Why the engine and not the config: `tool_choice` is attached alongside `tools`, i.e. *after* `params_policy.adapt()` has run, and `ParamPolicy`'s signature has no `tools`/`tool_choice`, so no rewrite inside `adapt` can reach it. `_VALID_PARAMS_KEYS` is introspected from `GenerationParams.__init__`, so the kwarg has to exist in the engine before a preset can set it. The alternative, a model-name branch in `loop.py`, is what AGENTS.md rule 3 forbids.

## Preset

`schema_sanitizer: strict` + `response_policy: array_dict_map` are a pair, and unlike the flag they are route-specific. The Azure serving stack's schema converter cannot resolve `$ref`/`$defs`; it 500s with `json_schema_converter.cc:3370: Cannot find field $defs in #/$defs/_LineItem`, taking down every probe whose tool schema uses a `$defs` block. `strict` inlines refs and drops `$defs`, and also rewrites typed dict-maps to arrays, which `array_dict_map` converts back. Same pairing as `openai_gpt5` and `xai_grok`. `dict_map_hints` was tried and rejected: it contradicts the array schema `strict` sends.

`reasoning_codec: openai` is set even though the thinking probes fail. With the implicit `none`, `NoReasoningCodec.extract` returns `None` unconditionally, so those three entries could not pass whatever the wire carried, which made them unfalsifiable rather than measured. It is also a silent data-loss risk: the Azure catalog advertises reasoning for this model, and `none` would have discarded it for a whole eval. `OpenAIReasoningCodec` is a strict superset, verified below.

## Measurement

Recorded 2026-08-01 against the live gateway: **21 pass / 10 fail / 6 skip** over the full probe suite. The 10 failures are `recursive_ref` x4, `re2_pattern`, `prompt_caching`, `implicit_prompt_caching`, and the three `thinking` probes. The 6 skips are the `TF_RUN_VARIANTS`-gated report-only variants.

Each `known_unsupported` entry carries its own dated evidence and mechanism in the certificate. Two are worth reading before approving:

- **`re2_pattern`** is measured with the sanitiser deliberately bypassed (the probe swaps in `PassthroughSchema`), so it records the Azure validator's own behaviour. Production is not exposed, because `strict` strips RE2-incompatible patterns before the wire.
- **the three `thinking` entries** were measured under codec `none`, where they cannot pass by construction. Be precise about what that means: there is no thinking ratchet (only `enum_slash_tolerance`, `implicit_prompt_caching` and `re2_pattern` have one), and `known_unsupported` makes the capability gate skip the probe, so a later run yields three skips rather than a verdict. The codec is set for the eval path, where `none` would silently discard any reasoning the gateway surfaces; promoting these to `required` needs a deliberate live re-probe with the gate lifted.

An earlier, partially converged run of the same suite read 14/17/6. **Do not read that as "the rejection killed every tool-calling probe"** - most probes send `auto`, so a blanket rejection could not have left 14 of them passing. That intermediate reading is not reconstructible and carries no weight beyond "worse before".

## Pricing

`$0.80/M` in, `$3.20/M` out, Azure GlobalStandard retail, read from the public retail-prices API on 2026-08-01 and confirmed against a live billed call (`$0.000336` for 124 in / 74 out). It is deliberately **not** an OpenRouter number, unlike every other key in `pricing.json`. Provenance is recorded on the certificate, because JSON takes no comment and `pricing-updater` rewrites the `_meta` block wholesale.

## Known defect, not fixed here

Response text arrives wrapped in `<|START_TEXT|>...<|END_TEXT|>` (observed 2026-08-01). `ResponsePolicy` cannot reach it, because that hook only post-processes tool-call arguments, so the markers land in `trajectory.yaml` and in front of any transcript grader or LLM judge, depressing scores for a reason that is not capability. **This needs an engine-side text hook before the model goes on the board.** It is filed on the certificate as a synthetic-vs-production asymmetry per `docs/ADD_NEW_MODEL.md` section 4. `BASIC_COMPLETION` stays `required` and honestly so: the probe asserts non-empty text, and wrapped text is non-empty.

## Live capability run

Run 2026-08-03 against the deployment's gateway, on the branch as it stands:

```
$ TF_COHERE_A_PLUS_GATEWAY_LIVE=1 pytest tests/integration/llm/ -k cohere -q -n 8 -rsp
================== 23 passed, 16 skipped in 68.22s (0:01:08) ===================

SKIPPED [4] conftest.py:141: ...known_unsupported: recursive_ref_tool_call
SKIPPED [1] conftest.py:141: ...known_unsupported: re2_pattern_tolerance
SKIPPED [1] conftest.py:141: ...known_unsupported: prompt_caching
SKIPPED [1] conftest.py:141: ...known_unsupported: implicit_prompt_caching
SKIPPED [1] conftest.py:141: ...known_unsupported: thinking_emits_blocks
SKIPPED [1] conftest.py:141: ...known_unsupported: thinking_replay_roundtrip
SKIPPED [1] conftest.py:141: ...known_unsupported: unsigned_thinking_replay
SKIPPED [3] test_variant_dict_map.py:177: shape-variant suite runs only in the observe stage
SKIPPED [3] test_variant_discriminated_union.py:216: shape-variant suite runs only in the observe stage
```

That reconciles with the certificate exactly, which is the point of quoting it:

* **23 passed** = the 21 `required` capabilities, live, plus the 2 ratchet tests. Both ratchets passing means the `re2_pattern_tolerance` and `implicit_prompt_caching` quirks are still intact upstream, so those two `known_unsupported` entries are confirmed rather than merely asserted.
* **16 skipped** = 10 `known_unsupported` (`recursive_ref` x4, `re2_pattern`, both caching, all three thinking) + 6 `TF_RUN_VARIANTS`-gated variants.
* The 10 skips are node-for-node the 10 failures recorded from the observe run, so the 21/10/6 measurement and this cert-validating run describe the same model.

Zero failures. No preset was changed to get here.

## Verification I can reproduce now

The live run above needed gateway access (the gateway sits behind an Azure Application Gateway; only allowlisted hosts or the corporate VPN reach it). Everything below is reproducible without it.

```
### A) new unit module
tests/unit/llm/test_params_policy_tool_choice.py .............           [100%]
============================== 13 passed in 2.85s ==============================

### B) certificate + preset honesty gates
tests/unit/llm/test_capability_certificate.py ..........                 [100%]
============================== 19 passed in 0.13s ==============================

### C) anti-over-reach gates (a fix must not invalidate an already-valid tool-call arg)
tests/unit/llm/test_policy_array_recovery.py .......                     [100%]
============================== 45 passed in 0.76s ==============================

### D) full unit + canonical
====== 4 failed, 4513 passed, 263 warnings, 3 errors in 91.87s (0:01:31) =======
# the 4+3 are pre-existing and fail on a clean checkout too:
#   3x test_rate_limit_probe_retry::TestTerminalKeyExhaustionIsNotARateLimit
#      (builds an LLMClient without isolating secrets, so a .env carrying
#       LLM_PROXY_* breaks it - see the isolation note below)
#   4x test_adapter_convert_* (wheel-build tests)

### E) lint (CI runs ruff and black via pre-commit)
All checks passed!
4 files would be left unchanged.
```

The certificate gates on a **dedicated** opt-in, `TF_COHERE_A_PLUS_GATEWAY_LIVE`, so any checkout that cannot reach the one gateway serving this route skips instead of firing live calls:

```
$ pytest tests/integration/llm/test_basic_completion.py -k cohere -rs
SKIPPED [1] conftest.py:102: TF_COHERE_A_PLUS_GATEWAY_LIVE not set - skipping live
capability test for openrouter__azure_ai_cohere-command-a-plus-05-2026.
```

Deliberately **not** `LLM_PROXY_BASE_URL`, which is the subtler trap: that variable is a generic public feature, so it means "this checkout has *a* gateway", not "this checkout has *the* gateway that serves this route". A contributor with their own LiteLLM proxy would open the gate, get model-not-found, and the RE2 ratchet accepts `400` / `bad request` as proof its quirk is intact, so it would report green having tested nothing. Worse, with no gateway key set the engine forwards the *provider* credential to the gateway host, so that contributor's OpenRouter key would go to a third party. Same contract as `LLM_PROXY_INT_TEST_API_KEY` in `test_gateway_live.py`.

Preset scope, and that the codec swap is behaviourally a no-op where nothing is surfaced:

```
azure_ai/cohere-command-a-plus-05-2026   codec=OpenAIReasoningCodec   auto=False
cohere-command-a-plus-05-2026            codec=OpenAIReasoningCodec   auto=False
cohere/command-a                         codec=NoReasoningCodec       auto=True
cohere/command-a-plus-05-2026            codec=NoReasoningCodec       auto=True
anthropic/claude-opus-4.7                codec=AnthropicReasoningCodec auto=True
openai/gpt-5.6                           codec=OpenAIReasoningCodec   auto=True
x-ai/grok-4.5                            codec=OpenAIReasoningCodec   auto=True

no reasoning_content: openai -> None | none -> None
replay no-op:         openai -> {}   | none -> {}
```

Note rows 3 and 4: the preset globs cover the gateway's `azure_ai/` route name and a bare alias only. The OpenRouter and direct-Cohere namings lack the literal `cohere-command-a-plus` substring and fall through to `default`. The preset comment says so, so the scope is not mistaken for "holds on any route".

Tests were mutation-checked rather than assumed. Breaking the production code fails exactly the intended test, e.g. widening the suppression from `auto` to all values:

```
FAILED ...::TestBuildKwargs::test_an_explicit_forcing_still_goes_through[required]
FAILED ...::TestBuildKwargs::test_an_explicit_forcing_still_goes_through[none]
```

The new unit module isolates secrets by swapping the `SecretManager` singleton for a `DictProvider`. `monkeypatch.delenv` is **not** sufficient: the providers read the on-disk `.env` and it wins straight back. Verified both ways (`.env` active -> gateway ON; under the fixture -> off).

## Remaining before this goes on the board

1. ~~`AGENTS.md` rule 1~~ **satisfied**, see "Live capability run" above: 23 passed / 16 skipped / 0 failed against the live gateway, reconciling node-for-node with the certificate.

2. The `<|START_TEXT|>` wrapping above needs an engine-side hook before any eval run counts.
3. Content filtering on the Azure deployment could only be set to the lowest blocking level, not disabled. Whether that is a comparability problem for the board is an open question; no provider currently on the board has it disabled either.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -37,3 +37,11 @@ OPENROUTER_API_KEY=your-openrouter-api-key-here
 # litellm transport sends an OpenAI-envelope request. Widen this only if your
 # gateway also serves the other provider's native route (see docs).
 # LLM_PROXY_PROVIDERS=openrouter,openai
+#
+# Opt-in for the gateway-only Cohere Command A+ capability probes. Set this ONLY on
+# a host whose LLM_PROXY_BASE_URL gateway actually serves
+# azure_ai/cohere-command-a-plus-05-2026. Having *a* gateway is not enough: on a
+# gateway that does not serve that route the probes would fire live calls at a model
+# it cannot reach, and one ratchet accepts the resulting 400 as proof its quirk is
+# intact, so it would report green having tested nothing.
+# TF_COHERE_A_PLUS_GATEWAY_LIVE=1
--- a/docs/LLM_LAYER.md
+++ b/docs/LLM_LAYER.md
@@ -701,7 +701,7 @@ class ParamPolicy(Protocol):
     ) -> dict: ...
 ```
 
-`GenerationParams` reads six preset-driven flags:
+`GenerationParams` reads eight preset-driven flags:
 
 | Flag | Default | Effect |
 |---|---|---|
@@ -711,6 +711,8 @@ class ParamPolicy(Protocol):
 | `reasoning_via_thinking_kwarg` | `false` | Budget reasoning → top-level `thinking={"type":"enabled","budget_tokens":N}` (Anthropic-native). |
 | `drop_sampling_when_thinking` | `false` | Pop `temperature` / `top_p` / `top_k` whenever the `thinking` kwarg was emitted (P3b — OpenRouter silently strips them today; Anthropic raw 400s). |
 | `reasoning_budget_default` | `None` | Default `budget_tokens` when `ReasoningConfig(mode="budget")` omits its own budget. |
+| `unsupported_effort_levels` | `[]` | Effort levels known broken upstream; `adapt` raises `ValueError` rather than silently remapping to a level the caller did not ask for. |
+| `supports_tool_choice_auto` | `true` | When `false`, `LLMClient._build_kwargs` omits `tool_choice` for the value `"auto"` **only**; `required` / `none` still pass through. Read outside `adapt`, because `tool_choice` is attached alongside `tools` and therefore after `adapt` has run. For providers whose `tool_choice` enum has no `AUTO`, e.g. Cohere, where omitting the parameter is the documented equivalent. |
 
 ### Reasoning routing matrix
 
--- a/tests/integration/llm/registry.py
+++ b/tests/integration/llm/registry.py
@@ -2474,6 +2474,15 @@ _ALL: list[MC] = [
     # billed call ($0.000336 for 124 in / 74 out). It is deliberately NOT an
     # OpenRouter number, unlike every other key in that file.
     #
+    # SERVING-PATH ASYMMETRY, and it is a board-level comparability decision rather
+    # than a transport detail. Every other certificate in this file is measured on
+    # OpenRouter, which docs/AUTO_INTEGRATION.md calls the calibrated path and keeps
+    # as the default precisely so numbers stay comparable. This model has no
+    # OpenRouter route at all, so its numbers come off Azure serverless through a
+    # private gateway, with a content filter that could only be set to its lowest
+    # blocking level rather than disabled. A human has to accept that before the
+    # number goes on the board; the mechanics below do not settle it.
+    #
     # NOT fixed here, and it has to be fixed before this model goes on the board:
     # response text arrives wrapped in ``<|START_TEXT|>...<|END_TEXT|>`` (observed
     # 2026-08-01). ``ResponsePolicy`` cannot reach it, because that hook only
@@ -2488,14 +2497,26 @@ _ALL: list[MC] = [
         model_id="openrouter__azure_ai_cohere-command-a-plus-05-2026",
         provider="openrouter",
         name="azure_ai/cohere-command-a-plus-05-2026",
+        # A DEDICATED opt-in, and deliberately neither of the two obvious choices.
+        #
         # NOT ``OPENROUTER_API_KEY``: this model is gateway-only, so an OpenRouter
-        # key is not what makes these probes runnable, the gateway transport is.
-        # ``LLM_PROXY_BASE_URL`` is the switch ``resolve_proxy_config`` reads to
-        # decide whether that transport is active, so gating on it lets a checkout
-        # without the gateway skip cleanly instead of firing the whole live probe
-        # suite at a model OpenRouter does not serve. Same contract as the
-        # dedicated variable in tests/integration/llm/test_gateway_live.py.
-        env_key="LLM_PROXY_BASE_URL",
+        # key is not what makes these probes runnable, and gating on it would fire
+        # the whole live suite at a model OpenRouter does not serve.
+        #
+        # NOT ``LLM_PROXY_BASE_URL`` either, which is the subtler trap. That
+        # variable is a generic public feature (see .env.example), so it says "this
+        # checkout has SOME gateway", not "this checkout has THE gateway that serves
+        # this route". A contributor with their own LiteLLM proxy would open the gate
+        # and get model-not-found, and the RE2 ratchet accepts "400" / "bad request"
+        # as proof its quirk is intact, so it would report green having tested
+        # nothing. Worse, when no gateway key is set the engine forwards the PROVIDER
+        # credential to the gateway host, so that contributor's OpenRouter key would
+        # go to a third party.
+        #
+        # So this names the one deployment that actually serves the route. Same
+        # contract as ``LLM_PROXY_INT_TEST_API_KEY`` in test_gateway_live.py: absent
+        # means skip, which is the state of every checkout that cannot reach it.
+        env_key="TF_COHERE_A_PLUS_GATEWAY_LIVE",
         required=frozenset(
             {
                 C.BASIC_COMPLETION,
@@ -2547,13 +2568,27 @@ _ALL: list[MC] = [
                 # The preset therefore now runs ``reasoning_codec: openai``, which is a
                 # strict superset of ``none``: ``OpenAIReasoningCodec.extract`` also
                 # returns ``None`` when the message carries no ``reasoning_content``,
-                # and its ``encode_for_replay`` is a no-op either way. It cannot have
-                # changed any of the 21 passes or the other 7 failures, and it turns
-                # these three into a real measurement on the next run from an
-                # allowlisted network: if the gateway does surface reasoning, the
-                # ratchet tests below fail loudly, instead of the engine silently
-                # discarding reasoning for a whole eval. Declaring unsupported is the
-                # conservative direction, since that is the side the ratchets guard.
+                # and its ``encode_for_replay`` is a no-op either way, so it cannot
+                # have changed any of the 21 passes or the other 7 failures. The one
+                # probe outside these three that touches ``result.reasoning`` is
+                # multi_turn_tool_use, and it only threads it into the next turn
+                # rather than asserting on it; since both codecs replay ``{}``, the
+                # request payload there is byte-identical.
+                #
+                # What it does NOT do is turn these three into a measurement. Be
+                # precise about that: there is no thinking ratchet (only
+                # enum_slash_tolerance, implicit_prompt_caching and re2_pattern have
+                # one), and ``known_unsupported`` makes the capability gate in
+                # conftest.py skip the probe outright, so the next allowlisted run
+                # yields three skips rather than a verdict. The codec is set for the
+                # EVAL path, where ``none`` would silently discard any reasoning the
+                # gateway surfaces. Promoting these to ``required`` needs a deliberate
+                # live re-probe with the gate temporarily lifted.
+                #
+                # One consequence worth knowing when comparing runs: the codec name
+                # goes into the per-trial ``resolved.*`` fingerprint via
+                # ``resolve_policy_names``, so the shipped fingerprint says ``openai``
+                # while 21/10/6 was measured under ``none``.
                 C.THINKING_EMITS_BLOCKS,
                 C.THINKING_REPLAY_ROUNDTRIP,
                 C.UNSIGNED_THINKING_REPLAY,
--- a/tests/integration/llm/registry.py
+++ b/tests/integration/llm/registry.py
@@ -2509,9 +2509,7 @@ _ALL: list[MC] = [
         # this route". A contributor with their own LiteLLM proxy would open the gate
         # and get model-not-found, and the RE2 ratchet accepts "400" / "bad request"
         # as proof its quirk is intact, so it would report green having tested
-        # nothing. Worse, when no gateway key is set the engine forwards the PROVIDER
-        # credential to the gateway host, so that contributor's OpenRouter key would
-        # go to a third party.
+        # nothing.
         #
         # So this names the one deployment that actually serves the route. Same
         # contract as ``LLM_PROXY_INT_TEST_API_KEY`` in test_gateway_live.py: absent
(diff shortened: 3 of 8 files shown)


bberkes-toloka:
## Manual observe record — `azure_ai/cohere-command-a-plus-05-2026`

The auto-integration run for this PR could not complete, so I ran the observe stage by hand
from a network that reaches the gateway. **Verdict: infra-dirty, needs human — twice over,
and both blockers are gateway-side configuration, not the model.**

No certificate is proposed. Writing one now would record `known_unsupported` for capabilities
the model demonstrably has, which the honesty rule in `docs/ADD_NEW_MODEL.md` forbids
("the cert MUST follow the synthetic test result").

### Blocker 1 — the gateway is unreachable from GitHub-hosted runners

The [automated run](https://github.com/Toloka/tolokaforge/actions/runs/30656526064) got, on
every single call:

```
403 Forbidden
Microsoft-Azure-Application-Gateway/v2
```

as an HTML page, not a JSON API error — so the request never reached LiteLLM; it was rejected
at the edge. Evidence that this is a network-position problem rather than credentials:

| Check | Result |
|---|---|
| HTML 403 from the Application Gateway | a bad key would be JSON 401/403 from LiteLLM |
| JSON auth errors in the whole run | **0** |
| Successful calls in the whole run | **0** |
| Same URL + same key + same model from an allowed network | **HTTP 200** |

### Blocker 2 — the gateway rejects `tool_choice` for this model

Running the probes from an allowed network gets past blocker 1 and straight into a second one.
Every tool-calling probe fails with a 400 *before the model sees the request*:

```
litellm.UnsupportedParamsError: azure_ai does not support parameters: ['tool_choice'],
for model=cohere-command-a-plus-05-2026. To drop these, set `litellm.drop_params=True`
or for proxy: `litellm_settings: drop_params: true`
```

This is **not** a model limitation. A/B on the same client, same tools, same prompt:

| Request | Result |
|---|---|
| tools, **no** `tool_choice` | OK — `tool_calls=1` (`get_weather`) |
| tools, `tool_choice="auto"` | 400 `UnsupportedParamsError` |

The model calls tools correctly. The proxy's `azure_ai` provider refuses the parameter and the
deployment does not have `drop_params` enabled, so litellm 400s instead of dropping it — which
is exactly the remedy litellm's own error message names.

### Partial observe results (before I stopped the run)

| Probe | Result |
|---|---|
| `basic_completion` | **PASS** |
| `cost_usd_populated` | **PASS** |
| `allof_merge_tool_call` (3 params) | FAIL — all three the `tool_choice` 400 |

I stopped after that: the remaining ~32 probes are tool-calling probes that would all fail
identically for the same infra reason, and burning the budget to re-confirm one root cause is
not worth it.

### What this PR does contain

A `pricing.json` entry, which is independently verifiable and correct regardless of the
blockers:

```json
"azure_ai/cohere-command-a-plus-05-2026": { "input": 0.8, "output": 3.2 }
```

Azure GlobalStandard paygo, cross-checked two ways: the Azure retail prices API, and a live
gateway call billing 124 in / 74 out at `$0.000336`, which the entry reproduces exactly. The
key is the litellm model string after `normalize_model_name` strips the one leading
`openrouter/`. Cost already returns from the gateway (`cost_source=litellm`), so this is the
local fallback rather than the primary source.

### To unblock

1. **`drop_params: true`** on the gateway for this model (or globally) — one line, and it is
   litellm's own recommended fix. Without it no tool-calling model routed through `azure_ai`
   can be probed or evaluated.
2. **Runner reachability** — the gateway needs to be callable from CI, or the arena eval for
   this model has to run from an allowed network. A self-hosted runner is the usual answer;
   allowlisting GitHub-hosted egress is broad and changes often.

Once (1) lands I can re-run observe and produce a real certificate. (2) is what the leaderboard
run itself will need.

### One more thing worth recording

The model wraps its response text in special tokens:

```
'<|START_TEXT|>OK<|END_TEXT|>'
```

That would contaminate any transcript-based grading. It is not fixable with an existing
`response_policy` — that Protocol only post-processes tool-call *arguments*, never the
assistant text — so it needs a new engine-side hook, i.e. an engine release rather than a
preset overlay. Flagging it now so it is costed into the eval, not discovered mid-run.

bberkes-toloka:
## Correction: the `tool_choice` finding, and retracting my proxy advice

Two things I said earlier in this PR were wrong. The Cohere API docs settle both.

### `tool_choice` omission is Cohere's contract, not a workaround

[Cohere's Chat API](https://docs.cohere.com/reference/chat) accepts only `REQUIRED` and
`NONE` for `tool_choice`. **There is no `AUTO`.** And its documented behaviour when the
parameter is omitted is, verbatim, *"the model is free to choose whether to use the
specified tools or not"* — which is exactly what `auto` means, and `auto` is the only
value this codebase ever sends.

So `supports_tool_choice: false` is **permanent and correct**, not a temporary
workaround. It expresses the intended semantics in the provider's own vocabulary.

I had speculated the opposite: that the model does support the parameter and litellm was
merely missing metadata for it, so the flag could be dropped once the catalog caught up.
That was an inference from the absence of a catalog entry, and the absence meant litellm
had *no information*, not that the capability was there.

litellm refusing the parameter is arguably the right call for this model. Only its
message misleads — `UnsupportedParamsError: azure_ai does not support parameters:
['tool_choice']` reads as "no tool-calling capability" when the truth is "that value is
not in Cohere's enum".

### Retracting the proxy-configuration advice

Earlier in this thread I suggested `drop_params: true` on the gateway as the fix, and
later floated `allowed_openai_params: ['tool_choice']` and a `model_info:
supports_tool_choice: true` override. **Do not do any of them.**

| Suggestion | Why it is wrong |
|---|---|
| `allowed_openai_params: ['tool_choice']` | pushes the non-existent `auto` through to Cohere — trades a clear client-side error for a provider-side one |
| `model_info: supports_tool_choice: true` | same, by telling litellm a capability claim that is false for the value we send |
| `drop_params: true` gateway-wide | silently discards unsupported parameters for **every** model and every consumer of that gateway — the exact class of silent failure these integration rules exist to prevent |

**No proxy change is needed for `tool_choice`.** The preset handles it, scoped to this
model, version-controlled with the evidence, with no cross-team dependency.

### The one thing that does need infra

The **403 from the Azure Application Gateway**. GitHub-hosted runners cannot reach the
gateway, and that is not fixable from our side. It is the only remaining blocker for
running this model in CI, and it is worth the infra ask that `tool_choice` is not.

### Left open honestly

Whether Azure's OpenAI-compatible layer would itself translate `auto` into Cohere's
native form. I found no documentation either way, and could not test it — the gateway is
returning 403 from this network right now, so the request cannot leave. It does not
change the decision (omission is correct either way), so it is recorded rather than
guessed at.

Comments only in `5379480` — preset behaviour, the certificate and every measurement are
unchanged.

PR #802, opened 2026-08-03
fix(test): add mkdir and write config before run orchestrator

### LINKED ISSUE #707: fix(tests): the run_trial/orchestrator parity test burns a paid trial and then dies on a missing directory
## Context

Found during #676 Stage 3, while using `tests/integration/test_run_trial_e2e.py` to produce real-provider evidence.

`_run_orchestrator_baseline` writes its run config to `output_dir.parent / "run.yaml"` before anything creates that parent directory, so the test dies with `FileNotFoundError: …/orch/run.yaml`. The fix is one `mkdir(parents=True)`.

**Pre-existing — it fails on `main` too**, and it has presumably never passed.

## Why this is worse than an ordinary broken test

The failure lands **after** the expensive half has already run. The `run_trial` leg completes first — a real agent loop against a real provider — and *then* the orchestrator leg fails on plumbing. So the test spends live provider budget on every invocation and can never report a result.

It is also a **parity lock that has never locked anything**: its purpose is to assert that `run_trial` and the orchestrator produce equivalent outcomes for the same task. Nobody can currently rely on that guarantee, and because the failure looks like an environment problem rather than a broken assertion, it reads as flaky rather than dead.

Related in kind to #700 and #702 — a check whose presence implies coverage it does not provide. Notably it would not have been caught by CI either way, since `test-gate` never runs (#700).

## Deliverables

- [ ] The parity test runs to completion and asserts the parity it was written for
- [ ] The cheap plumbing is validated **before** the expensive leg, so a setup error cannot cost provider budget
- [ ] If the test is not meant to run routinely, its cost is documented where a contributor will see it before invoking it

## Acceptance Criteria

- The test passes, or fails on a genuine parity difference with a message naming it
- A configuration or filesystem error surfaces before any provider call
- Running it twice costs what one run should cost

## Design Decisions

- **Order the fixture so cheap setup precedes paid work.** The one-line `mkdir` fixes this instance; putting config assembly before the first provider call is what stops the next instance costing money to discover.
- **Do not simply skip it.** It locks a real property — that the two entry points agree — and that property is worth having, especially with `run_trial` used as an evidence-gathering tool during this milestone.

## Out of Scope

- Broadening what the parity test compares
- The rag-service fixture timeout (#697), a separate cause in the same tier

## Dependencies

- Discovered during #676 Stage 3. Not blocking: the `run_trial` half runs green and produced the evidence it was used for.

kostiantynvoiku-web:
Fixes #707: the parity test wrote its baseline config to a directory that
didn't exist yet — and only after the paid `run_trial` leg had finished, so
every attempt burned a real trial and then died on `FileNotFoundError`.

- Extracted `_write_orchestrator_config` and added the missing
  `mkdir(parents=True, exist_ok=True)`.
- The test writes the config **before** the paid leg, so setup errors now
  surface while the run is still free.

Test-only change; the gated e2e runs in the push/nightly lane.

Closes #707

### DIFF
--- a/tests/integration/test_run_trial_e2e.py
+++ b/tests/integration/test_run_trial_e2e.py
@@ -87,14 +87,9 @@ def _models(provider: str, model: str) -> dict[str, dict[str, Any]]:
     }
 
 
-def _run_orchestrator_baseline(
-    provider: str, model: str, output_dir: Path, env: dict[str, str]
-) -> dict[str, Any]:
-    """Drive ``tolokaforge run`` for the same task and read the persisted trial.
-
-    Returns ``{"status", "binary_pass", "score", "env"}`` from the on-disk
-    ``trajectory.yaml`` / ``grade.yaml`` / ``env.yaml``.
-    """
+def _write_orchestrator_config(provider: str, model: str, output_dir: Path) -> Path:
+    """Write the baseline run config: cheap setup, kept ahead of the paid legs
+    so a configuration or filesystem error cannot cost provider budget."""
     cfg = {
         "models": _models(provider, model),
         "orchestrator": {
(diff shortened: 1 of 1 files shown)


CiroGamboa (approved):
LGTM

PR #852, opened 2026-08-04
fix(automation): gate self-referential capability claims and untested classes

balazs-toloka:
## Why

The auto-resolve fix-loop's only success signal is "the reprobe went green". For most probes that is a real provider measurement. For the two replay probes it is not: `test_thinking_replay_roundtrip` and `test_unsigned_thinking_replay` fire turn 1 live, then **mock the provider turn** and assert only our own outgoing request body. A codec that emits the expected shape satisfies them **by construction**, so the loop cannot distinguish "I fixed the model's behaviour" from "I wrote the payload I set out to write".

`deepseek/deepseek-v4-flash-0731` (#846) walked straight into that gap. It shipped a new reasoning codec, took `unsigned_thinking_replay` from 0/15 to 5/5, and certified the capability `required`. A live A/B afterwards, on the same conversation with and without the replay payload:

| | turn-2 `prompt_tokens` |
|---|---|
| with `reasoning_details` on the wire (2.8 kB, ~668 reasoning tokens) | 523 |
| without | 523 |

The route accepts the field and silently ignores it. Nothing in the pipeline could have caught that, and the prose rule that came closest lost to the competing "OR the reprobe shows it green" clause in the same paragraph.

## What this adds

**`cert_reconcile` was asymmetric by construction.** It guarded against *under*-crediting a model (false-pessimism, undeclared, core-unsupported) with no check on the over-crediting side.

- **SELF-REFERENTIAL** (violation) — a payload-only capability declared `required` against a failing **native** baseline. The policy may still ship; only the cert claim is refused.
- **UNBACKED** (warning) — `required` with no probe result at all.

`PAYLOAD_ONLY_CAPABILITIES` is drift-guarded against a *structural* fact rather than prose: the set must equal the probes that patch `tolokaforge.core.llm.client.completion` — the only transport seam, and the only patch target, in the live capability suite.

**New `automation.classgate`** (finalize gate, `automation check-new-classes`): every newly registered policy class must be referenced by the overlay **and** covered by a unit test under `tests/unit/llm/`. `docs/ADD_NEW_MODEL.md` step 5 already mandated the test; nothing enforced it, because the fix-loop's success signal never looks there.

## Prompt fixes for the two upstream causes

1. The payload-only rule is now stated at **probe level** and names the deterministic gate, instead of generalising about mechanisms. The old phrasing was both **evadable** (write a codec that is no longer "summary-only") and **wrong** (`THINKING_EMITS_BLOCKS` does pass live on such a codec).
2. The "grep before you write a class" rule listed only `response_policy.py` and `schema_sanitizer.py` — which is exactly why `reasoning_codec.py` went unsearched. It now covers every module behind `_POLICY_REGISTRIES` plus `model_presets.yaml`.

That second omission cost a redundant class. The shipped `gemini` codec already satisfies unsigned replay on that route (verified live), and `qwen3.8-max` had solved the **identical** fix-target days earlier with a one-line `reasoning_codec: gemini` swap. The new class also round-tripped *less* faithfully: it rebuilds the envelope from the flat summary, dropping the provider's own `format` and `index` that the shipped codec preserves.

## Verification

- `tools/automation` suite: **267 passed** (28 in the two touched files, incl. the real #846 and mimo baselines as fixtures).
- End-to-end: staging the real #846 registry line makes `classgate` fail with `UNTESTED-CLASS` + `UNREFERENCED-CLASS`.
- `ruff check` + `ruff format --check` clean; workflow YAML parses.
- Both new gates are additive and fail closed to `needs-human`; no existing integration path changes when a candidate adds no class and claims no payload-only capability.

## Note

This does not attempt to fix #846 itself. Given the finding above, the better resolution there is to drop the new codec and use `reasoning_codec: gemini`; that is a separate change.

### DIFF
--- a/.github/workflows/integrate-model.yml
+++ b/.github/workflows/integrate-model.yml
@@ -702,6 +702,14 @@ jobs:
             echo "::error::cert does not reconcile with the observe baseline - refusing to commit"
             exit 1
           fi
+          # Any NEWLY registered policy class must be referenced by the overlay and covered by
+          # a unit test (docs/ADD_NEW_MODEL.md step 5). The fix-loop's success signal is the live
+          # reprobe, so a missing unit test is otherwise invisible. set -e safe; exit 1 ->
+          # needs-human. Reads the STAGED presets.py diff, so it must run after `git add`.
+          if ! uv run automation check-new-classes --overlay observation/resolve/overlay.yaml; then
+            echo "::error::a new policy class is untested or unreferenced - refusing to commit"
+            exit 1
+          fi
           # Verify EXACTLY the staged tree (what we are about to commit), not the workspace.
           # `stash --keep-index` drops everything not staged so the worktree == the future
           # commit; then gate on: (1) it imports; (2) no-regression - the resolved policy must
--- a/.github/workflows/integrate-model.yml
+++ b/.github/workflows/integrate-model.yml
@@ -389,11 +389,16 @@ jobs:
           PY
 
       - name: Aggregate deterministic findings (report-only)
+        id: findings
         run: |
           uv run automation observe-findings observation \
             --out observation/findings.json \
             --summary-out observation/summary.md \
             --run-url "$RUN_URL"
+          # Hash the evidence BEFORE any agent runs. Step outputs live outside the
+          # workspace, so the finalize gate can require the findings it judges to be
+          # byte-identical to what observe actually measured.
+          echo "sha=$(sha256sum observation/findings.json | cut -d' ' -f1)" >> "$GITHUB_OUTPUT"
 
       - name: Upload observation bundle (artifact for debugging / re-runs)
         uses: actions/upload-artifact@v4
@@ -585,6 +590,13 @@ jobs:
               claude -p "$(cat /tmp/compose_$i.md)" --model "$AGENT_MODEL" \
               --allowedTools "Bash,Read,Edit,Write" --dangerously-skip-permissions \
               --max-turns "${RESOLVE_MAX_TURNS}" --verbose || true
+            # The compose agent just ran with unrestricted Bash in this workspace; the loop's
+            # own verdict tooling (reprobe / greencheck under tools/automation) must be
+            # pristine when it judges the attempt - an edit there would otherwise run HERE
+            # and then be stash-popped back at finalize without ever appearing in the PR.
+            # Restores index AND worktree from HEAD; committed (PR-visible) changes remain
+            # the human review's job.
+            git restore --source=HEAD --staged --worktree tools/automation
             # Agent-requested escalation: if it lacks the data to produce a CORRECT policy (a
             # data-bound quirk whose scope needs real-domain evidence the observe never surfaced),
             # it sets `needs_human: true` in decision.json. Break immediately - no wasted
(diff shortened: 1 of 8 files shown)


bberkes-toloka:
Applied the adversarial-review findings in three commits (`d50af55`, `ee819e2`, `d750856`). Finding-by-finding:

| # | Finding | Fix |
|---|---|---|
| 1 | The gate-demanded unit test never got committed (worktree-read gate, index-based commit, `tests/unit/llm/` unstaged) | `git add` now stages `tests/unit/llm/`, and the gate reads test sources from the INDEX (`git ls-files` + `git show :path`) - the test it credits is the test that ships |
| 2 | `classgate` failed open on git failure and on any CWD but the repo root | Root anchored to `__file__` (`parents[4]`, like `cert._load_cert`), git runs with `cwd=root`, any nonzero exit fails the gate; `TestRun` now runs against throwaway git repos, incl. the non-git fail-loud case and CWD independence |
| 3 | The binding regex failed open on single quotes, missing commas, wrapped values, dotted/instance/factory bindings | Diff-text regex replaced by an AST set-diff of the staged vs HEAD registry (dict literals of `_POLICY_REGISTRIES` slots, subscript assigns, `.update()`, inline slot dicts); every listed evasion shape is a parametrized test |
| 4 | Moved/reordered registry lines reported as new (false-positive hard fail) | Same AST set-diff: a move is by construction not a new `(registry, key, value)` triple; pinned by test |
| 5 | Substring matching under-enforced both checks | Word-boundary matching for class names (`MinimaxM3TagRecoveryResponse` no longer vouches for `TagRecoveryResponse`); keys count only in a YAML value position (`qwen` is not satisfied by `match: ["qwen/qwen3.8-max"]`) |
| 6 | Drift guard keyed on one literal patch string | Guard now matches the seam in all three spellings (dotted string, `patch.object(client, "completion")`, `setattr(client, "completion", ...)`) |
| 7 | Gates read agent-writable inputs; the adversary runs first in the same workspace | The loop re-pins `tools/automation` to HEAD (index + worktree) after every compose invocation, finalize does the same before its gates, and finalize refuses to run against a `findings.json` that no longer hash-matches the value captured at aggregation time (step output, outside the workspace). Committed, PR-visible changes remain the review's job |
| 8 | Docs missing the new gates | `docs/AUTO_INTEGRATION.md` now enumerates reconcile-cert (incl. SELF-REFERENTIAL / UNBACKED), `check-new-classes` with staged-tree semantics, and the tamper hardening |
| 9 | Docstring/message claimed a preset-entry check that was not implemented | Implemented for real: the reference check reads the staged `model_presets.yaml` too (`git show :path`), and the messages state exactly what is checked |

Also applied the first design note: a payload-only capability `required` with NO probe result is now a hard SELF-REFERENTIAL violation (it could previously slip through as the soft UNBACKED warning when every repeat's junit was unparseable), and the UNTESTED-CLASS message no longer overstates the check ("prefer a captured real-response fixture" instead of claiming one is verified).

Not done, deliberately: the two remaining design notes (policing registry-key REBINDS beyond the new set-diff visibility - a rebind now at least surfaces as a new binding; and executing the mentioned unit test instead of name-matching it) are follow-up-sized and left out of this PR.

Verification: 290 automation tests pass (was 271), `ruff check` clean, black clean, actionlint clean (pre-commit on all changed files), workflow YAML parses. Note for merge sequencing: the workflow hunks (finalize step, resolve loop) overlap textually with open PR #882 on `improve/auto-integration-reliability`; whichever merges second resolves a trivial conflict in `integrate-model.yml`.

PR #864, opened 2026-08-04
fix(docker): take the runner build context from the builder in core_stack (v0.14.1 still broken)

balazs-toloka:
**v0.14.1 still cannot run a Docker-runtime eval.** #858 fixed the wrong copy of the list.

## What #858 missed

#858 taught `builder._runner_definition()` to resolve the runner build context from packaged copies on a wheel install. But `stacks/core.py` kept its **own duplicate** of the repo-relative list, and that is the path a real run takes:

```
orchestrator.run -> service_stack.start_all -> build_images
  -> assemble_build_context(repo_root(), svc.dockerfile, svc.context_files)
```

`svc` comes from `core_stack()`, not from `get_image_definition()`. So v0.14.1 fails identically:

```
FileNotFoundError: Declared context path not found: pyproject.toml
  (resolved to .../site-packages/pyproject.toml)
```

That message is the **relative-path branch** of the assembler, which is the proof the factory was never consulted.

My own verification of #858 called `get_image_definition("runner")` directly — a code path production does not use. That is why it looked fixed.

## Fix

`core_stack()` now reads `get_image_definition("runner")["context_files"]` instead of duplicating the list, so the wheel-install branch actually reaches production and the two lists cannot drift apart again. A new test locks the equality.

`ServiceDefinition.context_files` widens to `list[str | tuple[str, str]]` — the resolved wheel-install form carries one `(source, destination)` pair (`_python_version.txt` -> `.python-version`), and a Pydantic `list[str]` would have rejected or stringified it.

## Verification — through the real entry point this time

From a wheel built off this branch and installed into a clean venv (`repo_root()` is `site-packages`, i.e. the failing condition):

1. `core_stack().services["runner"].context_files` -> **6/6 entries resolve**, each landing under the name the Dockerfile `COPY` expects
2. `assemble_build_context(...)` -> context carries `pyproject.toml`, `README.md`, `LICENSE`, `.python-version`, `scripts/hatch`, `tolokaforge`
3. **`stack.build_images()` builds both the db-service and runner images** — the exact call in the failing traceback

Tests: `black --check` clean (747 files), `ruff` clean, **45 passed** across the docker, build-context and runner-subset suites, including the new no-duplication test.

## Supersedes #862

Folds in that black formatting fix. `main`'s lint has been red since #858 because I formatted it with `ruff format`; this repo formats with **black** (pre-commit runs ruff for *linting only*). Red lint also skips `test-full` / `test-gate` / `test-smoke`, so landing this makes those run again.

Needs a **v0.14.2** to be usable by the arena eval (TECHDEL-546).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,67 @@ jobs:
       run: uv run pre-commit run --all-files --show-diff-on-failure
 
   # ============================================================================
+  # Wheel-install gate — the ONE thing every other job is blind to
+  # ============================================================================
+  # Every other job `uv sync`s a source checkout, so `builder.repo_root()` is the
+  # repo and the runner image's build context always resolves. Consumers install
+  # the engine as a wheel, where `repo_root()` is `site-packages` — and that path
+  # shipped broken TWICE (v0.14.0 and v0.14.1): the runner context resolved to
+  # non-existent repo-root paths and every Docker-runtime run died in
+  # `build_images()` before a single trial. This job installs the built wheel into
+  # a clean venv and drives the real production entry point. No Docker needed —
+  # assembling the context is where the bug lives.
+  wheel-install-gate:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+    - uses: actions/checkout@v5
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@v7
+      with:
+        version: "0.9.5"
+        enable-cache: true
+        cache-dependency-glob: "uv.lock"
+
+    - name: Set up Python (pinned in .python-version)
+      run: uv python install "$(cat .python-version)"
+
+    - name: Build the wheel
+      run: uv build --wheel
+
+    - name: Install the wheel into a clean venv (no source tree on the path)
+      run: |
+        uv venv /tmp/wheelenv
+        VIRTUAL_ENV=/tmp/wheelenv uv pip install dist/*.whl
+
+    - name: Runner build context must assemble from the installed wheel
+      # cd / so the source checkout cannot be picked up via CWD.
+      run: |
+        cd /
+        /tmp/wheelenv/bin/python - <<'PY'
+        from pathlib import Path
+        from tolokaforge.docker.builder import assemble_build_context, repo_root
+        from tolokaforge.docker.stacks.core import core_stack
+
+        root = repo_root()
+        assert not (root / "pyproject.toml").is_file(), (
+            f"expected a wheel install, but {root} looks like a source checkout"
+        )
+
+        svc = core_stack().services["runner"]
+        build_dir = assemble_build_context(root, svc.dockerfile, svc.context_files)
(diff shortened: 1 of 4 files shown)


bberkes-toloka:
## Adversarial review round — both required fixes applied

An independent review pass confirmed the production fix and **found that the test I added could not fail**. Fixed in `3849d9b`.

### 1. The test was tautological (proven, then fixed)

The reviewer reverted only `core.py` and the test still passed. Cause: in a source checkout the builder returns the *same six strings in the same order* as the literal the bug reintroduces, so the equality assert held either way — and it also compared `core_stack()`'s value against a fresh call of the very function that produced it.

It now runs **inside** the wheel-install simulation (the existing `repo_root` / `installed_package_dir` monkeypatch fixture), calls `core_stack()` under the patch, and asserts `assemble_build_context` actually succeeds. Proven to gate:

| state of `core.py` | result |
|---|---|
| reverted to `fdd01f5` (bug restored) | **FAIL** — `Declared context path not found: pyproject.toml` |
| this PR | PASS |
| whole file | 15 passed |

The equality assert stays as a drift check, documented as necessary-but-not-sufficient.

### 2. No CI job exercised the wheel-install path at all

`lint`, `test-smoke`, `test-full` and `test-gate` all `uv sync` a source checkout, so `repo_root()` is always the repo and the runner context always resolves. That is exactly why this bug class shipped **twice**. Added a `wheel-install-gate` job: builds the wheel, installs it into a clean venv, and drives `core_stack()` + `assemble_build_context` from `CWD=/` so the source tree cannot leak in. No Docker needed — assembling the context is where the bug lives.

### Review findings I did NOT act on

- `tests/unit/test_docker_build_context.py:238,244` use `set(context_files)` / `entry.endswith(".whl")`, which would `AttributeError` on a tuple entry. Not reachable today (those tests use the source-checkout definition); left alone rather than change passing tests' semantics.
- **`tolokaforge/docker/stacks/test.py:51` is a third runner `ServiceDefinition`** with no `context_files` and `context="."`, verified broken on an installed engine. Dev-only (`dx docker up --profile test`) and pre-existing — worth a separate issue, out of scope here.
- `expected_image_ref("runner")` disagrees with the stack's actual tag because `builder` passes `PYTHON_VERSION` and `core.py`'s `runner_build_args` starts empty. Pre-existing, no production consumer.

### Independent confirmations worth recording

The reviewer reproduced the **original failure on v0.14.1** (wheel-installed into a throwaway venv → the verbatim `FileNotFoundError`), so the diagnosis is confirmed rather than assumed. They also built the runner image from **both** the wheel-install and source-checkout contexts and got the **same image ID** `sha256:8e6e6d7e7c7f…`, and confirmed the in-container hatch build produces an **identical 93-file subset wheel** from either context despite the packaged context carrying `__pycache__` dirs. The built container serves gRPC on `:50051` with a green healthcheck.

Still needs a **v0.14.2** to unblock the arena eval (TECHDEL-546) — this branch is still `version = "0.14.1"`.

PR #890, opened 2026-08-04
feat(grading): deterministic trace checks, milestone 28

azorej:
# Milestone 28: Deterministic trace checks

Milestone: https://github.com/Toloka/tolokaforge/milestone/28
Umbrella / design authority: #684

## TL;DR

This milestone adds `trace_checks` — a fourth grading component that grades tool-call trajectories with declarative matchers and a closed set of temporal, counting and boolean constraints — and it does **not** change how any existing pack is graded except where that pack was already being graded wrongly. The shape of the change is a new subsystem (one substrate-neutral evaluator both grading substrates call, plus an authoring gate, a zero-spend replay command, and rubric-migration tooling) sitting on top of a rebuilt foundation: a machine-readable substrate-parity manifest for every author-facing `grading.yaml` key, a complete grade-time trajectory record joined by the provider's `call_id`, one composition rule for `state_checks`, and one combine algebra. The correctness half of that foundation shipped separately to `main` as PR #748 because it fixed numbers `main` was producing wrong every day; the feature half is this branch. Compatibility costs are real but bounded and each is named in the CHANGELOG: the runner-engine version lock now bites on every pack (rebuild the image *before* rolling the engine), `tolokaforge validate` exits non-zero for the first time, packs declaring removed keys stop loading with the replacement named, metrics rates are now over *measured* trials, and four shipped example packs' scores are explicitly not comparable across the change. Nothing is retired from any LLM judge rubric: one criterion is narrowed against recorded verdicts with its veto preserved, and the rest ship as measured candidacy declarations.

Closes #677, closes #678, closes #679, closes #680, closes #681, closes #682, closes #683, closes #696, closes #714, closes #720, closes #726, closes #733, closes #744, closes #758, closes #773, closes #784.

(Already on `main` via PR #748, the milestone's correctness half: #675, #676, #686, #692, #689, #705 — plus #218, #716, #736, #523 outside the milestone. #746 closed as part of the combine work.)

## Impact on existing tasks — read this first

- **Today, with no `grading.yaml` changes:** every pack that grades correctly today keeps its verdict. No grading snapshot moved across any stage of any issue on this branch — that is the load-bearing number for work that rewrites how two substrates fold component scores. `trace_checks`, `alternatives`, `severity`, `bind`, `min_assistant_turns` and `migration.yaml` are all additive and default to absent.
[...]

PR #905, opened 2026-08-06
fix(llm): user simulator restarts the conversation after the agent answers (CBT-021)

azorej:
## Bug

In CBT-021 runs, the LLM user simulator re-sent its scripted opening **verbatim** after the agent had fully answered it, restarting the conversation. The agent re-ran the whole flow, created a duplicate Zendesk ticket, and the trial failed on `state_checks: 0.0`. The restart always hit exactly the second user turn; the third behaved normally.

## Root cause

`UserSimulator._llm_reply` builds the simulator's LLM view by role-flipping the shared transcript (its own past USER turns → `assistant`, agent turns → `user`). The trial's seeded opening flips to `assistant` at index 0, and to satisfy the providers' leading-user-role constraint the code **dropped it**:

```python
if sim_context and sim_context[0].role == MessageRole.ASSISTANT:
    sim_context = sim_context[1:]
```

So on the simulator's first live call, its entire history was just the agent's answer — no trace of its own request. Its instructions say to open with the scripted line, its context showed it had said nothing, so it re-sent the opening. Reproduced deterministically and against the run's real user model (claude-sonnet-4.6 re-emits the opening character-for-character when given the truncated context). #619 previously added a prompt rule against restarting — it can't work: no instruction makes the model remember a turn absent from its context.

## Fix

- Keep the flipped opening; lead the context with the synthetic agent-side greeting `SIMULATOR_GREETING` (the same constant the runner already dispatches at bootstrap — now shared, not duplicated) so the provider constraint is met without deleting the simulator's memory of having asked.
- Skip dialogue-free turns (agent tool-call turns) instead of replaying them as empty user messages.
- Fail loud if the flipped context cannot end on a user-role turn (agent's last turn carried no dialogue text): a trailing assistant-role message is a prefill the provider would continue — silently reproducing the restart symptom via a different route. The `RuntimeError` classifies as a counted `ERROR` trial; it is unreachable on the #619 deferred-stop terminating turn (the pending-stop short-circuit returns before `actor.reply`).
- Bump `simulator_schema_version` 1→2 so analytics can gate cross-run comparisons of user-sim behaviour; `docs/OUTPUT_FORMAT.md` updated and the simulator context contract documented in `docs/LLM_LAYER.md`.

## End-to-end validation with real providers

A/B on the real CBT-021 task (toloka-partners/tolokaforge-tasks `dev/ots`, frozen_mcp_core adapter, Docker stack) with the original failing lineup — agent `openai/gpt-5.6-sol`, user/judge `anthropic/claude-sonnet-4.6`:

| | Baseline (engine v0.11.0 unpatched) | v0.11.0 + this fix backported |
|---|---|---|
| Trials passed | 1/2 | **2/2** (score 1.0) |
| Verbatim restart | trial 1: yes | none |
| Zendesk tickets | trial 1: 2 (duplicate) | 1 per trial |
| `state_checks` | trial 1: 0.0 | 1.0 |

The `_llm_reply` logic is unchanged between v0.11 and this branch, so the validation transfers directly.

## Review

Two subagent review rounds (design critic + implementation reviewer, then a fresh full-diff reviewer) applied in commits 2 and 3: trailing-prefill hole closed, greeting literal deduplicated with the runner, docs scrubbed of past-state phrasing, task-pack content removed from tests, stale Nova docstring updated, duplicate test lock removed. Verdict on actor-model genericity: the change is correctly actor-local (concrete `UserSimulator` only; `Actor`/`TurnPolicy` seams untouched), and the greeting is generic harness filler already imposed at bootstrap.

## Tests

- `tests/unit/test_user_simulator_context.py` — deterministic locks: opening survives, greeting leads, dialogue-free turns skipped, trailing-role invariant raises, shared transcript never mutated.
- `tests/integration/test_user_simulator_live.py` — the failure shape against a real Sonnet user model (one call per integration run; documented in `tests/README.md`).
- Full unit + canonical suites green (6,855 passed).

### Note on `tests/data/run_trial_capstone_golden.jsonl`

The golden was replaced with genuine generator output (`FixtureConductor` → `marshal_result`, byte-verified in-process). This picked up **two** deltas: the `simulator_schema_version: 2` stamp from this PR, and `grade.trace_checks_summary` from #890 which the golden never received — it is stale on `main` too, masked because `test_external_harness_e2e`'s fixture errors (`[Errno 9] Bad file descriptor` in `_reject_ungradeable_packs`) before reaching the byte compare, on `main` CI as well. That fixture failure is pre-existing and deserves its own issue.


CiroGamboa:
Code review — 6 findings, none blocking. Ranked most→least severe.

**1. Empty-content skip can regress the restart bug on empty seeded opening — `tolokaforge/core/llm/client.py:2281`**
If `_bootstrap_via_simulator` returns `text=""` with no tool_calls, `if result.text:` skips the sanitizer and the runner appends `Message(role=USER, content="")` to `self.messages`. On the next `_llm_reply`, the empty opening is dropped by the new content filter, `sim_context` starts with the flipped `USER(agent_response)`, the leading-`ASSISTANT` guard no-ops, no `SIMULATOR_GREETING` is prepended — and the simulator sees no history of ever having asked. Same restart symptom this PR fixes. Suggest either rejecting empty `first_user_text` in `_bootstrap_via_simulator` / `_seed_first_user_message`, or making the trailing-invariant + greeting logic robust to an empty first USER.

**2. Empty-content skip can produce consecutive USER turns → 400 from Anthropic/OpenAI — `tolokaforge/core/llm/client.py:2281`**
`_sanitize_user_text` returns whitespace-only text unchanged (`if not text.strip(): return text` → `"  "` stays `"  "`). The runner stores that as `USER("  ")`. In `_llm_reply` the whitespace USER is dropped by the new filter while the surrounding ASSISTANTs flip to USER, yielding e.g. `[ASSISTANT(open), USER(agent1), USER(agent2)]` — strict-alternation break. Previously the whitespace USER survived (flipped to ASSISTANT) and alternation held. Simplest fix: make `_sanitize_user_text` substitute a placeholder for whitespace-only text too, or drop-through-both-sides in `_llm_reply`.

**3. `last_system_prompt` stamped before the trailing-invariant raise — `tolokaforge/core/llm/client.py:2272`**
Line 2272 assigns unconditionally; the raise at 2310 can fire before `llm_client.generate` is called. `runner.py:406-408` still pulls `last_system_prompt` into `_user_system_prompt_captured`, so an ERROR trial's `prompts.yaml` names a system prompt that was never dispatched — misleading forensic evidence on fail-loud triage. Assign after `generate()` returns (or clear on the raise path).

**4. RuntimeError wording — `tolokaforge/core/llm/client.py:2311`**
Message says "the agent's last turn carried no dialogue text", but the check also fires when `sim_context` is empty (no agent turn at all — e.g. `[ASSISTANT(''), TOOL(...)]`). Rephrase to something like "no agent dialogue turn to answer" and include the observed role sequence to help on-call.

**5. OUTPUT_FORMAT.md phrasing — `docs/OUTPUT_FORMAT.md:257`**
"always `2` today" carries over the "always `1`" phrasing from the pre-bump generation and contradicts the monotonic-bump rule right next to it. Drop the "always" framing or restate as "current value: `2`".

**6. Multimodal `content_blocks`-only turns dropped — `tolokaforge/core/llm/client.py:2281` (latent)**
Filter looks only at `content`, so an ASSISTANT turn with `content=""` and non-empty `content_blocks` (image-only reply) is skipped, and the trailing-invariant then false-positive-raises. No current call site produces such a turn (only TOOL messages set `content_blocks` today), so this is latent — worth an inline note if we keep the filter, or an `or msg.content_blocks` in the guard.

Overall the fix is well-scoped and the test coverage is solid; #1 and #2 are the only ones I'd want closed before merge.

azorej:
All review findings addressed in 4eaee04a. Point-by-point:

**1. Empty seeded opening → restart regression path** — Fixed. `_bootstrap_via_simulator` now raises on an empty/whitespace first message instead of seeding a blank USER turn the flipped context would later drop. Locked by `test_empty_bootstrap_first_message_fails_loud`. (The caller-seeded `initial_user_message` path was already guarded: `ConversationalTurnPolicy.bootstrap` routes whitespace to the simulator.)

**2. Consecutive same-role turns after the skip** — Fixed, and more broadly than the reported case: `_llm_reply` now coalesces adjacent same-role turns in the flipped context (joined with a blank line), which closes both the whitespace-reply shape reported here and the pre-existing consecutive-user shape the hygiene review flagged (agent interim narration + final answer around a tool turn — that one predated this PR, since the old empty separator turns never made alternation legal, just padded it). Locked by `test_adjacent_same_role_turns_coalesce` and `test_whitespace_user_turn_dropped_and_alternation_preserved`.

**3. `last_system_prompt` stamped before the guard raise** — Fixed. The stamp now happens only for a dispatched request, after the trailing-invariant check, so an ERROR trial's `prompts.yaml` never names a prompt that drove no generation. A generation *attempt* that fails downstream still captures the prompt, matching prior forensic semantics. Locked by assertions in both raise tests.

**4. Error wording** — Fixed: "no agent dialogue turn to answer", plus both the flipped-context and shared-transcript role sequences in the message.

**5. "always `2` today"** — Fixed: "current value: `2`".

**6. `content_blocks`-only turns (latent)** — Inline note added at the skip site stating the gap and that no USER/ASSISTANT call site produces such a turn today (the flip has never carried blocks). Left as a note rather than an `or msg.content_blocks` guard: including a blocks-only turn would still emit an empty-text message, since the flip drops blocks — handling multimodal turns properly is its own change.

Also from the hygiene review's nits: the double key-gate in the live test is intentional (clearer skip message), and the PR body's stale "renamed and strengthened" phrasing was already corrected in the current description.

Full unit+canonical suites green (6,858). CI note: the current `test-smoke` failures are the ongoing GitHub Actions incident (jobs die in *Set up job* resolving action downloads — githubstatus.com shows Actions degraded); will re-run once the incident clears.

azorej:
Third review round (fresh subagent, focus on 4eaee04a): no blockers; verdict APPROVE WITH CHANGES. Applied in the latest commit: the live test's claim now matches what its assertions can actually catch (verbatim or re-introducing restart — paraphrase detection needs a grader, not a substring) and it additionally rejects any reply re-stating the customer number; LLM_LAYER invariants now state the code's real conditions (assistant-first flipped context, including bootstrapped openings; the no-dialogue-at-all branch); pre-fix narration present-tensed; dead None-guard dropped. Full unit+canonical suites green (6,858).

Issue #645, opened 2026-07-28
[design] Decouple model integration from the engine


bberkes-toloka:
## Problem

Integrating a model into the engine, or adjusting how an already-integrated
model is driven, requires cutting an engine release even when the change
contains **no engine code**.

## The two kinds of model change

Model integration work falls into two buckets:

**Bucket A: data and certification.** Preset routing that composes *existing*
policy classes (schema sanitizer, prompt/content/response policy, reasoning
codec, cache policy), a price-table entry, and a capability certificate
asserting what the model actually supports on the wire. All of this is
declarative. None of it changes engine behaviour for any other model.

**Bucket B: engine code.** A genuinely new policy, codec or sanitizer class,
because no existing class expresses the model's wire contract.

Bucket B belongs in the engine and should ride an engine release. Bucket A is
the common case, and it should not. Today both do.

## Cost of the current coupling

- **Release cadence is driven by model onboarding rather than by engine
  change.** Releases carry no engine delta, which makes the version line a poor
  signal of what actually changed.
- **Version drift downstream.** Consumers that pin the engine by tag end up
  spread across engine versions purely as a side effect of when each model was
  onboarded. For any consumer that compares results across models, "which
  engine version did this run on" becomes a per-model question instead of a
  constant, and every difference then needs its own justification.
- **Often two releases per model**, as ADR 0002 already records: integration
  tests pass, release, a live run surfaces a model-specific failure pattern,
  the fix is to re-route to a policy class *already in tree*, release again.
  Both releases ship zero engine code.

## What already exists

This is not a from-scratch request. The seam is half open:

- **Preset overlay.** `--presets-file` / `RunConfig.engine.presets_file`, loaded
  and validated at `tolokaforge/core/llm/presets.py:176`, failing loud on
  unknown policy names. Documented in `docs/ADD_NEW_MODEL.md` under "Adding a
  model without an engine release".
- **Pricing overlay.** `observability.pricing_overlay_path`, applied at
  `tolokaforge/dx/cli/main.py:655` via `pricing.reload_pricing()`.
- **Report-side pricing override.** The results collector accepts an explicit
  pricing file instead of the bundled table.
- **ADR 0002** (`docs/adr/0002-external-model-registry.md`) diagnosed exactly
  this problem and chose the overlay as the mechanism.

So the run path can already consume an out-of-tree model registration. In
practice it is not used, and the reasons are structural.

## What still couples model integration to the engine

[...]

PR #908, opened 2026-08-06
docs(adr): ADR-0030 tolokaforge-models split (widened) + promote ADR-0002

CiroGamboa:
## Summary

- Proposes **ADR-0030 — Model data as a second PyPI wheel `tolokaforge-models` from the same monorepo**. Design PR only — zero code changes.
- **Widened 2026-08-07** per [@bberkes-toloka's review comment](https://github.com/Toloka/tolokaforge/issues/645#issuecomment-5213977204) — policy / codec / sanitizer *classes* are now in scope (selectively), not deferred. Rationale: 3 of 5 recent auto-integrations needed new engine classes, and ADR-0002's own deferral condition ("policy classes recurring per model") is now met on the evidence.
- Promotes **ADR-0002** from `Proposed` → `Accepted` with an "advanced by ADR-0030" back-link.
- Updates `docs/adr/README.md` index.

## Why

Between v0.8.4 → v0.11.2 (2026-07-15 → 2026-07-27, 12 days), 2 of 9 releases were pure zero-engine model shims — v0.9.1 and v0.9.3, both with no `### Feat` / `### Fix` sub-sections in `CHANGELOG.md` (bare version headers). And per bberkes's evidence review: 3 of 5 auto-integrations since 2026-07-17 needed *new engine classes* — `gemini-3.5-flash` (`response_policy.py`, `schema_sanitizer.py`), `inkling` (`prompt_policy.py`), `deepseek-v4-flash` (`reasoning_codec.py`). The policy modules changed 5 times since the OSS cleanup, every one driven by model onboarding.

The overlay seam ADR-0002 shipped opened the *run-time* half of the decoupling. This ADR closes the *release-time* half — across the widened Bucket A.

## What the widened ADR states (short version)

- **Single publisher, single package** for `tolokaforge-models`. Not a plugin ecosystem. Deliberately narrower than ADR-0002's own Option 3.
- **Independent versioning + `__api_version__` integer + `minimum_engine_version`** as install-time guardrails.
- **12 requirements from bberkes's review** stated verbatim as design principles.
- **Bucket A widens:** data + per-model subclasses of stable engine bases + provider bindings + cert exclusion data + per-model cert bodies + vendor-inference table.
- **Bucket B narrows:** only genuinely new base classes / new lifecycle stages / new capability categories / base-shape changes remain engine work.
- **Two new policy slots** (`assistant_text_policy`, `params_policy`) so Cohere marker handling stops living as an `if target == "anthropic"` branch and generation-param knobs stop requiring signature changes.
- **Slot values become `{name, params}`.** Policies are configured, not just selected.
- **Provider bindings become data** (`providers.yaml`): endpoint, credential env, rotation, rate-limit patterns, routability, slug rewrite. Nova hardcode + `UNROUTABLE_PROVIDERS` + `_is_rate_limit_exception` regexes + `OPENROUTER_API_KEYS` rotation all move out.
- **Capability enum authority moves to the models package.** Weaker engine authority over "what supported means"; stronger decoupling. Conscious trade-off.
- **Certification harness in `tolokaforge.testing.certify`** — with certificate-borne exclusion data, per-capability probe registration decorator, and per-model test bodies collectable from the models wheel.
- **Forwarding stub for downstream `importlib.resources` consumers** (like `benchmark-results-collector`) for one release cycle.
- **Interface stays one seam** (`tolokaforge.core.model_data`) plus one public certification seam (`tolokaforge.testing.certify`). Future repo split remains cheap.

## Colleague review focus points

Explicitly want reactions on:

1. **Is the § Requirements list complete?** Any known integration pain missed?
2. **Capability enum ownership — models-package-authoritative.** Weaker engine authority vs. cleaner decoupling. Right call?
3. **Base-class hook policy — "add hook + subclass in same PR" landing across two wheels.** Workable? Or two-release protocol needed?
4. **Downstream forwarding stub — one-release deprecation window.** Long enough? Consumers to notify beyond `benchmark-results-collector`?
5. **Any policy classes today that do NOT fit "inherit from stable base, use only public API"?** Those stay Bucket B; would like the current list surfaced.
6. **ADR-0002 Option 4 selective adoption criterion.** YES for subclasses of stable bases + public API only; NO for new bases / new lifecycle stages. Right cut?

## Follow-up implementation PRs (gated on this ADR)

Tracked under [GH #645](https://github.com/Toloka/tolokaforge/issues/645). Consolidatable to 8-9 PRs; landing order chosen so each unblocks the next:

1. Fail-loud entry-point registry semantics.
2. Certification seam extraction + certificate exclusion data + probe registration (load-bearing).
3. `models_fingerprint` in `engine_run_state.json`.
4. New extension slots — `assistant_text_policy` + `params_policy`.
5. Policy configuration — slot values `{name, params}` + nested unknown-key rejection.
6. Provider bindings as data.
7. Public helper promotion.
8. Base-class hooks landing.
9. Create `tolokaforge-models` package + cut over (THE main structural PR).
10. Downstream forwarding stub + deprecation notice.
11. Docs flip.
12. Auto-integration workflow retarget.
13. Acceptance test — replay last N integrations, assert Bucket A candidates ship with zero engine commits.

## Test plan

- [ ] Colleague review — the six focus points above.
- [ ] Explicit reaction from @bberkes-toloka on whether the widening captures the intent of the review comment.
- [ ] ADR file renders correctly (headings, tables, front matter, § Requirements list scannable).
- [ ] README index entry lands cleanly.
- [ ] No implementation PRs open in parallel — implementation begins only after this merges.

### DIFF
--- a/docs/adr/0002-external-model-registry.md
+++ b/docs/adr/0002-external-model-registry.md
@@ -155,7 +155,7 @@ the operator threading the flag manually.
 
 - Related ADRs:
   - [0001 — Record architecture decisions in ADRs](0001-record-architecture-decisions.md).
-  - [0030 — Model data as a second PyPI wheel](0030-tolokaforge-models-split.md) — completes Option 3 of this ADR (packaged model-data artifact) and lands the certification-harness relocation this ADR flagged as required to make the seam usable.
+  - [0030 — Model data as a second PyPI wheel](0030-tolokaforge-models-split.md) — lands the packaged model-data artifact this ADR's § Context anticipated, plus the certification-harness relocation this ADR flagged as required to make the seam usable. Deliberately narrower than Option 3: single Toloka-published wheel, not a multi-publisher plugin registry.
 - Related code:
   - `tolokaforge/core/llm/presets.py` (loader, registries, validator).
   - `tolokaforge/adapters/__init__.py:86` (`ensure_registered_adapter`, the
--- a/docs/adr/0002-external-model-registry.md
+++ b/docs/adr/0002-external-model-registry.md
@@ -1,10 +1,11 @@
 # 0002. External model registry — operator-overridable preset data
 
-- **Status:** Proposed
+- **Status:** Accepted
 - **Date:** 2026-06-17
 - **Deciders:** @cirogam22
 - **Supersedes:** —
 - **Superseded by:** —
+- **Advanced by:** [ADR-0030](0030-tolokaforge-models-split.md) — completes the seam via a packaged, versioned model-data wheel and the certification-harness relocation this ADR deferred.
 
 ## Context and Problem Statement
 
@@ -152,7 +153,9 @@ the operator threading the flag manually.
 
 ## Links
 
-- Related ADRs: [0001 — Record architecture decisions in ADRs](0001-record-architecture-decisions.md).
+- Related ADRs:
+  - [0001 — Record architecture decisions in ADRs](0001-record-architecture-decisions.md).
+  - [0030 — Model data as a second PyPI wheel](0030-tolokaforge-models-split.md) — completes Option 3 of this ADR (packaged model-data artifact) and lands the certification-harness relocation this ADR flagged as required to make the seam usable.
 - Related code:
   - `tolokaforge/core/llm/presets.py` (loader, registries, validator).
   - `tolokaforge/adapters/__init__.py:86` (`ensure_registered_adapter`, the
--- a/docs/adr/0030-tolokaforge-models-split.md
+++ b/docs/adr/0030-tolokaforge-models-split.md
@@ -1,10 +1,12 @@
 # 0030. Model data as a second PyPI wheel — `tolokaforge-models` from the same monorepo
 
 - **Status:** Proposed
-- **Date:** 2026-08-06
+- **Date:** 2026-08-06 (widened 2026-08-07 per [issue-#645 comment](https://github.com/Toloka/tolokaforge/issues/645#issuecomment-5213977204))
 - **Deciders:** @CiroGamboa
 - **Supersedes:** —
 - **Superseded by:** —
+
+> **Widening note (2026-08-07).** The original draft of this ADR (open in PR #908 as of 2026-08-06) put policy, codec, and sanitizer *classes* explicitly out of scope, following ADR-0002's own § Considered Options which deferred its Option 4. A review by @bberkes-toloka against the auto-integration history showed that boundary does not hold: 3 of 5 recent auto-integrations needed new engine classes, and the policy modules changed 5 times since 2026-05-28, every one driven by onboarding a specific model. The ADR is widened in place — new decision-drivers, new § Requirements section, new extension-point surface, provider-bindings-as-data, install-time validation — so PR #908 lands the full picture rather than the first half of it. Sections added by the widening are marked in the text.
 
 ## Context and Problem Statement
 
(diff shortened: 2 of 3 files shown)


bberkes-toloka:
Read the widened version. All twelve requirements are in, and not as a nod: the new extension slots, provider-bindings-as-data, install-time validation and the ADR-0002 Option 4 reversal are the substance of the ask, and § Follow-ups is sequenced so each step unblocks the next. This is a materially better document than the one I reviewed.

**One framing note first, because it shapes all four points below.** The goal I am measuring against is sharper than "fewer engine releases": **one engine version should absorb any number of model integrations.** Onboarding the twentieth model onto engine 0.16 should still leave you on engine 0.16. That is what makes § Context's comparability argument actually hold — "which engine version did this run on" stops being a per-model question — and it is what lets the auto-integration pipeline take a model end to end with no human in the release path. Read that way the useful question is not what fraction of integrations need an engine release, but whether the engine version can stay put across a whole lineup refresh.

Three places where the mechanism under-delivers against the requirement it answers, some smaller fixes, and answers to two of your review questions. Line references are against `origin/main` (`8107796`), this PR's base.

### 1. The headline number, measured against the ADR's own Bucket rule

§ Decision Drivers targets *"~1 integration in 20 needing an engine release, not 3 in 5"*. The Bucket criterion sends base-class hooks to the engine: *"if it needs a not-yet-added base hook ... the hook lands ... in the same engine PR"*. Applied to the seven integrations in § Evidence:

| case | engine code then | after ADR-0030 |
|---|---|---|
| #475 kimi-k3 | none | models |
| #614 claude-opus-5 | none | models |
| #845 qwen3.8-max | none | models |
| #474 muse-spark-1.1 | shared test body | models, once exclusions are certificate data |
| #846 deepseek-v4-flash | clean codec subclass | models |
| #719 inkling | private `_dict_maps` symbol | models, after follow-up 7 promotes the helper |
| **#560 gemini-3.5-flash** | **`StrictSchema` hook** | **still engine** |

That is 1 in 5 on your own denominator (the five the pipeline resolved end to end) and 1 in 7 across all seven, against a 1-in-20 target. Against the framing above it lands harder: at 1 in 7, a seven-model lineup refresh still moves the engine version once, so the comparability property is not yet bought. Gemini-3.5-flash is the case the widening was largely written about, and under the widened design it would still cut an engine release.

This does not invalidate the ADR. It does mean the target reads better as measured than as an aspiration. The reason a hook is needed at all is that the bases are single large traversals with flags bolted on: `carry_scalar_dict_map_value` is one boolean, defined at `schema_sanitizer.py:214` and consulted at exactly one site, `:688`, inside a 630-line `StrictSchema`. Bases expressed as ordered named steps, with sequence and parameters coming from data, would turn most "new hook" cases into a rearrangement. That is a bigger change than this ADR should absorb, but it is the thing that moves 1-in-7 toward 1-in-20.

**I would move follow-up 13 (the acceptance test) to the front.** It is the only artifact that reports whether the mechanism is delivering, and built first it reports that at every step rather than at the end.

**One spec detail for it, which is also why front-loading matters.** The replay has to reconstruct each integration against the engine tree *as it stood on that integration's date*. Follow-up 8 sweeps the `StrictSchema` hook in retroactively, so a replay against post-follow-up-8 `main` finds the hook already present, passes #560, and reports 0-in-7. The quantity the test exists to measure is "how often did a model need a hook that did not yet exist", and that is only answerable at the tree of the day.

**And follow-up 12 should move forward with it.** § Auto-integration workflow retargeting makes repointing `integrate-model.yml` *"deliberately the last implementation step"*, on the reasoning that the manual `tolokaforge_models/` path should be proven end to end first. Proving the manual path first is right; leaving the automated one until last is what I would reconsider. Until the pipeline is retargeted, every real integration still goes down the engine path, so the property in the framing note is untested in production and the acceptance test is measuring a route nothing uses. I would land follow-up 12 right after the first manual Bucket A integration succeeds, rather than after everything else, so the engine version can be observed staying put across real onboardings while there is still room to change the design.

### 2. The forwarding stub: the warning has no delivery path, and the failure mode gets quieter

**There is no site to attach the warning to.** The known consumer reads bytes through `importlib.resources.files("tolokaforge").joinpath("core/data/pricing.json")`. That does run `tolokaforge/__init__.py`, but nothing under `core/` is imported (`core/data/` is not even a package), so the read is indistinguishable from any other import of the engine for any reason. A warning in `__init__.py` fires for everyone and names no one.

**A build-time stub goes stale silently.** It is generated when the engine is released; `tolokaforge-models` then ships on its own cadence. From the first independent models release onward the stub serves prices that are internally consistent and wrong. Today's failure mode is at least detectable: the collector's `load_pricing_table` path logs `engine_pricing_missing` and returns `{}`, so costs read as zero. Plausible-but-stale prices in a leaderboard cost column are not detectable.

**Suggestion: no stub, and fix the consumer first.** (1) `benchmark-results-collector` reads through `tolokaforge.core.model_data.bundled_pricing_path()` and fails loud on an empty table instead of returning `{}`, shipped ahead of the move; (2) the data then moves with no stub at all. I can take step 1 on the collector side. Worth a sweep for other resource-path readers before step 2 — `_runner_subset.py` documents `importlib.resources.files("tolokaforge")` as a supported lookup across both wheel variants, so the contract is broader than one consumer.

### 3. A ninth lifecycle stage: assembling the request message list

Requirement 1 names eight stages and the widened design gives all eight a slot. The message list itself is not among them: how the conversation is assembled before it goes on the wire (ordering, merging, filler injection) has no extension point.

`ToolContentPolicy.inject_empty_assistant_filler` is a per-preset flag, but the string it injects is hard-coded at `client.py:1213`, and the comment above it records that applying it too broadly caused a Gemini regression on 2026-04-30. So the value is model-specific by construction and stays in the engine after the widening. `assistant_text_policy` does not cover it: that reshapes what the model returns, not what we send.

**A correction here that makes the case stronger, not weaker.** § Extension-point surface describes Cohere's `<|START_TEXT|>…<|END_TEXT|>` handling as *"today an `if target == "anthropic"`-adjacent branch in `client.py`"*. There is no such branch — `START_TEXT` appears nowhere under `tolokaforge/` on any ref. I think this is my fault: my issue comment said that work "touches `client.py`", meaning the git branch does, and it read as an existing branch in the file. The actual state is an unfixed blocker, recorded at `tests/integration/llm/registry.py:2487` on the open Cohere integration (#929):

> response text arrives wrapped in ``<|START_TEXT|>...<|END_TEXT|>`` (observed 2026-08-01). ``ResponsePolicy`` cannot reach it, because that hook only post-processes tool-call arguments, so the markers land in ``trajectory.yaml`` and in front of any transcript grader or LLM judge, depressing scores for a reason that is not capability. It needs an engine-side text hook.

That is better evidence than a branch to relocate. It is a model blocked right now because a lifecycle stage did not exist when it arrived, which is exactly the Bucket B category routed to an engine release. `assistant_text_policy` is not a refactor of shipped code; it is a slot invented because a model showed up that no slot fit, six days ago. The ninth stage will be found the same way, and the 1-in-20 target needs a budget line for "new stage discovered by a new model". Cohere is worth its own row in § Evidence.

On the ninth slot itself: the narrow version is worth doing regardless — the filler text becomes policy data alongside the existing flag, which removes a hard-coded model-specific string from the engine and needs no new slot. If requirement 1 is kept as stated, the general version is a slot receiving the assembled message list before send, and it needs a guard, because a message-list rewrite is powerful and in an evaluation a bad rewrite produces wrong results rather than an error. I would pair it with an invariant check on the way out (roles valid, tool-call ids still paired, nothing dropped), with the filler as its first user.

### 4. Smaller fixes

**"Class-name-existence at engine import" inverts the safe order.** `tolokaforge/__init__.py` is 67 lines with a lazy `__getattr__`, and `presets.py` imports every policy module at its own top (lines 21-63), so by the time a models policy module can be imported, `GeminiSchema` and friends resolve from `sys.modules`. Models importing engine bases is fine on that ordering. The package `__init__` is the one placement that breaks it, since it would have to pull the LLM layer and the models package from a point where neither is loaded. After `presets.py`'s own imports, or in `prepare` alongside `write_engine_run_state`, is early enough to fail before a run spends anything and keeps the ordering intact.

**Keep `minimum_engine_version` a runtime check.** As a real `Requires-Dist` the two wheels' constraints become mutually referential, and a pair like "engine 0.16 needs models >= 1.0" plus "models 1.4 needs engine >= 0.17" is unsolvable for the resolver rather than a clear error at startup. The ADR has this right; one sentence saying it is deliberate would stop it being tidied into a real dependency later.

**The nested-unknown-key hardening buys strictness at the price of a new engine-release trigger.** This one is a refinement of my own ask. `_check_block` (`presets.py:302`) today validates slot names against the registries and `params:` keys against `_VALID_PARAMS_KEYS`; anything else inside a preset block passes. Made fail-loud, the legal key set becomes an engine-side allow-list, so a models-wheel preset carrying a knob the installed engine does not know hard-fails at load instead of no-opping. `gemini_drop_placeholder_signature` and the `api_call_timeout_s` family are exactly that shape: engine keys added for model-specific reasons. `{name, params}` absorbs the ones that are policy-class constructor arguments, not the ones that are `ModelCapabilities` fields. I still want the hardening — I would just validate against engine-known keys union the keys the models wheel declares, so strictness lands on typos rather than on version skew.

**The Nova provider record needs one more field.** Nova is three sites, not one: `client.py:688` (`_configure_nova_base_url`), `client.py:938` (`_format_model_name` returns the bare name for Nova), and the kwargs block the ADR cites. That block also sets `custom_llm_provider = "openai"` (`:1912`), for which the record has no field, and the slug handling is two steps rather than one template: strip a `nova/` prefix if present (`:1914`), then ensure an `openai/` one (`:1917`). The generic default is `self.provider.split("/")[0]` at `:1734`.

**Moving subclasses out is a public-API deletion.** Every class-adding integration also edited `tolokaforge/core/llm/__init__.py` (#560 +6, #719 +2, #846 +2) and the names are in that module's `__all__`. § The monorepo layout's *"Zero changes to any caller"* wants a caveat for that.

### 5. Answers to two of your review questions

**Q2, Capability enum ownership.** I would keep the enum engine-authoritative and let certificates carry additional capability names as data. The shared suite references enum members by attribute; if the models wheel owns the enum, a member renamed or dropped there breaks engine code through attribute access. That is the same failure class the class-name-existence check catches for policy classes, except attribute access gives you no registry to check against. Certificate-borne extras get requirement 8's decoupling without moving the authority.

**Q5, the current list of classes that do not fit "stable base + public API only".** Three, failing in different ways:

- `GeminiRecursiveSchema` (`schema_sanitizer.py:850`) is a clean subclass of `GeminiSchema`, but only because `StrictSchema` gained `VALUE_FIELD` (`:203`) and `carry_scalar_dict_map_value` (`:214`) for it — 16 added lines in #560. It fits once follow-up 8 lands the hook, which is §1 above.
- `MinimaxM3TagRecoveryResponse` (`response_policy.py:474`) is not a subclass at all. It is a composite reusing the module-private `_coerce_empty_containers` (`:69`) and `_coerce_json_strings` (`:107`). It fits after follow-up 7.
- `RefResolvingDictMapHints` (`prompt_policy.py:107`) is the deepest. It imports `_find_additional_properties` from `_dict_maps`, deliberately excluded from that module's `__all__` (`["DictMapParam", "detect_dict_maps"]`), and overrides `_build_hints`. Follow-up 7 already names `_build_hints`, so this is covered in intent — but the parent is a `@staticmethod` (`prompt_policy.py:64`) and the child an instance method with a `# type: ignore[override]` (`:145`). Turning it into an extension point is a change to the base class's *shape*, which § Docs flip's own Bucket B definition routes to the engine. Worth calling out so follow-up 7 is not scoped as a pure rename.

Small correction while on that list: § Evidence names the second class `MiniMaxM3StrictSchema`. It is `MinimaxM3TagRecoveryResponse` in `response_policy.py` — worth fixing so follow-up 7's promotion sweep is pointed at that module's privates.

bberkes-toloka:
Read the merged version. Everything landed, and the two places where you went past what I asked are the better calls: `capability_extras` is a cleaner answer to requirement 8 than either of the options I offered, and pinning the `minimum_engine_version` rationale in the text ("do not tidy it into a real dependency later") is exactly the kind of note that survives a year.

One thing to fix before implementation starts, and it is bookkeeping rather than design. **The `Capability` enum reversal did not propagate.** The doc now decides it four times in one direction and states the opposite in three places, and one of those is the requirements list an implementer builds from:

| line | says now | should say |
|---|---|---|
| 119, § Requirements 8 | *"The `Capability` enum lives in the models wheel"* | enum stays engine-authoritative; `capability_extras` carries the decoupling |
| 104, § Decision | lists the `Capability` enum as models-wheel content, and names three `ModelCertificate` fields | drop the enum; four fields, including `capability_extras` |
| 316, § monorepo layout | `tolokaforge_models/certificates/_capability.py  # Capability enum + widened ModelCertificate` | `ModelCertificate` only; the enum lives under `tolokaforge/testing/certify/` per § Certification |

The settled statements are at 208, 386 and review point 2 (✅), so the direction is not in doubt — these three are just stale.

**A fourth, subtler one in the same family.** Follow-up 2 says *"Move `_capability.py`, `registry.py`, fixtures, ~30 test bodies"*. After the reversal that file holds two things with different destinations: the enum stays engine-side, `ModelCertificate` moves. Worth splitting explicitly in the ticket, or the implementation quietly restores what the reversal decided against.

**And one unrelated to the enum.** § Decision Drivers still carries *"class-name-existence check at engine import"*, which § Install-time validation now explicitly rules out (*"**not** in `tolokaforge/__init__.py`"*, placement after `presets.py`'s imports or in `prepare`). Same fix, one bullet.

Nothing else from my side. On the landing order: I had asked for the workflow retarget to move up and I was over-reaching — it cannot precede the package existing, so right after the cutover is the earliest real slot, and the acceptance test at position 3 is what actually delivers continuous measurement. That reads right as merged.

PR #914, opened 2026-08-06
fix(grading): hash-source rule skips a pack whose adapter may supply the source (#911)

### LINKED ISSUE #911: bug: pre-run grading gate refuses external-adapter packs whose adapter supplies the hash source
## Context

v0.15.0's pre-run authoring gate refuses any pack whose `state_checks.hash` block is enabled but declares neither `expected_state_hash` nor `golden_actions` — as an **error**, fatal at every `evaluation.grading_validation.fail_on` setting. That rule is correct for native packs, but it also fires on packs graded by external adapters that supply the comparison source themselves: the `frozen_mcp_core` / `tlk_mcp_core` adapters replay `fixtures/golden_actions.json` and compute the golden hash internally, and the bare `hash: {enabled: true, weight: 1.0}` block is those adapters' own documented convention (the adapter synthesizes exactly that block for tasks without a grading.yaml).

The gate already routes adapter-owned readings to the never-fatal `unchecked` channel — tool names and the golden-replay world are skipped for external adapters precisely because "reporting a reading the adapter does not use would reject packs that run fine." The hash-source rule needs neither of those inputs, so it is the one hash rule that escapes the skip and fires on external packs anyway.

Observed: every `evaluate` shard of the downstream benchmark run fails before any trial executes — all `ots_bank_hr_d365` packs are rejected (e.g. https://github.com/toloka-partners/tolokaforge-tasks/actions/runs/31127175869). Downstream evaluation is blocked on the released v0.15.0, so this needs a patch release.

## Deliverables

- [ ] A pack that names a grading file and is graded by an external adapter is no longer refused for an enabled hash block with no declared source: the finding lands on the `unchecked` channel with a reason saying the adapter may supply the source, mirroring how tool names and the replay world are skipped.
- [ ] Native packs keep the current hard error, verbatim.
- [ ] The signal saying "the authored block is the only place a hash source can come from" is explicit (the same pattern as the existing unresolvable tool-inventory / replay-world / combine-layer answers), not a proxy read off an unrelated fact.
- [ ] Behaviour-lock test reproducing the failing downstream shape: external adapter + `grading:` file + `hash: {enabled: true, weight: 1.0}` → run proceeds, skip surfaced.

## Acceptance Criteria

- `tolokaforge run` over an `ots_bank_hr_d365`-shaped pack selection (external adapter, bare enabled hash block) reaches trial execution instead of raising "cannot be graded as written".
- `tolokaforge validate` on the same packs exits 0 and prints the skip beside the task.
- `tolokaforge validate` on a **native** pack with an enabled, sourceless hash block still fails with the existing message.

## Design Decisions

- **Unchecked, not silently passed**: a gate that checked nothing must not read as a clean bill of health — the skip is surfaced beside the task like every other adapter-owned reading.
- **Explicit new signal over a proxy**: gating off the unresolvable tool inventory or replay world would make the skip reason lie about what was unresolvable; the rule's question ("is the authored block the only hash source?") deserves its own answer type with an `unresolvable()` state.
- **Hotfix off `main` (v0.15.0), independent of #730**: the typed-hash-block work in flight does not change this rule; coupling the downstream unblock to that review would delay it.

## Out of Scope

- Letting adapters *answer* the hash-source question (report a fixture-backed source and whether it exists) instead of unchecked silence — that is the follow-up adapter hook, tracked in toloka-partners/tolokaforge-tools (adapter side) and a future harness issue once the hook contract is designed.

azorej:
Closes #911.

## What

`_check_hash_source_declared` refused every frozen-core pack at v0.15.0: their convention is `hash: {enabled: true, weight: 1.0}` with the comparison source living in a golden-actions fixture the authored block never names, and this was the one hash rule that fired whatever the adapter — every `evaluate` shard of the downstream benchmark run aborted before its first trial (e.g. toloka-partners/tolokaforge-tasks run [31127175869](https://github.com/toloka-partners/tolokaforge-tasks/actions/runs/31127175869)). The finding is an **error**, so no `fail_on` setting could waive it.

## How

A new `HashSourceLayer` beside `CombineLayer` / `ReplayWorld`: what supplies a hash source beneath the authored block. Resolved for a native task (the block is the whole layer — the current refusal, verbatim); `unresolvable()` for a task an external adapter grades, which moves the finding either half of the rule would have drawn to the never-fatal `unchecked` channel, addressed where the refusal would have been. Both pack gates (pre-run + `tolokaforge validate`) resolve it through `hash_source_layer_under_adapter`, the same discrimination `tool_inventory_under_adapter` / `replay_world_under_adapter` already make. The skip fires only on the shape a native pack would have been refused for, so healthy external packs stay noise-free.

Letting an adapter *answer* with the source it supplies (so a missing fixture is refused before the trial is paid for) is deliberately out of scope — tracked in toloka-partners/tolokaforge-tools#94.

## Behaviour locks

- gate: the OTS shape under an unresolvable layer is unchecked at `state_checks.hash.enabled`, not refused; a clean block reports nothing on either layer; both `_RULES` rows for the unresolvable layer land in `unchecked`
- orchestrator: an external-adapter pack naming a grading file with the bare enabled hash block reaches its trials, skip logged beside the task
- CLI: `tolokaforge validate` passes the same pack (`1 valid, 0 invalid`) and prints the skip; native sourceless-hash refusals unchanged

## Tests

```
uv run pytest tests/ -m "unit or canonical" --deselect tests/unit/test_persistent_shell_local.py
6864 passed, 1701 deselected in 122.23s
```

(`test_persistent_shell_local.py` fails identically on an untouched `main` checkout in this environment — unrelated.)

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -507,12 +507,14 @@ llm_judge:                                 # the judge MODEL is set once per run
 
 A whole-state hash is read only where the flag turns it on: `state_checks.hash`
 carrying an `expected_state_hash` or `golden_actions` under an `enabled` that is not
-truthy is rejected at load, because both substrates test the flag before reading any
-source and the pack would otherwise grade its state without the comparison the author
-wrote. The refusal is addressed at the source the pack declared. Every other
-authoring rule this file's grading block is checked against — tool names, argument
-names, `regex` compilation — is in
-[GRADING.md](GRADING.md#what-is-validated-before-a-run).
+truthy is rejected at load where the task declares `adapter_type: native`, because
+both substrates test the flag before reading any source and the pack would otherwise
+grade its state without the comparison the author wrote. The refusal is addressed at
+the source the pack declared; under any other `adapter_type` the same shape is
+reported unchecked at the same address, because an external adapter may supply the
+source the authored block never names. Every other authoring rule this file's
+grading block is checked against — tool names, argument names, `regex` compilation —
+is in [GRADING.md](GRADING.md#what-is-validated-before-a-run).
 
 The rubric is a structured `Rubric` (per-criterion scoring + a required gate),
 not a free-text blob; a free-text `rubric: "<text>"`, an `output_schema` field,
--- a/docs/GRADING.md
+++ b/docs/GRADING.md
@@ -2376,8 +2376,9 @@ Findings come in three classes:
 | `db_probes` beside a non-empty `jsonpaths`, or beside a `hash` block enabled with a source — raised as a config load error before the gate is reached, so it is reported alone | error | `state_checks.db_probes` |
 | a `transcript_rules` block declaring no rule at all — every list empty, both turn bounds absent, and a `tool_expectations` expecting neither tool | error | `transcript_rules` |
 | a `custom_checks` block with no `enabled` key, which the component's own default leaves unrun | error | `custom_checks` |
-| either hash source declared under a `hash.enabled` that is not truthy — written `false`, `0`, `null`, or absent | error, one for the block | `state_checks.hash.<the declared source>`, and `expected_state_hash` where both are declared |
-| a truthy `state_checks.hash.enabled` with neither `expected_state_hash` nor a non-empty `golden_actions` | error | `state_checks.hash.enabled` |
+| either hash source declared under a `hash.enabled` that is not truthy — written `false`, `0`, `null`, or absent — where the task's declared `adapter_type` is `native` | error, one for the block | `state_checks.hash.<the declared source>`, and `expected_state_hash` where both are declared |
+| a truthy `state_checks.hash.enabled` with neither `expected_state_hash` nor a non-empty `golden_actions`, where the task's declared `adapter_type` is `native` | error | `state_checks.hash.enabled` |
+| either flag/source mismatch above, where the task declares any other `adapter_type` — an external adapter may compute the source it compares against from its own fixtures, the way the frozen-core family replays a golden-actions fixture the block never names | unchecked | the address the error would have carried |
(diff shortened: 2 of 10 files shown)


azorej:
**Review + downstream verification status** (session-local reviewer & eval agents):

**Adversarial review** returned 1 blocker + 2 majors, all addressed in 1e18165c:
- *Docs*: two GRADING.md sections (substrate-parity bullet, "flag and source declared together") still asserted the unconditional refusal — now qualified to the native reading, with external-adapter packs named as reachable by design.
- *Native-arm lock*: a mutation making `hash_source_layer_under_adapter` return unresolvable for **every** adapter type previously survived the whole unit+canonical lane. New CLI test (`test_validate_refuses_a_native_packs_sourceless_hash_through_the_resolver`) exercises the resolver end to end with the native default and fails under that mutation (verified by re-running it).
- *Corpus guards*: both `test_example_pack_grading_corpus.py` walks now resolve the hash layer off each pack's own `task.yaml` (as they already do the replay world), so the rule's error branch stays exercised over the 58/93-pack walks.

**Downstream verification** against `toloka-partners/tolokaforge-tasks@dev/ots` with this branch:
- `tolokaforge validate` over `ots_bank_hr_d365`: **177 valid, 0 invalid** — all 9 CI offenders pass, each printing the new unchecked skip. All domains: **2002 valid, 0 invalid**.
- Negative control: same pack flipped to `adapter_type: native` → refused with the verbatim v0.15.0 error.
- Real `tolokaforge run` of BEN-124 (frozen adapters + live keys): past `_reject_ungradeable_packs` (finding logged as WARNING), stack up healthy, trial provisioned, real agent turns started. The adapter's `golden_actions_count=10` in the TaskDescription confirms the fixture-supplied source this fix exists for.

azorej:
Review loop complete: the session reviewer re-verified each finding and **approved at `392cccae`** — native-arm mutation lock confirmed independently (the always-unresolvable mutation kills exactly the new CLI test), corpus walks resolve the layer, and no unqualified refusal claims remain in docs (GRADING.md + CONFIG.md swept on the new head). Doc-only final commits; code unchanged since the verified `1e18165c`.

Issue #927, opened 2026-08-07
enhancement: Resolve the TypeSense address per-stack instead of smuggling it through each task's SearchConfig


azorej:
## Context
The TypeSense address is a property of the running stack (inside the Docker network it is always the same alias and container port), yet today each task's SearchConfig carries host/port/api-key values that the orchestrator must rewrite after stack start and each adapter must snapshot-vs-propagate correctly. This double-bookkeeping already produced one release-blocking regression (stale description cache serving the host-side address) and forces every adapter to understand Docker networking it has no business knowing.

## Deliverables
- [ ] The runner resolves the TypeSense connection (host/port/api-key) from stack-level configuration injected at container start, falling back to per-task SearchConfig for older adapters
- [ ] Per-task SearchConfig keeps only genuinely per-task search facts (domain name, documents path)
- [ ] The orchestrator's post-stack-start rewrite of adapter TypeSense params (and its cache invalidation) is removed once no consumer needs it
- [ ] Adapters in tolokaforge-tools stop emitting connection details into SearchConfig (follow-up PR there, pinned after the engine release)

## Acceptance Criteria
- Full_stack KB runs succeed with adapters that emit no connection details in SearchConfig
- Runs with current (connection-emitting) adapters still succeed during the transition
- API key reaches the runner via the SecretManager path, never as a raw env literal in a service definition

## Design Decisions
- **Stack-level injection at container creation**, because the in-network alias and container port are static by construction — no post-start rewrite, no ordering hazard, no per-task copies to go stale.
- **Env-first with SearchConfig fallback in the runner**, because it makes the engine release deployable before the adapters are slimmed (same pinning discipline as the tools adapter refresh).

## Dependencies
- Sequencing: engine change releases first; the tolokaforge-tools SearchConfig slimming lands only after downstream re-pins that engine version.
- Related in spirit to toloka-partners/tolokaforge-tools#94 (formalising the harness↔adapter contract instead of routing stack facts through task descriptions).

## Out of Scope
- The rag-service plane (`SearchConfig.enabled`) — unchanged
- Host-side indexing (adapters keep using the host-mapped address for that)

azorej:
[...]

Issue #926, opened 2026-08-07
bug: Runner degrades silently when a configured TypeSense client is unusable — run reports success while every search call fails


azorej:
## Context
In [tolokaforge-tasks run 31150454474](https://github.com/toloka-partners/tolokaforge-tasks/actions/runs/31150454474) the search plane was dead for the whole run (326 failed `search_policy` calls), yet CI concluded "success" and the results were graded as measured agent behaviour. When a task's search configuration names a TypeSense host, the runner currently logs a warning and continues if it cannot register a usable client — every subsequent search call then fails at the agent's expense, which violates the fail-fast rule (AGENTS.md rule 1) and corrupts scores invisibly.

Review of the #925 fix (PR #928) surfaced a second fail-soft site in the same plane, on the orchestrator side: the TypeSense→runner-network bridge downgrades every failure to a warning (broad exception swallow plus two silent early-returns when the TypeSense container or the runner network is missing). Each of those paths leaves the run going with the host-side address reaching the runner container — the #925 symptom arrived at by a different route, while the neighbouring stack-startup code raises loudly for comparable failures.

## Deliverables
- [ ] Trial registration fails loudly when the task declares a TypeSense host but no usable client can be registered (unreachable server, missing docindex snippets, or unavailable registry)
- [ ] The orchestrator's TypeSense→runner-network bridge fails loudly too: a run that enabled TypeSense and could not bridge it to the runner network aborts instead of warning and continuing
- [ ] Failures name the task (or run-level component), the address tried, and the reason — surfaced as run-level errors, not per-tool-call warnings
- [ ] Behaviour-lock tests covering the unreachable-server, missing-snippets, and failed-bridge cases

## Acceptance Criteria
- A run whose search plane is misconfigured aborts at (or before) the first affected trial with an actionable error instead of completing with degraded scores
- Tasks without a knowledge base are unaffected; runs with TypeSense disabled are unaffected
- Unit + canonical lanes green

## Design Decisions
- **Fail at trial registration, not at first tool call**, because registration is the earliest point where the runner knows both the search requirement and the client state — failing there costs zero paid agent turns.
- **Fail the bridge at stack startup**, because that is where the orchestrator already knows TypeSense was requested and the network exists — matching the loud-raise convention of the surrounding stack-startup code.
- **Keep "no KB declared" as a non-error**, because domains without docindex legitimately run searchless; only a declared-but-unusable search plane is a failure.

## Dependencies
[...]

PR #928, opened 2026-08-07
fix(core): the TypeSense Docker rewrite drops the description cache

### LINKED ISSUE #925: bug: Pre-run grading gate caches TaskDescriptions with host-side TypeSense address — knowledge search dead in every full_stack run
## Context
Since v0.16.0, every `search_policy` call in Docker full_stack runs fails with "Search service is not available" (e.g. [tolokaforge-tasks run 31150454474](https://github.com/toloka-partners/tolokaforge-tasks/actions/runs/31150454474): 326 failures across all 4 shards; agents document the outage in case notes and escalate, silently corrupting scores). The pre-run grading gate introduced in #890 resolves every task's description at run start — before the Docker stack exists — so the run-scoped description cache holds the host-side TypeSense address; the Docker-network rewrite that happens after stack start never reaches any trial, and inside the runner container that address points at the runner itself.

## Deliverables
- [ ] Trials in a full_stack run receive the Docker-reachable TypeSense address even when the pre-run gate resolved their descriptions first; `search_policy` returns real results again on frozen packs with a knowledge base
- [ ] A behaviour-lock test that fails on the current code: descriptions resolved before the stack starts must be re-resolved with the rewritten TypeSense address afterwards

## Acceptance Criteria
- A full_stack benchmark run on a KB domain (e.g. `ots_bank_hr_d365`) completes with zero "Search service is not available" tool errors
- The new test fails when the cache-invalidation is removed and passes with it
- Unit + canonical lanes green

## Design Decisions
- **Invalidate the description cache when the orchestrator rewrites the adapter's TypeSense params**, because the cache is correct at every other point in the run — the rewrite is the single event that makes cached entries stale. Rebuilding descriptions once more per run re-runs only cheap registration checks.
- **Keep the pre-run gate resolving through the cache**, because the gate's registration-check-on-build contract is the reason the cache exists; bypassing it would double description builds for nothing.

## Out of Scope
- Failing loud when the search plane is down at trial registration (separate bug)
- Moving the TypeSense address out of per-task SearchConfig entirely (separate enhancement)

## Rollout
Ship as v0.16.1; `tolokaforge-tasks` `dev/ots` re-pins afterwards (same flow as the #911/v0.16.0 cycle).

azorej:
Fixes #925.

## What broke

Since v0.16.0, every `search_policy` call in a full_stack Docker run fails with `Search service is not available` (326 failures across all 4 shards of [tolokaforge-tasks run 31150454474](https://github.com/toloka-partners/tolokaforge-tasks/actions/runs/31150454474); agents document the outage in case notes and escalate, so KB scores are corrupted while CI reports green).

The pre-run grading gate (#890) resolves every selected task through `_task_description()` at the top of `run()` — before the Docker stack exists — so the run-scoped `_task_desc_cache` fills with descriptions whose `SearchConfig` carries the **host-side** TypeSense address. When `_connect_typesense_to_runner_network` later rewrites `adapter.params["typesense"]` to the Docker alias (`typesense:8108`), that rewrite only affects future description builds — and the warm cache guarantees there are none. Trials ship `host=127.0.0.1` into the runner container, where that address is the runner itself: mcp_core registers no usable client, `get_typesense_for_domain` returns `None`, and the frozen pack's `search_policy` raises on every call.

## The fix

One hunk: the rewrite clears `_task_desc_cache` right after propagating the Docker address to the adapter params, so trial-facing resolves rebuild against the rewritten values. The gate keeps resolving through the cache (its registration-check-on-build contract is why the cache exists); the cost of the invalidation is one extra description build per task per run.

## Behaviour locks

`tests/unit/test_orchestrator_typesense_cache_invalidation.py`, driving a real `Orchestrator` through the real `_connect_typesense_to_runner_network` (only the Docker SDK boundary and the adapter are stand-ins):

- a description resolved before the stack starts is re-resolved with the Docker alias afterwards — **fails on current main** (`127.0.0.1` is served back from the cache), passes with the fix;
- the cache is dropped, not disabled: one rewrite costs exactly one rebuild per task.

## Verification

- `pytest tests/unit -k orchestrator`: 233 passed
- full unit lane: 5771 passed; the only failures are the 8 `test_persistent_shell_local.py` cases that fail identically on untouched main in this devcontainer (environmental)
- canonical lane: 1101 passed
- `ruff check` clean on touched files; `ruff format` drift in `orchestrator.py` is pre-existing (lines ~206, fails the check on baseline too — AGENTS.md gotcha #3); `black --check` clean

## Rollout

Ship as v0.16.1, then re-pin `tolokaforge-tasks` `dev/ots` (same flow as the #911/v0.16.0 cycle). Follow-ups filed separately: #926 (runner fails loud when a declared search plane is unusable), #927 (stack-level TypeSense address instead of per-task SearchConfig).

https://claude.ai/code/session_01CajmZnqsPoV4MC8E96UuEf

### DIFF
--- a/tests/unit/test_orchestrator_typesense_cache_invalidation.py
+++ b/tests/unit/test_orchestrator_typesense_cache_invalidation.py
@@ -14,6 +14,13 @@ The lock sits on the two private seams the regression lives between —
 (the rewrite) — because the end-to-end surface needs a Docker daemon the unit
 lane does not have. The orchestrator is real; only the Docker SDK boundary and
 the adapter are stand-ins.
+
+One contract this file assumes rather than proves: the mcp_core adapters read
+``params["typesense"]`` at ``to_task_description`` time, not at construction
+time (their construction-time snapshot serves host-side indexing only). No
+in-repo adapter exercises that read, so an adapter that started snapshotting
+would regress to #925 with these tests still green — that end of the contract
+belongs to the external adapters' own suites.
 """
 
 from __future__ import annotations
--- a/tests/unit/test_orchestrator_typesense_cache_invalidation.py
+++ b/tests/unit/test_orchestrator_typesense_cache_invalidation.py
@@ -1,0 +1,177 @@
+"""A description cached before the TypeSense Docker rewrite must not reach a trial.
+
+The pre-run grading gate resolves every selected task through the run-scoped
+description cache before the Docker stack exists, so each cached ``SearchConfig``
+carries the host-side TypeSense address. After the stack starts,
+``_connect_typesense_to_runner_network`` rewrites the adapter's TypeSense params
+to the Docker-network alias — and inside the runner container the host-side
+address points at the runner itself, so a trial served the stale description has
+every ``search_policy`` call die with "Search service is not available" (#925).
+The rewrite therefore drops the cache, and this file locks that.
+
+The lock sits on the two private seams the regression lives between —
+``_task_description`` (the cache) and ``_connect_typesense_to_runner_network``
+(the rewrite) — because the end-to-end surface needs a Docker daemon the unit
+lane does not have. The orchestrator is real; only the Docker SDK boundary and
+the adapter are stand-ins.
+"""
+
+from __future__ import annotations
+
+import sys
+import types
+from pathlib import Path
+from typing import Any
+
+import pytest
+
+from tolokaforge.core.conductor import InMemoryConductor
+from tolokaforge.core.models import (
+    EvaluationConfig,
+    ModelConfig,
+    OrchestratorConfig,
+    RunConfig,
+    TypeSenseConfig,
+)
+from tolokaforge.core.orchestrator import Orchestrator, OrchestratorDeps
+from tolokaforge.core.runtime import InMemoryRuntimeBackend
+from tolokaforge.runner.models import SearchConfig, TaskDescription
+
+pytestmark = pytest.mark.unit
+
+HOST_SIDE = {"enabled": True, "mode": "local", "host": "127.0.0.1", "port": 8108, "api_key": "k"}
+
+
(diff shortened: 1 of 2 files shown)


azorej:
## Real-run verification + review

**End-to-end verification on real infrastructure** — this branch, full Docker stack, frozen adapter at the same tools rev CI used (`4e60454`), task `ots_bank_hr_d365/LTO-033` (the task whose CI trials documented the search outage in case notes and escalated), real models via OpenRouter (Opus 4.8 agent / Sonnet 4.6 user / Gemini 3.6 Flash judge):

- `Updated TypeSense config for Docker networking (host=typesense, port=8108)` fired, followed by the new cache invalidation
- `connectors_typesense_search_policy` (query: *"jury duty leave request process"*) → **status: success**, returning real KB snippets from the leave-management policy doc
- **Zero** `Search service is not available` errors anywhere in the run (326 of them in the equivalent v0.16.0 CI run)
- Trial completed the jury-duty request instead of escalating: **score 1.0** (state hash match + judge 1.00), 11 tool calls, 8 turns, `failed_attempts=0`
- Run cost: ~$0.85 including judge

**Adversarial review** (branch-code-reviewer): **APPROVE WITH NITS.** The reviewer independently re-verified the behaviour lock (replacing the `clear()` with `pass` fails both tests), confirmed the cache-clear is atomic with the params rewrite on every failure path, audited all four `_task_desc_cache` consumers for whole-cache-clear safety, and validated the test stand-ins against the real `EngineStack`/`Network`/`Container` attribute surfaces. Both nits are addressed in `11af44df` (comment-rot fixes + recording the out-of-repo adapter contract the test assumes). The one Major it raised is pre-existing fail-soft behavior in the TypeSense bridge — folded into #926's scope rather than expanding this PR.

PR #942, opened 2026-08-07
feat(llm): address the gateway in its own dialect and by its own route name

balazs-toloka:
## The problem

An evaluation run cannot use a gateway for the models the gateway actually serves. Two independent defects, both measured against a live LiteLLM proxy:

**1. The gateway is addressed in the wrong dialect.** litellm's OpenRouter transformation unconditionally adds `usage: {"include": true}` to the request body ("ALWAYS add usage parameter to get cost data from OpenRouter", `openrouter/chat/transformation.py`). That is an OpenRouter extension. A gateway forwards it to whatever upstream backs the route, and every non-OpenRouter upstream rejects it:

```
anthropic-direct/claude-sonnet-4-6   usage: Extra inputs are not permitted   (Anthropic)
anthropic/claude-sonnet-4-6          usage: Extra inputs are not permitted   (Bedrock)
```

**2. The gateway is addressed by the wrong name.** The gateway's route name is not the engine's model string, and is not derivable from it. Worse, the two defects are coupled: the OpenRouter transport strips one leading `openrouter/` and the OpenAI transport does not, so *changing the dialect changes the name that arrives*. Measured, same model, same gateway:

| dialect | name on the wire | result |
|---|---|---|
| `openrouter` | `anthropic/claude-sonnet-4.6` | `model identifier is invalid` |
| `openai` | `openrouter/anthropic/claude-sonnet-4.6` | works |

and in the opposite direction for a gateway-only model:

| dialect | name on the wire | result |
|---|---|---|
| `openrouter` | `azure_ai/cohere-command-a-plus-05-2026` | works |
| `openai` | `openrouter/azure_ai/cohere-command-a-plus-05-2026` | `is not a valid model ID` |

So neither dialect is right for both, and no fixed name rule works either.

### What this cost, concretely

A real auto-integration run over the `litellm` route ([31172961198](https://github.com/Toloka/tolokaforge/actions/runs/31172961198)) went `needs-human: observe infra-dirty` with **95 API errors and 948 `model identifier is invalid`** in the wire probes. The candidate reached the gateway; the user simulator alongside it did not, because its name is only valid under the other dialect. `integrate-model.yml`'s own comment records the constraint that produced this: *"the gateway must serve BOTH models"*.

This is also the deferred work `proxy.py`'s module docstring already names:

> Widening the default to *every* provider would need this module to force the OpenAI envelope (`custom_llm_provider="openai"` on routed calls, as the Nova path already does) rather than only swapping the base URL. That is a larger change than a transport swap [...] and is deliberately not attempted here.

## The fix

A gateway is an OpenAI-compatible endpoint, so routed calls speak that dialect and address the gateway by **its** route name, resolved from `GET {base}/models` at client construction and cached per base URL.

The route name is whichever of `<provider>/<name>` or `<name>` the catalog contains. That is deterministic, needs no fuzzy matching, and covers both shapes above.

Three outcomes, deliberately different:

| catalog says | what happens |
|---|---|
| names the model | route through the gateway under that name, OpenAI dialect |
| answers and omits it | the gateway does not serve it, so call the provider directly |
| unreadable, or an empty answer | keep the gateway, send the untranslated name; a failure starts a 60s cooldown so a hanging catalog is not re-fetched per construction |

The third row is the one worth arguing about. **Unreadable is not absence.** Treating a network blip as "not on the gateway" would silently take a run off the gateway, which is precisely the unattributed-spend outcome this transport exists to prevent. Only a catalog that answers *with something* is authoritative: an empty `data` list is a broken answer (a key without model-list permission, a gateway mid-reload), not a claim that the gateway serves nothing.

Three revisions of this branch got parts of that wrong and were caught rather than shipped: the first disabled the gateway on an unreadable catalog (eight existing tests failed); the second still left it on an *empty* one; the third mapped an empty answer correctly but then negative-cached it, pinning the whole process on the degraded path after one blip. All three are now pinned by tests, the last one against a real local endpoint.

The second row is what makes a mixed run possible at all: a gateway-only candidate and a simulator the gateway does not carry can now run side by side, each on the transport that serves it. That removes the "gateway must serve BOTH models" constraint.

### Ambiguity is refused, not guessed

When the catalog serves one model under several names, those names can be backed by *different upstreams*. Picking one silently would change the serving path, which is a leaderboard-comparability decision rather than a transport detail, and `gateway_catalog.py` already argues this in its own docstring. `LLM_PROXY_PREFERRED_ROUTE` names the namespace that wins; without it the resolver raises and says so.

A documented hard requirement comes with the deterministic rule: **gateway route names must mirror upstream names.** An aliased route (`sonnet-4.6`, `team-default`) is invisible to the two derived candidates, resolves as "not served", and the call quietly goes direct. The docs record that, and record the extension point if everything should ever route through the gateway: a namespace-matched wildcard (accept `openrouter/*` only for `provider: openrouter` configs, where the fallback is the same upstream the board is calibrated on), not looser matching.

## Verification

End to end against the live gateway, all three cases, on this branch:

```
candidate (gateway-only)   gateway=yes          route=azure_ai/cohere-command-a-plus-05-2026
   -> SUCCESS 126 in / 52 out
simulator (on both)        gateway=yes          route=openrouter/anthropic/claude-sonnet-4.6
   -> SUCCESS 15 in / 4 out
not on gateway             gateway=NO (direct)  route=None
   -> SUCCESS 20 in / 4 out
```

The middle line is the 948-error case from the failed run, now working.

```
$ pytest tests/unit/llm/test_gateway_route.py \
         tests/unit/llm/test_gateway_routing_applied.py \
         tests/unit/llm/test_gateway_catalog_fetch.py \
         tests/unit/llm/test_llm_proxy.py \
         tests/canonical/test_llm_gateway_envelope_contract.py -q   86 passed
$ ruff check tolokaforge/ tests/                                    All checks passed!
$ black --check tolokaforge tests                                   clean

$ pytest tests/unit tests/canonical -q
47 failed, 6889 passed, 10 errors
```

**Those 47 are pre-existing on this `main`, not from this branch.** Zero of them are in
`gateway_route`, `test_llm_proxy` or the envelope contract. Checked rather than assumed:
the three largest failing files run on a clean `origin/main` checkout, with none of this
applied, give `31 failed, 42 passed`.

### The wiring is tested, not just the resolver

A resolver returning the right answer into a call site that ignores it is a feature that does not exist, so both halves are mutation-checked. Disabling the route rewrite plus the dialect switch fails 2 tests; never dropping the proxy when the catalog omits the model fails 1. Both of those mutations passed the whole suite before `test_gateway_routing_applied.py` existed.

The unit suite does no network I/O for the wiring: an autouse fixture in `tests/unit/llm/conftest.py` stubs the fetch and clears the cache for the whole LLM lane. Without it, every proxy-configured client construction really attempted a GET, and the suite was green only because that fetch failed.

The fetch itself is tested separately and for real, against a local `http.server`: request shape, the gateway key and admission headers riding along, every unreadable shape (empty list, wrong type, malformed body, 403, 500, unreachable host), and all three cache branches. A gateway that hangs is bounded to three attempts per base URL rather than costing the timeout on every client construction.

## Not in this PR

**The Slack poller now agrees with the engine about reachability** (its `lookup()` mirrors the engine's candidate order, so prefixed-only entries no longer report `absent`), but it still cannot read a header-admission gateway: it reads only the base URL and key from `os.environ` (by design, so a local `.env` cannot answer a real poll) and ignores `LLM_PROXY_HEADERS`. The full unification, one fetcher consuming a `ProxyConfig` with identical empty-catalog semantics everywhere, is tracked in #948.

**Pricing on gateway routes.** `normalize_model_name` keys off the engine's model string, which is unchanged, so pricing behaves exactly as before this PR. A gateway route whose name is absent from the pricing table still degrades `cost_source` to `unknown`, as it did before.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -41,3 +41,5 @@ OPENROUTER_API_KEY=your-openrouter-api-key-here
 # litellm transport sends an OpenAI-envelope request. Widen this only if your
 # gateway also serves the other provider's native route (see docs).
 # LLM_PROXY_PROVIDERS=openrouter,openai
+# Namespace that wins when the gateway serves one model under several names:
+# LLM_PROXY_PREFERRED_ROUTE=openrouter/
--- a/docs/AUTO_INTEGRATION.md
+++ b/docs/AUTO_INTEGRATION.md
@@ -190,7 +190,7 @@ The name looked up is the one that actually reaches the gateway: litellm strips 
 provider prefix, so this run's `provider: openrouter` + `name: <slug>` config puts the **bare
 slug** on the wire. An `openrouter/<slug>` (or `openrouter/*`) catalog entry is therefore *not*
 evidence for this run — reaching a prefixed route needs the gateway-named config in
-[`docs/LLM_LAYER.md` § the model name must be the gateway's route name](LLM_LAYER.md#the-model-name-must-be-the-gateways-route-name),
+[`docs/LLM_LAYER.md` § naming a gateway route explicitly](LLM_LAYER.md#naming-a-gateway-route-explicitly),
 which this workflow does not use.
 
 The report distinguishes two strengths, because they are not equally trustworthy:
--- a/docs/LLM_LAYER.md
+++ b/docs/LLM_LAYER.md
@@ -564,6 +564,24 @@ different upstreams, which is a serving-path choice rather than a transport deta
 `LLM_PROXY_PREFERRED_ROUTE` names the namespace that wins; without it the resolver
 raises rather than guessing.
 
+**Hard requirement on the gateway: route names must mirror upstream names.** The
+resolver derives exactly two candidates and does no fuzzy matching, so a route named
+as an arbitrary alias (`sonnet-4.6`, `team-default`, or a separator variant like
+`claude-sonnet-4-6` for a config that says `4.6`) is invisible to it. The model then
+resolves as "not served" and the call goes to the provider directly, with only a
+per-client warning. On a gateway-only model that direct call fails; on any other it
+runs unattributed. If a deployment cannot rename such a route, the exact-name entry
+has to be added alongside the alias.
+
+Wildcard entries (`openrouter/*`, `anthropic/*`) are deliberately **not** accepted as
+evidence: a wildcard says the gateway will forward the request, not that the model
+exists behind it. Measured on a live gateway, `anthropic/*` accepted a name that its
+Bedrock backing then rejected as invalid, while the very same wildcard also "covered"
+a nonexistent model. If routing every model through the gateway ever becomes the
+goal, the extension point is a *namespace-matched* wildcard (accept `openrouter/*`
+only for `provider: openrouter` configs, where the fallback is the same upstream the
+board is calibrated on), not looser matching.
+
 Preset resolution and pricing are unaffected: both key off `ModelConfig.provider` /
 `.name`, not off the wire name.
 
@@ -593,23 +611,25 @@ next to its own hardcoded base URL; a gateway replaces the base URL but not the
 rewrite, so litellm would get a provider-less model string and raise
 `BadRequestError` before sending anything.
 
-### The model name must be the gateway's route name
+### Naming a gateway route explicitly
 
-Enabling the gateway is **not** a drop-in for existing run configs. litellm
-strips exactly one provider prefix before sending, so `provider: openrouter` +
-`name: anthropic/claude-opus-4.7` puts `anthropic/claude-opus-4.7` on the wire.
-A gateway resolves that against *its own* model table, which need not mean what
-the provider would mean by it.
+Route resolution (above) makes existing run configs work through the gateway
+unchanged, so this section is the **manual override**: pinning a config directly to
+one specific gateway route, either because the catalog is unreadable from the
+running host or because a deployment needs a route the resolver's two candidates
+cannot derive (an alias, see the hard requirement above).
 
-Observed on a real LiteLLM proxy: that name matched the gateway's catch-all
-`anthropic/*` route, which is backed by **Bedrock**, so the request was served
-by a different upstream than the config asked for. It failed loudly there only
-because that particular Bedrock model rejects `temperature=0.0`; a
-closer-matching route would have silently evaluated a different serving path.
-For a leaderboard that is a comparability break, not a transport detail.
+The cautionary tale that motivates being explicit at all: before route resolution
+existed, `provider: openrouter` + `name: anthropic/claude-opus-4.7` put the bare
+`anthropic/claude-opus-4.7` on the wire (litellm strips one prefix), and on a real
+LiteLLM proxy that matched the catch-all `anthropic/*` route, backed by **Bedrock**,
+a different upstream than the config asked for. It failed loudly there only
+because that particular Bedrock model rejects `temperature=0.0`; a closer-matching
+route would have silently evaluated a different serving path. For a leaderboard that
+is a comparability break, not a transport detail.
 
-So name the model the way the gateway names it, and pick the provider so that
-litellm's prefix strip leaves that name intact:
+To pin a route by hand, name the model the way the gateway names it, and pick the
+provider so that litellm's prefix strip leaves that name intact:
 
 ```yaml
 # Gateway route "openrouter/anthropic/claude-opus-4.7"
@@ -637,10 +657,13 @@ couplings described below:
   `extra_body.reasoning`. That is correct for an OpenAI-shaped gateway endpoint,
   but verify it for reasoning models before trusting a run.
 
-**This was a transport swap and nothing else until the gateway route resolution above; routed calls now also carry the gateway's dialect and its route name.** `_build_kwargs` sets `api_base`,
-`api_key`, and `extra_headers`; the litellm model string keeps its original
-`<provider>/<name>` shape. Two distinct couplings hang off model naming, and
-only the second is to that formatted string:
+What `_build_kwargs` does differs by path. **On a resolved route** it rewrites
+`model` to the gateway's route name, forces `custom_llm_provider="openai"`, and
+drops OpenRouter-only body fields (`extra_body.provider`). **On the
+unrouted / unreadable-catalog path** it sets only `api_base`, `api_key` and
+`extra_headers`; the model string keeps its `<provider>/<name>` shape, and
+OpenRouter's headers and provider pin still apply. Two distinct couplings hang off
+model naming, and only the second is to the formatted string:
 
 - Preset `match:` globs resolve off `ModelConfig.name` and the `providers:`
   overlay off `ModelConfig.provider` (see [`presets`](#presets)). Re-prefixing
@@ -653,10 +676,11 @@ only the second is to that formatted string:
   verbatim. A second prefix guarantees a pricing-table miss, degrading
   `cost_source` to `"unknown"` and tripping `Capability.COST_USD_POPULATED`.
 
-Because the provider string is untouched, provider-specific request shaping
-still applies on top of the gateway: OpenRouter's `HTTP-Referer` / `X-Title`
-headers and its `extra_body.provider` upstream pinning both survive. On a
-header-name collision the gateway's configured header wins, since that is
+`ModelConfig.provider` is untouched either way, so OpenRouter's
+`HTTP-Referer` / `X-Title` headers still apply on top of the gateway; the
+`extra_body.provider` upstream pin applies only on the unrouted path (a resolved
+route drops it, since a non-OpenRouter upstream rejects it as an unknown field).
+On a header-name collision the gateway's configured header wins, since that is
 explicit operator configuration and the other is an engine default.
 
 ### Verifying a gateway from CI
--- a/docs/LLM_LAYER.md
+++ b/docs/LLM_LAYER.md
@@ -450,8 +450,9 @@ posting to a route the gateway does not serve.
 | `LLM_PROXY_HEADERS` | JSON object of static headers added to every request, e.g. `{"X-Team-Id": "research"}`. Wins over the engine's own provider headers on a name collision. A value may reference a secret as `${secret:NAME}`, see below. |
 | `LLM_PROXY_REQUEST_ID_HEADER` | Header *name* that receives a fresh UUID4 per request. A static env var cannot express "new value per call". |
 | `LLM_PROXY_PROVIDERS` | Comma-separated provider allow-list, replacing the default. Read the routing table below before widening it. |
+| `LLM_PROXY_PREFERRED_ROUTE` | Namespace that wins when the gateway serves one model under several names, e.g. `openrouter/`. Without it an ambiguous lookup raises rather than guessing a serving path. |
 
-All five resolve through `SecretManager`, so `.env`, the process environment,
+All six resolve through `SecretManager`, so `.env`, the process environment,
 and the runner container's `TOLOKAFORGE_SECRETS_JSON` behave identically.
 A malformed value raises `ProxyConfigError` at the first `LLMClient`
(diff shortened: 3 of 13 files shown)


bberkes-toloka:
## Review

**Direction: architecturally right, and in the right layer.** Speaking the gateway's own dialect and resolving its own route name from `GET /models` is exactly the deferred work the old `proxy.py` docstring named, and it lands where it belongs: wire identity assembled in the transport (`_build_kwargs`), while `ModelConfig` stays authoritative for presets and pricing, with a test pinning that decoupling. The deterministic two-candidate rule, refused ambiguity with an explicit operator override, and the fail-loud posture all match the engine's existing config philosophy, and keeping the serving-path choice with a human is the correct call for leaderboard comparability. I read the branch (not just the diff) at the three wiring points (constructor, `_build_kwargs`, `_call_with_key_rotation`): the rotation guard cannot revert the dialect on retry, the pre-existing gateway-key rotation dead-end composes with it, and the catalog GET carries the same admission headers as completions via `request_headers()`.

Findings below: one behavior bug, one test gap, two operational risks, then minor notes.

### Bug: an empty catalog is negative-cached for the process lifetime

`gateway_route.py:79`. A `{"data": []}` answer maps to `catalog = None` (line 68) and then falls through to `_CATALOG_CACHE[key] = catalog`. Every later construction returns the cached `None` (lines 50-51) and never re-fetches. That contradicts both the cache comment ("Successful reads only, cached ... a failure is transient and caching it would keep a whole run on the degraded path after one blip") and the design argument that an empty answer is "a gateway mid-reload", i.e. transient. Concretely: one mid-reload blip at the first client construction pins the whole process on the untranslated-name path, which for `openrouter/`-prefixed models is the 948-error failure mode this PR fixes. One-line fix: only write the cache when the answer is non-empty (`if catalog is not None: _CATALOG_CACHE[key] = catalog`), or decide the opposite and fix the comment; as written, code and stated policy disagree.

### Test gap: `fetch_gateway_catalog` is untested, and the PR body claims a test that is not in the diff

The body says the empty-catalog behavior "is now pinned by a test against a local endpoint returning `{"data": []}`". No such test exists in this diff (`git grep '"data": \[\]' tests/` on the branch is empty). The only empty-catalog test is resolver-level (`resolve_gateway_route(..., frozenset())`), which never exercises the fetch path. So the entire fetch half (HTTP handling, auth and `LLM_PROXY_HEADERS` propagation, empty-to-None mapping, and the cache semantics, including the bug above) is unpinned; the wiring tests patch it out by design. A small test against a local `http.server` (or monkeypatched `urlopen`) covering success-is-cached / failure-is-not-cached / empty-is-not-cached would close this and would have caught the caching bug.

### Operational risk: a hang-mode catalog outage stalls every client construction

Failures are deliberately not cached and the default timeout is 15s, so a catalog endpoint that hangs (rather than refuses) costs up to 15s per `LLMClient` construction for the whole run, plus a thundering herd when workers construct concurrently (no lock around the fetch; benign for correctness, expensive in wall clock). Consider a short negative-cache TTL or a per-process retry cap: either preserves "one blip must not degrade the whole run" while bounding the pathological case.

### Test isolation: pre-existing proxy tests now do real network I/O

Every `LLMClient(...)` construction in `test_llm_proxy.py` (about 9 sites, `https://gateway.example.com` configured) now performs a real `urlopen` to `.../models` at construction time. They stay green only because that fetch fails, which is exactly the fragility the PR body criticizes in its own words ("the suite was green only because that fetch failed"). On a network with wildcard DNS, or in a sandbox where connects hang instead of refusing, these become 15s-per-test or change behavior. `tests/unit/llm/conftest.py` exists and is empty; an autouse fixture there defaulting `fetch_gateway_catalog` to `None` (the new wiring tests already override it explicitly) would restore isolation cheaply.

### Minor

- `client.py:1762-1767`: routed calls still attach OpenRouter-specific body content when `openrouter.provider_order` is set (`extra_body["provider"]`). Wire-content-wise unchanged from before this PR, but it is the same class of problem as the `usage` extension this PR removes: a bare-name route backed by a non-OpenRouter upstream will reject it as an unknown field. Worth either dropping it when `self._gateway_route` rewrote the target, or a line in the routed-call couplings list in the docs.
- `resolve_gateway_route` docstring says catalog `None` "disables routing"; at the client it means "keep the gateway, skip rewriting". Suggest matching the wording to the client's semantics, since the test file already documents that split precisely.
- `gateway_route.py:90`: `head` is unpacked and never used.
- The alias-miss failure mode is a quiet gateway bypass with only a per-client warning: a route named as a bare alias (e.g. `claude-sonnet-4.6` with no namespace, which LiteLLM operators do use) resolves as "not served" and the call goes direct. The documented two-shape rule is fine for the observed gateway; if a deployment ever needs "gateway or fail", a strict flag is the extension point, not fuzzier matching.

### Two structural watch-items (not blockers)

1. **Network I/O inside `LLMClient.__init__`.** The cost is already visible in the tests: patching two import bindings plus clearing a module-global cache around every test. An injectable fetcher (a parameter, or hung off `ProxyConfig`) would remove both, and it is the same seam the named follow-up (the Slack poller consuming a `ProxyConfig`) needs anyway.
2. **The dialect invariant is enforced in two distant places.** `_build_kwargs` sets it; the guard at `client.py:1951` must not revert it. Mutation tests pin both today; still, assembling the full wire identity (model, `custom_llm_provider`, `api_base`, `api_key`) in one helper would make the invariant unbreakable rather than well-guarded.

Not re-verified by me: the live-gateway runs and the "47 pre-existing failures on main" claim.

azorej:
Review of this branch (mechanism is sound and unusually well-tested — the points below are the problems found, most severe first).

## 1. Blocker — `docs/LLM_LAYER.md` is now internally contradictory, and twice contradicted by this PR's own code

- `docs/LLM_LAYER.md:640` — "**This was a transport swap and nothing else until the gateway route resolution above; routed calls now also carry…**" is migration-history phrasing (Core Rule 8: docs describe current state only, no "was/until/now").
- The same paragraph still claims "the litellm model string keeps its original `<provider>/<name>` shape" — false on the resolved-route path, where `_build_kwargs` rewrites `model` and sets `custom_llm_provider="openai"`.
- Lines ~656-658 say OpenRouter's `extra_body.provider` upstream pinning "survives" — directly contradicted by `extra_body.pop("provider", None)` in `client.py` and by this PR's own test `test_the_provider_pin_does_not_ride_to_the_gateway`.

Suggested rewrite, split by path, current-state only:

> On a resolved route, `_build_kwargs` rewrites `model` to the gateway's route name, forces `custom_llm_provider="openai"`, and drops OpenRouter-only body fields (`extra_body.provider`). On the unrouted / unreadable-catalog path it sets only `api_base`, `api_key`, and `extra_headers`, and the model string keeps its `<provider>/<name>` shape; OpenRouter headers and the provider pin apply there.

## 2. Blocker — § "The model name must be the gateway's route name" still states the pre-PR requirement as current

`docs/LLM_LAYER.md:596-625` opens with "Enabling the gateway is **not** a drop-in for existing run configs … So name the model the way the gateway names it" — which is exactly what this PR retired (the headline verification is an unmodified `openrouter/anthropic/claude-sonnet-4.6` config working through the gateway). Two sections 70 lines apart now give contradictory instructions. Suggest rewriting it as the manual-override companion ("Naming a gateway route explicitly"), keeping the Bedrock anecdote and pricing caveats. Note the anchor is referenced from `docs/AUTO_INTEGRATION.md:193` and `tools/automation/src/automation/gateway_catalog.py:139` — if the heading changes, both need updating in this PR.

## 3. Major — genericity gap: a gateway route can be an arbitrary alias, and then the run silently leaves the gateway

`_candidates()` covers exactly two shapes: the full `<provider>/<name>` string and its one-prefix-stripped tail. But LiteLLM (and most gateways) let operators name a route anything — `model_name: sonnet-4.6`, `team-default`, etc. For an aliased route neither candidate matches, the catalog "answers and omits it", and the call goes **direct to the provider — silently**. That is the unattributed-spend outcome this transport exists to prevent, reached without a warning that distinguishes "gateway genuinely doesn't serve this" from "gateway serves it under a name the resolver can't derive". `LLM_PROXY_PREFERRED_ROUTE` can't help: it only disambiguates among candidates that already matched; it cannot introduce a name.

Deterministic-no-fuzzy-matching is a fine design choice, but there should be an escape hatch — e.g. an explicit per-model route override (env or run-config), or at least documenting the constraint "gateway routes must mirror upstream names" as a hard requirement in `docs/LLM_LAYER.md`.

## 4. Major — `automation.gateway_catalog` now mispredicts and misdescribes engine behaviour

`tools/automation/src/automation/gateway_catalog.py:133-140` (`lookup`) reasons "litellm strips exactly one provider prefix, so … an `openrouter/<slug>` catalog entry is NOT evidence for this run." After this PR that is false: the engine resolves candidates `[openrouter/<slug>, <slug>]`, so an `openrouter/<slug>`-only entry **is** reachable — literally the 948-error case this PR fixes. The Slack poller will now report `absent`/`unknown` for models the engine can serve. The "Not in this PR" deferral covers the `ProxyConfig`/header-admission unification, not this stale claim. Either mirror `_candidates` in `lookup` here (small), or rewrite both docstrings to the actual current contract and file the follow-up.

## 5. Major — new orphan variable not covered by the orphan test

`ENV_PREFERRED_ROUTE` was added to the orphan check in `proxy.py`, but `test_companion_without_base_url_raises` in `tests/unit/llm/test_llm_proxy.py:187-194` still parametrizes only the other four names. Dropping `ENV_PREFERRED_ROUTE` from the orphan tuple leaves the whole suite green. One-line fix:

```diff
         "LLM_PROXY_PROVIDERS",
+        "LLM_PROXY_PREFERRED_ROUTE",
     ],
 )
```

## 6. Minor — failure give-up is permanent for the process, with no recovery

`MAX_CATALOG_ATTEMPTS = 3` consecutive failures permanently stop catalog fetches for the process. Client constructions cluster at run start, so a gateway that is briefly down during warmup burns all three attempts in seconds and pins the entire run on the degraded path (gateway kept, untranslated name — the path that produced the original 948 errors). This is the same shape as the negative-caching bug the third revision of this branch fixed, one level up. A time-based backoff instead of a hard cap would let a long run recover.

## 7. Minor — migration contrast in `proxy.py` module docstring

`proxy.py:85` "Routed calls **now DO** force the OpenAI envelope", `:88` "is therefore **still** not supported" — drop "now DO"/"still", state it plainly.

## 8. Minor — two `GET {base}/models` fetchers with divergent empty-catalog semantics

`gateway_route.fetch_gateway_catalog` treats an empty `data` list as unreadable; `automation.gateway_catalog` returns `[]` as authoritative and classifies everything `absent`. `automation` already depends on `tolokaforge`, so the fetch/parse could be shared even before the `ProxyConfig` unification. Fine to defer with an issue that names the divergence.

## 9. Nits

- `tests/unit/llm/test_gateway_routing_applied.py:464` — the `gateway_route.fetch_gateway_catalog` patch is dead: `client.py` imports the name, so only the `tolokaforge.core.llm.client.fetch_gateway_catalog` patch on the next line is live. The `_isolate` fixture's cache-clears also duplicate the directory autouse fixture.
- `tests/unit/llm/test_gateway_catalog_fetch.py:202` — `# type: ignore[arg-type]` with no one-line justification.
- Latent semantic trap: `resolve_gateway_route` returns `None` both for "catalog omits the model" and for an empty frozenset (`if not catalog`), but the client maps resolver-`None` to "go direct". Safe today only because the fetch maps empty → `None` first; a future call site passing an empty catalog gets "go direct" where the design says "no information — keep the gateway". Worth an assert or a comment at the client split.
- Cache is keyed by `base_url` alone while the fetch authenticates with a specific `ProxyConfig`'s key/headers — two configs sharing a base URL but different credentials would share one cached catalog. Unlikely, noting for the record.

azorej (dismissed):

azorej (approved):

PR #1000, opened 2026-08-10
fix(llm): admit the parameters an operator declares, when litellm's map cannot

balazs-toloka:
## The problem

litellm decides which OpenAI parameters a provider may be sent by looking the model up in its own map, and a **vendor-native** provider answers from that entry alone. A model the map does not carry is read as supporting NO parameters, so `tools` is refused inside our process, before a request is ever sent:

```
litellm.UnsupportedParamsError: meta does not support parameters: ['tools', 'tool_choice'], for model=muse-spark-1.2
```

Measured 2026-08-10, identical request, three litellm versions:

| litellm | `tools` on the `meta` route |
|---|---|
| 1.83.14 | passes |
| 1.93.0 | refused |
| 1.96.0 | refused |

The strictness arrived in a patch release. A routine dependency bump can therefore turn a working vendor-native model into one that cannot make a single tool call, and the error names the provider rather than the missing data, so it reads as "this vendor does not do tool calls".

It is not about the model: the identical request returns a correct tool call once the parameters are admitted. The gap is upstream DATA.

## The change

litellm's own answer is `allowed_openai_params`, a per-call kwarg naming the parameters to admit past the map gating for that one request; its error message says so. This turns an operator's declaration into that list.

`litellm_params.py` carries **no list of models**. A model missing from a third-party map is not a fact about an engine release, and a list in the wheel would tie every future gap to the release cadence — the argument [ADR 0002](docs/adr/0002-external-model-registry.md) already made for preset data, which [ADR-0030](docs/adr/0030-tolokaforge-models-split.md) carries further. So entries are operator data, in the same preset overlay:

```yaml
litellm_models:
  meta/muse-spark-1.2:
    supports_function_calling: true
    supports_reasoning: true       # the config sets models.agent.reasoning
    evidence: "2026-08-10, litellm 1.96.0: no entry, so meta refused tools before
      sending; admitting them returns a correct tool call."
```

`LLMClient.__init__` computes the list once; `_build_kwargs` attaches it when non-empty, so for every model litellm already knows the kwarg is absent and nothing about the request changes.

An entry **declares**; it does not copy. Only the parameters its flags name are admitted, and the allow-list only ever ADDS to what litellm already permits — so an undeclared parameter is still refused loudly, and the declaration is the boundary in both directions. `supports_reasoning` is declarable because a config that sets `models.agent.reasoning` sends `reasoning_effort`, which litellm refuses for an unmapped model exactly as it refuses `tools`; an entry has to cover what its config asks for.

**Nothing is written into litellm's global map.** That keeps three problems from existing rather than solving them: our own price cannot end up labelled `cost_source="litellm"` (the label meaning provider-authoritative), no entry of ours can outlive the day upstream ships a richer one, and there is no process-global mutation to synchronise across the trial thread pool. Once upstream carries the model, the allow-list is a no-op.

Entries are validated at overlay load, which is the CLI boundary, so a malformed one fails before the orchestrator exists rather than as a run of failed trials. Validation is louder than the preset blocks beside it, deliberately: a preset that fails to apply changes how a request is shaped, while a dropped entry here decides whether a request is sent at all.

`config validate` consults the overlay too. It loads and schema-validates the block already, and without this it reported the model unable to call functions while the run worked — a preflight contradicting the thing it checks.

Overlay keys are always `<provider>/<model>`, and the vendor segment is lowercased on both sides of the lookup, so a config and an overlay that disagree about the case of `Meta` still meet. The lookup composes that key from the provider when a config's model id carries no vendor, which keeps the Nova path reachable.

## Verification

- **`TestTheCanonicalContract`** pins what the installed litellm must keep doing, since the premise here is that its patch releases change parameter gating: an unmapped model is refused, naming the parameters admits them, an undeclared name is still refused, and `completion` threads the kwarg end to end (with its paired refusal test). Note `inspect.signature(litellm.completion)` does not list the kwarg on any tested version — it is absorbed by `**kwargs`, so only an actual call proves it.
- The rest of `test_litellm_params.py` covers what an entry admits, that nothing is written into litellm's map, the bare-name path, vendor-case symmetry, and every validation failure.
- `test_config_validator.py` covers the preflight, parametrised over litellm answering `False` and `None`.
- Full engine unit suite: 5866 passed. One pre-existing failure in `test_docker_build_context.py`, which fails identically on `main`.
- `ruff` and `black` clean.

Docs: `docs/LLM_LAYER.md` § "When litellm has never heard of the model".

## Two things that look like fixes and are not

- `drop_params: true` silences the error by stripping `tools`, turning every tool-use trial into a no-tool trial. The eval then measures a configuration mistake and reports it as model capability.
- `extra_body` passes ungated, so a refused parameter can be smuggled through it — and a provider that silently ignores the key is then invisible, which is the same failure wearing a different hat.

A preset `params:` block cannot reach this either: they validate against a closed set introspected from `GenerationParams.__init__`, and the refusal happens upstream of every policy slot.

## What review changed

The first draft mutated `litellm.model_cost`: a model list in the wheel, capability flags cloned from a "witness" sibling, and prices. Review took it apart, and all of it is gone:

- Cloning laundered one observed fact into fifteen. `supports_reasoning` travelled, and litellm then put `reasoning_effort` on the wire for a model nobody had measured for it.
- The price guard failed open: cost keys were overridden but not stripped, so a model absent from our table inherited the sibling's prices.
- The stand-down path stored upstream's dict as if it were ours, so the second client construction registered over the richer upstream row.
- A hardcoded list defeats the seam it sits beside.

The per-call kwarg deletes that whole class of problem — registration, both ledgers, the lock, the stand-down branch, and the test fixture that restored five kinds of hidden litellm global state through a private API — while keeping the schema, the evidence requirement, the allow-list and the validation unchanged.

### DIFF
--- a/docs/ADD_NEW_MODEL.md
+++ b/docs/ADD_NEW_MODEL.md
@@ -88,7 +88,7 @@ don't need to cut a new wheel. Put the same preset entry in a separate YAML
 file and point the engine at it:
 
 ```yaml
-# my_overlay.yaml — same schema as model_presets.yaml
+# my_overlay.yaml — the model_presets.yaml schema, plus `litellm_models:`
 presets:
   my_provider_new_model:
     match: ["my-provider/new-model*"]
@@ -117,6 +117,12 @@ overlay path and the offending key):
   `match:` globs let you shadow a bundled preset.
 - Same-named overlay presets replace the bundled entry (logged at INFO so the
   swap is visible).
+- A `litellm_models:` entry needs a non-empty `evidence` and at least one
+  capability set true, its key must be a full `<provider>/<model>` litellm id,
+  and unknown keys are rejected. That block is how a model litellm's own map
+  does not carry gets its parameters admitted - without it the provider
+  refuses `tools` before the request is sent. See
+  [`docs/LLM_LAYER.md`](LLM_LAYER.md#when-litellm-has-never-heard-of-the-model).
 
 For distributed runs, the overlay path passed to `tolokaforge prepare` is
 persisted into the run queue's state file. Subsequent `tolokaforge worker`
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -348,7 +348,24 @@ Set the overlay path two ways; precedence is **CLI flag > config field**:
      presets_file: ./overlay.yaml   # relative to the working directory
    ```
 
-The overlay file uses the same schema as `model_presets.yaml`. Overlay
+The overlay file uses the schema of `model_presets.yaml` (`default:`,
+`presets:`, `providers:`) plus one block the bundled file has no use for:
+
+```yaml
+litellm_models:                    # models litellm's own map does not carry
+  meta/muse-spark-1.2:
+    supports_function_calling: true
+    supports_reasoning: true       # needed when the config sets models.agent.reasoning
+    evidence: "2026-08-10, litellm 1.96.0: no entry, so meta refused tools
+      before sending; admitting them returns a correct tool call."
+```
+
+Each entry declares what a model accepts on the wire, with the observation
+behind it, and admits exactly the parameters its flags name for that model's
+calls. Nothing else changes, and nothing is written into litellm's global
+map. See [`docs/LLM_LAYER.md`](LLM_LAYER.md#when-litellm-has-never-heard-of-the-model).
+
+Overlay
 presets are prepended to the iteration order so first-match-wins lets you
 shadow a bundled preset. Same-named overlay presets *replace* the bundled
 entry (logged at INFO so the replacement is visible). Policy-name strings
--- a/docs/LLM_LAYER.md
+++ b/docs/LLM_LAYER.md
@@ -23,6 +23,7 @@ for the design rationale and the canonical litellm surface.
 | [`response_policy.py`](../tolokaforge/core/llm/response_policy.py) | Tool-call argument post-processing |
 | [`capabilities.py`](../tolokaforge/core/llm/capabilities.py) | `ModelCapabilities` frozen dataclass |
 | [`presets.py`](../tolokaforge/core/llm/presets.py) | YAML preset loader → `ModelCapabilities`. Also implements the **operator-overridable preset overlay** (`--presets-file`, `engine.presets_file`) so new model registrations don't require an engine release — see [ADR 0002](adr/0002-external-model-registry.md) and [`docs/CONFIG.md` § Preset overlay file](CONFIG.md#preset-overlay-file-no-engine-release-required). |
+| [`litellm_params.py`](../tolokaforge/core/llm/litellm_params.py) | Turns overlay-declared capabilities into litellm's `allowed_openai_params`, so a vendor-native provider does not refuse `tools` for a model its map lacks — see [§ When litellm has never heard of the model](#when-litellm-has-never-heard-of-the-model) |
 | [`proxy.py`](../tolokaforge/core/llm/proxy.py) | Optional LLM-gateway transport (`ProxyConfig`), e.g. a LiteLLM proxy; configured entirely by env |
 | [`client.py`](../tolokaforge/core/llm/client.py) | `LLMClient`, `GenerationResult`, `UserSimulator` |
 
@@ -419,6 +420,92 @@ unchanged — litellm's native `_remove_additional_properties` only runs for
 Vertex AI, hosted vLLM, and WatsonX. Our `StrictSchema` and `DictMapHints`
 policies in `tolokaforge/core/llm/` handle **all** GPT-5 tool-schema
 adaptation independently of litellm, so this gap is transparent to callers.
+
+## When litellm has never heard of the model
+
+litellm decides which OpenAI parameters a provider may be sent by looking the
+model up in its own map. For most providers that decision is generic, but a
+**vendor-native** provider answers from the entry alone. A model the map does
+not carry is therefore read as supporting NO parameters, and the request is
+refused inside our process, before anything is sent:
+
+```
+litellm.UnsupportedParamsError: meta does not support parameters:
+['tools', 'tool_choice'], for model=muse-spark-1.2
+```
+
+Measured 2026-08-10 across litellm versions with an identical request: 1.83.14
+passes the tools through, 1.93.0 and 1.96.0 refuse them. The strictness arrived
+in a patch release, so a routine dependency bump can turn a working
+vendor-native model into one that cannot make a single tool call — and the
+error names the provider rather than the missing data, so it reads as "this
+vendor does not do tool calls".
+
+It says nothing about the model. The identical request driven through litellm's
+`openai` transport against the same `api_base` returns a correct tool call —
+the gap is upstream DATA, and the fix is to supply the entry rather than to
+wait on someone else's release.
+
+litellm's own answer to this is `allowed_openai_params`, a per-call kwarg
+naming the parameters to admit past the map gating for that one request; its
+error message says so. `litellm_params.py` turns an operator's declaration into
+that list, and it ships with **no list of models**. A model missing from a
+third-party map is not a fact about an engine release, and a list in the wheel
+would tie every future gap to the release cadence — the argument
+[ADR 0002](adr/0002-external-model-registry.md) already
+made for preset data. So the entries are operator data, declared in the same
+preset overlay (`--presets-file` / `RunConfig.engine.presets_file`):
+
+```yaml
+litellm_models:
+  meta/muse-spark-1.2:
+    supports_function_calling: true
+    supports_reasoning: true       # the config sets models.agent.reasoning
+    evidence: "2026-08-10, litellm 1.96.0: no entry, so meta refused tools
+      before sending; the same request through litellm's openai transport
+      against api.meta.ai returned a correct tool call."
+```
+
+The key is the litellm model id, because that is the lookup litellm performs.
+An entry **declares**; it does not copy. Only the parameters its flags name are
+admitted, so a capability nothing observed is never asserted on the model's
+behalf.
+
+Two rules keep it honest:
+
+- **The flags are an allow-list.** A parameter stays refused until someone
+  declares the capability with evidence, and extending the map is a decision
+  about what we are willing to assert. `supports_reasoning` is on it because a
+  config that sets `models.agent.reasoning` sends `reasoning_effort`, which
+  litellm refuses for an unmapped model exactly as it refuses `tools`.
+- **Nothing is written into litellm's global map.** The kwarg is per call, so
+  our own price can never end up labelled `cost_source="litellm"` (the label
+  meaning provider-authoritative), no entry of ours can outlive the day
+  upstream ships a richer one, and there is no process-global mutation to
+  synchronise across the trial thread pool. Once upstream carries the model,
+  the allow-list is a harmless no-op.
+
+An undeclared capability is **refused**, not dropped: the allow-list only ever
+ADDS to what litellm already permits, so a config that sets
+`models.agent.reasoning` against an entry that does not declare
+`supports_reasoning` fails per call rather than quietly running without it. An
+entry has to cover what its config asks for.
+
+Validation is at overlay load and is louder than the preset blocks beside it: a
+preset that fails to apply changes how a request is shaped, while a dropped
+entry here decides whether a request is sent at all.
+
+Three things that look like fixes and are not:
+
+- **`drop_params: true`** silences the error by stripping `tools`, turning
+  every tool-use trial into a no-tool trial. The eval then measures a
+  configuration mistake and reports it as model capability.
+- **A preset `params:` block** cannot reach this. They validate against a
+  closed set introspected from `GenerationParams.__init__`, and the refusal
+  happens upstream of every policy slot.
+- **`extra_body`** passes ungated, so smuggling a refused parameter through it
+  works — and a provider that silently ignores the key is then invisible,
+  which is the same failure wearing a different hat.
 
 ## `proxy` — routing calls through an LLM gateway
 
--- a/tests/unit/llm/test_litellm_params.py
+++ b/tests/unit/llm/test_litellm_params.py
@@ -1,0 +1,294 @@
+"""A model missing from litellm's map is refused its parameters before sending."""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import litellm
+import pytest
+import yaml
+
+from tolokaforge.core.llm.litellm_params import (
+    DECLARABLE_FLAGS,
+    FLAG_PARAMS,
+    allowed_openai_params,
+)
+from tolokaforge.core.llm.presets import litellm_model_entries, set_overlay_path
+
+pytestmark = pytest.mark.unit
+
+TOOLS = [
+    {
+        "type": "function",
+        "function": {"name": "f", "parameters": {"type": "object", "properties": {}}},
+    }
+]
+
+#: An id litellm will never carry, so the contract tests below fail only for
+#: the reason they name. Keying them on a real model made them go red the day
+#: upstream shipped it - reporting a broken contract when the good thing
+#: happened. The gating is identical for any id the map lacks.
+MID = "meta/tolokaforge-canary-not-a-real-model"
+NAME = "tolokaforge-canary-not-a-real-model"
+EVIDENCE = "2026-08-10, litellm 1.96.0: no entry, so meta refused tools before sending"
+
+
+@pytest.fixture()
+def overlay(tmp_path: Path):
+    """Install a `litellm_models:` overlay, the way an operator would."""
+
+    def _install(entries: dict) -> Path:
+        path = tmp_path / "overlay.yaml"
+        path.write_text(yaml.safe_dump({"litellm_models": entries}))
+        set_overlay_path(str(path))
+        return path
+
+    yield _install
+    set_overlay_path(None)
+
+
+def _entry(**overrides) -> dict:
+    return {"supports_function_calling": True, "evidence": EVIDENCE, **overrides}
+
+
+def _params(allowed: list[str] | None = None, **kwargs) -> list[str] | None:
+    """What litellm would put on the wire, or None when it refuses the call."""
+    extra = {"allowed_openai_params": allowed} if allowed else {}
+    try:
+        return sorted(
+            litellm.utils.get_optional_params(
+                model=NAME,
+                custom_llm_provider="meta",
+                tools=TOOLS,
+                tool_choice="auto",
+                **extra,
+                **kwargs,
+            )
+        )
+    except litellm.UnsupportedParamsError:
+        return None
+
+
+class TestTheCanonicalContract:
+    """What the installed litellm must keep doing for this design to hold.
+
+    The premise of the whole feature is that litellm patch releases change
+    parameter gating, so the escape hatch it offers is pinned here rather than
+    assumed. Measured across 1.83.14 / 1.93.0 / 1.96.0.
+    """
+
+    def test_an_unmapped_model_is_refused_its_tools(self):
+        assert MID not in litellm.model_cost
+        assert _params() is None
+
+    def test_and_admitted_when_the_call_names_them(self):
+        assert "tools" in (_params(["tools", "tool_choice"]) or [])
+
+    def test_the_allow_list_only_ever_adds(self):
+        """The property the declaration rests on: it cannot loosen anything."""
+        assert _params(["tools", "tool_choice"], reasoning_effort="high") is None
+        assert "reasoning_effort" in (
+            _params(["tools", "tool_choice", "reasoning_effort"], reasoning_effort="high") or []
+        )
+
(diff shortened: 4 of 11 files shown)


CiroGamboa:
Code review — 8 findings. One real correctness bug (#1), one test-coverage gap that hides it (#2), the rest are convention / simplification.

**1. Stand-down path overwrites upstream on the second call — `tolokaforge/core/llm/litellm_map.py:133`**
When a model is already in `litellm.model_cost` and the operator's overlay is redundant, call 1 hits the `model_id in litellm.model_cost and model_id not in _REGISTERED` guard, logs "redundant", stores `_REGISTERED[M] = dict(litellm.model_cost[M])` (upstream's rich dict), returns False. Call 2 — every subsequent `LLMClient` construction, i.e. every worker/trial — compares `_REGISTERED.get(M) == fields` (rich upstream dict vs. our 3-key minimal dict → False), and `model_id not in _REGISTERED` is now False, so the stand-down guard is skipped. We fall through to `litellm.register_model({M: fields})` and **replace upstream's rich entry with our minimal one — prices vanish, `cost_source` stops resolving, upstream-declared flags we didn't list are dropped.** Fix: sentinel-flag the stand-down (e.g. a `_STOOD_DOWN` set, or short-circuit on `model_id in _REGISTERED` before the fall-through) so subsequent calls no-op.

**2. Test misses the second-call regression — `tests/unit/llm/test_litellm_map.py:169`**
`test_a_model_litellm_already_knows_is_untouched` only calls `ensure_model_known` once, so #1 ships green. Add a follow-up call and assert it also returns False and that `litellm.model_cost[witness]` is byte-equal to a pre-call snapshot.

**3. Change-history narrative in docs — `docs/LLM_LAYER.md:460`**
"An earlier draft cloned the flags of a sibling model…" describes a review-pivot, not current behavior. Per AGENTS.md Core Rule 8, "lean code / no change narrative" applies here too. Rewrite as a statement of what the code IS today ("An entry declares; it does not copy — what is registered is exactly what the operator wrote down plus the two structural fields …"), move the pivot story to the PR body.

**4. Same narrative in the module docstring — `tolokaforge/core/llm/litellm_map.py:30`**
Same fix as #3, applied to the module-level docstring.

**5. Dead constant — `tolokaforge/core/llm/litellm_map.py:88`**
`_STRUCTURAL_FIELDS = ('litellm_provider', 'mode')` is defined and cross-referenced in the docstring but never read; `_fields_for` hardcodes the two keys as literals. Either use the constant (`{k: ... for k in _STRUCTURAL_FIELDS}`) or delete it — otherwise a future contributor adding a third structural field changes the tuple and nothing happens.

**6. Validator accepts half-bare model ids — `tolokaforge/core/llm/presets.py:626`**
`'/foo'` and `'foo/'` both contain a `/` but leave one side empty. The validator's own docstring says a bare name never matches; half-bare should fail for the same reason. `provider, _, name = model_id.partition('/'); if not provider or not name: raise ValueError(...)`.

**7. Nova can't use this facility — `tolokaforge/core/llm/client.py:627`**
`_format_model_name` returns the bare `config.name` (no `nova/` prefix) for `provider=='nova'`, but `_validate_litellm_models` requires overlay keys of the form `<provider>/<model>`. An operator adding `litellm_models: {'nova/muse-spark-nova-1': {...}}` will pass validation, then `ensure_model_known('muse-spark-nova-1')` will look up the entry keyed `'nova/muse-spark-nova-1'` and get `None` — silent no-op. Either normalize the lookup key on the Nova path or align `_format_model_name` with the overlay key shape.

**8. Contradictory operator logs — `tolokaforge/core/llm/litellm_map.py:145`**
"Declared %s to litellm as %s" fires on the buggy overwrite path in #1, so an operator sees "litellm already carries M; overlay entry is redundant" followed by "Declared M to litellm as …" for the same model in the same process. Resolves once #1 is fixed; if the log stays where it is, worth pulling it into each branch so each write path names its own semantics.

Overall the shape is right — hand-declared capability flags for missing models is the right seam. #1 + #2 are the only ones I'd block merge on; the rest are cleanups.

bberkes-toloka:
All eight addressed. Thank you for #1 in particular — it was real, and worse than a stale ledger entry: I reproduced it before fixing.

```
upstream keys before: 24 | has price: True
call 1 -> False
call 2 -> True          <- fell through and registered over upstream
upstream keys after : 101
```

**#1 stand-down clobber.** Two ledgers now, because they answer different questions: `_REGISTERED` is what we wrote, `_UPSTREAM_OWNED` is what we must not write. The stand-down adds to the second and never stores upstream's dict as if it were ours, so call 2 and every later one short-circuit.

**#2 test gap.** `test_a_model_litellm_already_knows_is_untouched` now calls three times and asserts the entry is byte-equal to a pre-call snapshot. Verified by mutation: with the fix reverted the test fails, with it restored it passes.

**#3 / #4 change narrative.** Removed from both the module docstring and `LLM_LAYER.md`; they now state what the code is. The pivot story lives in the PR body and the commit message.

**#5 dead constant.** `_STRUCTURAL_FIELDS` deleted along with its docstring cross-reference. `_fields_for` is four lines and reads fine without the indirection.

**#6 half-bare ids.** `partition` now, so `/foo`, `foo/`, `/` and a bare name all fail the load. Parametrised test over the four.

**#7 Nova.** Rather than change what `_format_model_name` sends litellm, the lookup composes the overlay key from the provider: `ensure_model_known(model_name, provider)`. Overlay keys stay one shape — `<provider>/<model>` — and a bare model id is reachable. Two tests: found via the provider, and not found without one.

**#8 contradictory logs.** Each write path now logs inside its own branch, and the contradiction is gone with #1.

Full engine unit suite: 5862 passed (the one pre-existing `test_docker_build_context` failure deselected; it fails identically on `main`). `ruff` and `black` clean.

One thing worth flagging that your review surfaced indirectly: `ruff format` and `black` disagree on eight unrelated files under `tests/unit/llm/`. I had run ruff over the directory and pulled that churn into the diff; it is reverted, and the diff is five files again. Whether the repo wants both formatters pointed at the same files is a separate question.

azorej:
### On the open question: we probed both routes, and the answer is a third one

The PR ends on the right question — whether to route `meta` through litellm's `openai` transport instead of patching the map. We took a deeper pass at what that transport-override version would concretely look like. It is disqualified. But the probing surfaced a better option than either design, and it's a small refactor of this PR rather than a rival to it.

#### 1. The transport override re-imports the same gap for reasoning — it's out

Probed on litellm 1.93.0 (the dev-resolved version), unknown model via `custom_llm_provider="openai"`:

| request | result |
|---|---|
| `tools` + `tool_choice` | passes |
| `reasoning_effort` | **refused** — `UnsupportedParamsError: openai does not support parameters: ['reasoning_effort'], for model=muse-spark-1.2` |
| `parallel_tool_calls` | passes, ungated |

The `openai` transport gates `reasoning_effort` through `litellm.supports_reasoning(model)` — a lookup into **the same map** this PR patches. The worked example declares `supports_reasoning: true` precisely because its config sets `models.agent.reasoning`, so the transport route fixes the headline symptom and reproduces the bug for the capability the config actually needs. Rescuing it means smuggling `reasoning_effort` through `extra_body` (which passes ungated), where a provider silently ignoring the key becomes invisible — the exact "measure a configuration mistake and report it as capability" failure this PR refuses. Row 3 shows the transport is also *looser* than declared: the allow-list property is lost in both directions. Add the build-out costs — explicit key threading (litellm would look up `OPENAI_API_KEY`), precedence rules against `proxy.py`'s gateway `api_base` and the OpenRouter `custom_llm_provider` branch, and a permanent bypass with no stand-down when upstream ships the entry — and there is nothing left arguing for it.

#### 2. What the probing surfaced: litellm's sanctioned per-call escape hatch

litellm accepts `allowed_openai_params` as a per-call kwarg — param names admitted past the map gating for that one request. Verified on the **vendor-native `meta` transport**:

```python
get_optional_params(
    model="muse-spark-1.2", custom_llm_provider="meta",
    tools=TOOLS, tool_choice="auto", reasoning_effort="high",
    allowed_openai_params=["tools", "tool_choice", "reasoning_effort"],
)
# passes -> ['extra_body', 'reasoning_effort', 'stream', 'tool_choice', 'tools']
```

`completion` threads it end-to-end, it's stripped before the request body is built, and litellm's own `UnsupportedParamsError` message instructs: *"If you want to use these params dynamically send allowed_openai_params=[...] in your request."* This is the vendor-intended API for exactly this situation. Registering entries in `model_cost` is fighting litellm; this is using it.

Concretely:

- **Survives unchanged:** the `litellm_models:` overlay block, the `evidence` requirement, the flag allow-list, all of `presets.py`'s loud CLI-boundary validation, the docs argument. The design's spine — an entry *declares*, with evidence, and nothing undeclared travels — is untouched.
- **Changes:** `litellm_map.py` shrinks to a pure function mapping declared flags to param names (`supports_function_calling` → `tools`, `tool_choice`; `supports_reasoning` → `reasoning_effort`; `supports_parallel_function_calling` → `parallel_tool_calls`). `LLMClient.__init__` computes the list once; `_build_kwargs` attaches it when non-empty.
- **Deleted outright:** `litellm.register_model`, `_REGISTERED`, the lock, the stand-down logic — and with them the second-call clobber the re-review found. We traced it independently; it's real, with one nuance: 1.93.0's `register_model` *merges* rather than replaces, so the second call overwrites upstream's same-named flags rather than the whole entry. Merge-vs-replace is itself version-dependent across litellm releases, which is its own argument for not building on it. Also gone: the autouse fixture restoring five kinds of hidden litellm global state (`model_cost`, per-provider model lists, the lowercase-id map, two LRU caches, via a private API), and the thread-safety machinery — per-call kwargs need none of it.
- **Properties preserved:** undeclared params are still refused loudly (the allow-list only *adds* to the supported set); no prices ever enter litellm, so `cost_source` stays honestly `local`. "Upstream always wins" stops needing logic: once upstream ships the entry, the allow-list is a harmless no-op.
- **New obligation:** a canonical test pinning that the installed litellm accepts the kwarg, same style as `test_llm_gateway_envelope_contract.py`.

One caveat, stated honestly: per-call kwargs don't fix other consumers of litellm's map (`get_model_info`, token counting). The engine has exactly one — which brings us to:

#### 3. A latent gap in the PR as shipped: preflight contradicts the overlay

`config_validator.py:261-267` calls `litellm.supports_function_calling(model)` — which returns `False`, not an exception, for a map-missing model (verified) — and for a non-OpenRouter provider emits a hard **ERROR**: *"Model does not appear to support function calling (required for agent)"*. `tolokaforge config validate` never constructs an `LLMClient`, so `ensure_model_known` never fires there. The command even accepts `--presets-file` and schema-validates the `litellm_models:` block — then declares the model unable to call functions. Preflight says broken; the run works. Untested in this PR, and fixable identically under either design: the validator consults `litellm_model_entries()` before trusting litellm's map.

#### Requested changes

1. Keep the overlay schema, validation, tests-of-rationale, and docs.
2. Replace the registration machinery with `allowed_openai_params` in `_build_kwargs`.
3. Teach `config_validator` about `litellm_model_entries()`.
4. Add the canonical pin for `allowed_openai_params` on the installed litellm.

The registration design should not merge as-is: it carries a confirmed ordering bug, depends on version-unstable `register_model` semantics in a PR whose entire premise is that litellm patch releases change behavior, and the per-call variant deletes that whole class of problem while keeping everything else this PR got right.

bberkes-toloka:
All four requested changes are in. You were right, and the rewrite deletes more than it adds.

I verified the load-bearing claim before rewriting around it, on **both** 1.93.0 and the 1.96.0 the eval legs resolve:

```
  nothing allowed                              UnsupportedParamsError
  allowed=[tools,tool_choice]                  OK   ['extra_body','stream','tool_choice','tools']
  allowed=[...]+reasoning_effort               OK   [..., 'reasoning_effort', ...]
  UNDECLARED reasoning_effort still refused?   UnsupportedParamsError
```

One correction to the write-up, in your favour: `inspect.signature(litellm.completion)` does **not** list `allowed_openai_params` on either version, so a signature check reads as "not threaded". It is absorbed by `**kwargs` and forwarded — confirmed by an actual `completion(..., mock_response=...)` call, which raises without the kwarg and returns with it. Worth knowing because the signature is the first place anyone will look.

**1. Schema, validation, docs kept.** The overlay block, the `evidence` requirement, the flag allow-list and all of `presets.py`'s CLI-boundary validation are untouched.

**2. Registration replaced.** `litellm_map.py` is gone; `litellm_params.py` is a pure map from declared flag to parameter names. `LLMClient.__init__` computes the list once, `_build_kwargs` attaches it when non-empty. Deleted with it: `register_model`, both ledgers, the lock, the stand-down branch, the second-call clobber, and the autouse fixture that restored five kinds of hidden litellm global state through a private API. The test file lost 60 lines and gained coverage.

**3. `config_validator` teached.** It consults `litellm_model_entries()` before trusting the map, so the command that loads and schema-validates the block no longer declares the model unable to call functions. A broken overlay still fails on its own louder path rather than becoming a function-calling verdict.

**4. Canonical pin added.** `TestTheCanonicalContract` pins four properties on the installed litellm: an unmapped model is refused, naming the parameters admits them, an undeclared name is still refused, and `completion` threads the kwarg end to end.

Also re-ran the end-to-end path (real `RunConfig`, the CLI's own overlay activation, a real `LLMClient`, from a fake tasks-repo root):

```
5. params the client will admit     -> ['tools', 'tool_choice', 'reasoning_effort']
6. tools with that allow-list       -> ['extra_body', 'reasoning_effort', 'stream', 'tool_choice', 'tools']
7. litellm's global map untouched?  -> True
```

Full engine unit suite 5859 passed (the pre-existing `test_docker_build_context` failure deselected). `ruff` and `black` clean.

Your point about `extra_body` is now in the docs beside `drop_params`, as the second thing that looks like a fix and is not — same failure, different hat.

CiroGamboa:
Re-reviewed the rewrite (`50a5de6a`). The per-call `allowed_openai_params` approach is a much cleaner shape than mutating `litellm.model_cost` — all 8 findings from the first pass are resolved (stand-down overwrite gone with the deleted module, validator now rejects half-bare ids with a parametrized test, Nova path handled via `_entry_key`, "earlier draft" narrative rewritten as measured evidence). Nice pivot.

Two minor asymmetries I noticed in the new code, both low-impact:

**A. Case-normalization mismatch — `tolokaforge/core/llm/litellm_params.py:_entry_key`**
The bare-name / Nova branch lowercases the provider to build the lookup key (`vendor = (provider or "").strip().lower()`), but `_validate_litellm_models` accepts overlay keys in any case (`"Meta/muse-spark-1.2"` validates the same as `"meta/muse-spark-1.2"`). An operator who happens to capitalize the vendor in the YAML would silently fail to match on the Nova path. Two symmetric fixes:
- Normalize `model_id.partition("/")[0]` to lowercase at overlay load and store the entry under the lowered key.
- Or reject non-lowercase providers in the validator ("litellm's ids are lowercase; …").

Not urgent — nobody writes `Meta/…` by hand — but a validator that accepts what the lookup won't find is the exact seam that produces "the overlay is loaded but the model still refuses tools" reports.

**B. `_declared_function_calling` only overrides on `False`, not `None` — `tolokaforge/core/config_validator.py:280`**
```python
fc_support = _model_supports_function_calling(litellm_name)
if fc_support is False and _declared_function_calling(litellm_name):
    fc_support = True
```
The observed failure mode is litellm returning `False` for the unmapped model (that's what the commit body pins to 1.93.0/1.96.0). If a future litellm patch instead returns `None` (unknown) for the same shape, the overlay declaration is silently not consulted and preflight regresses to "unknown". Widening the guard to `if fc_support is not True and _declared_function_calling(litellm_name):` keeps the preflight consistent with the run under both mapping outcomes. Not a bug today — a stability hedge against the same class of upstream drift this PR is answering.

Both fine to defer; flagging so they're visible.

bberkes-toloka:
Both taken — they were cheap, and B is the same class of drift the PR exists to answer, so deferring it felt wrong.

**A. Case.** Fixed on both sides rather than by picking one: the vendor segment is lowercased at overlay load (`_lower_vendor_keys`) and in `_entry_key`, so a config and an overlay that disagree about `Meta` still meet. I chose normalising over rejecting because rejecting fails a file that says exactly what it means, and litellm's ids are lowercase anyway.

Only the vendor is normalised — the model name keeps its case, since that half is the vendor's business, and a test pins that too.

Verified by mutation: removing the lowering in `_entry_key` fails
`test_the_vendor_case_does_not_decide_whether_it_matches[meta/… - Meta/…]` and nothing else. Five cases, including the bare-name path (`Nova/…` declared, `NOVA` asked).

**B. `is not True`.** Widened, with the reason in a comment next to it: today an unmapped model reads `False`, and the premise of this feature is that litellm's answers move between patch releases. Parametrised over `False` and `None`, asserting neither produces a function-calling issue when the overlay declares the model.

Full engine unit suite 5866 passed, `ruff` and `black` clean.

Thanks for both passes — the second-call clobber in particular was worth catching before this went near a run.

azorej:
### Verified the rewrite at `50a5de6a` — the pivot delivers. Two must-dos before undraft, and one bot finding to disregard.

All four requested changes are in and I re-traced each against the code rather than the thread. `litellm_params.py` is a pure map with no litellm import, no global state, no lock; the kwarg attaches only when non-empty; `config_validator` consults the overlay; and `TestTheCanonicalContract` is exactly the right pin — especially `test_completion_threads_the_kwarg_end_to_end` with its paired refusal test. Good catch on the `**kwargs` absorption: verifying by actual call instead of `inspect.signature` is the correct instinct, and worth the comment you gave it.

#### Must-do 1: the stale doc sentence (Claude re-review #1 — confirmed)

`docs/LLM_LAYER.md` § "When litellm has never heard of the model" still says *"What is registered is exactly what the operator wrote down, plus the provider and `mode: chat` litellm needs to route the call"*. Nothing is registered anymore; the bullet two paragraphs down says so itself. Dead-design text, same rule the module docstring was already held to.

#### Must-do 2: the PR title and body describe the deleted design

Nobody has flagged this yet: the description still reads *"What is **registered** is exactly…"*, *"**Registration is lock-guarded** (trials run in a thread pool…)"*, the Verification section counts stand-down, re-registration, and thread-safety tests that no longer exist, and the "Open question" section is now settled in the code. A squash-merge bakes this into history as the change's permanent explanation of itself. Rewrite it to describe the `allowed_openai_params` design before undraft — the same "state what the code is" standard the docs and docstrings were held to applies to the merge commit's source text.

#### Disregard: Claude re-review finding #2 (reasoning preflight)

The claim was that `_model_supports_reasoning` "returns `False` for a map-missing model" and warns despite a declared `supports_reasoning: true`. Traced it: that function is **prefix-heuristic based** (`_REASONING_SUPPORTED_PREFIXES` / a hardcoded unsupported list) and never consults litellm's map. For `meta/muse-spark-1.2` it returns `None`, and the validator emits INFO ("cannot confirm"), not a WARNING. There is no contradiction to fix — don't spend time on it. (If you want polish there someday: an overlay-declared `supports_reasoning` could upgrade that INFO to a confirmed pass, and would also cover the edge where a model matches a hardcoded-unsupported prefix while the overlay declares otherwise. Follow-up material, not this PR.)

#### CiroGamboa's A and B

- **A (case mismatch in `_entry_key`)** — real, agreed it's the "validator accepts what the lookup won't find" seam. Fine as a follow-up; lowercasing the provider segment at overlay load is the fix that keeps one canonical key shape.
- **B (`is False` vs `is not True`)** — cosmetic today: when litellm returns `None`, no ERROR is emitted on either path, so the widened guard changes nothing observable. Skip or defer freely.

With the doc sentence and the description rewrite done, this is ready to undraft and merge from where I stand.

bberkes-toloka:
Both must-dos are in.

**1. The stale doc sentence.** Gone — it now reads "Only the parameters its flags name are admitted", matching the module docstring. I re-grepped the section for the whole dead vocabulary (`registered`, `registration`, `register_model`, `lock-guarded`, `stand-down`, `witness`) and the only hit left is the pre-existing `presets.py` row in the module map, which uses "model registrations" in the ADR-0002 sense.

**2. The description.** Title and body rewritten to describe the shipped design. Gone with it: the registration/lock/stand-down language, the Verification section's count of tests that no longer exist, and the "Open question" section — settled in the code, and leaving it open in the merge commit's source text would be the same lie in a different place. The measured version table stays, because it is still what motivates the feature. The review pivot is now one section at the end, stated as what changed rather than as how the code works.

Thanks for tracing the reasoning-preflight claim rather than passing it on — I would have spent time on it. Noted the follow-up idea (an overlay-declared `supports_reasoning` upgrading that INFO, and covering the hardcoded-prefix edge); not in this PR.

On CiroGamboa's A and B: both are already in as of `b693c49`, before your comment landed. A because "the validator accepts what the lookup won't find" is the seam that produces the one report nobody can act on, and I fixed it on both sides — lowercased at overlay load and in `_entry_key` — with a mutation check confirming only the case test fails when the lowering is removed. B because it costs nothing and the failure it hedges against is the same upstream drift this PR answers; you are right that it changes nothing observable today.

Ready to undraft from my side.

CiroGamboa (approved):
LGTM

azorej (approved):

Issue #939, opened 2026-08-07
feat(automation): retarget integrate-model workflow to tolokaforge_models/ for Bucket A; invert ADD_NEW_MODEL.md


CiroGamboa:
## Context

The Slack-triggered auto-integration pipeline today commits to the engine branch even for pure-data model changes — the piece of ADR-0002's original critique that remained active until now. ADR-0030 § Auto-integration workflow retargeting makes repointing the pipeline for Bucket A the change that puts the *"engine version stays put across a lineup refresh"* property under real-integration pressure. It lands last on purpose: the manual `tolokaforge_models/` path is proven end-to-end by a real model integration first.

## Deliverables
- [ ] `.github/workflows/integrate-model.yml` classifies the resolved candidate as Bucket A (only existing base classes needed) or Bucket B (new base class or new lifecycle stage) and commits to `tolokaforge_models/` on Bucket A
- [ ] Bucket B falls through to the existing engine PR flow unchanged
- [ ] `docs/ADD_NEW_MODEL.md` pre-flight table is inverted: Bucket A default routes to `tolokaforge_models/`; Bucket B routes to `tolokaforge/`
- [ ] `docs/LLM_LAYER.md` linked from the flipped guide
- [ ] Classification step tested against synthetic Bucket A and Bucket B inputs

## Acceptance Criteria
- A new-model Slack request that only needs existing base classes + a new certificate + a new preset lands as a PR against `tolokaforge_models/` — no engine changes, no engine release
- A new-model Slack request that requires a genuinely new base class or lifecycle stage lands as an engine PR (Bucket B path)
- The first live Bucket A integration after this lands preserves the engine version (the acceptance-test metric from #932 reflects this)

## Design Decisions
- **Classification is deterministic, not heuristic.** Bucket A = the resolved preset uses only shipped base classes + shipped extension slots; Bucket B = anything else. The workflow reads the resolved policy graph, not string-matches the request.
- **Land after the first manual Bucket A integration.** ADR-0030 is explicit about proving the manual path end-to-end before automating it; automation reveals mismatches, but doing so on a proven path bounds the debugging surface.
- **Docs flip lands with the workflow retarget.** Splitting them across two PRs creates a window where the doc says "target `tolokaforge_models/`" and the workflow says "target engine".

## Out of Scope
- Any further widening of Bucket A (deferred to future ADRs)
- The retarget for downstream repos (this is engine-repo automation only)

## Reference
[...]

Issue #938, opened 2026-08-07
feat(models-wheel): create tolokaforge-models package and cut over the engine loader


CiroGamboa:
## Context

THE structural PR of the split arc. Every preceding ticket lands preparatory work in the engine (public API, extension points, provider-bindings-as-data, consumer migration). This PR creates the `tolokaforge-models` wheel, moves the data + per-model subclasses + certificates + per-model test bodies into it, wires the engine to load through a single accessor module, and installs the fail-loud install-time guardrails. All previous children in this umbrella are prerequisites.

## Deliverables
- [ ] `tolokaforge_models/` top-level directory holds preset routing, pricing, provider bindings, vendor-inference table, capability certificates, per-model policy subclasses, per-model certification bodies
- [ ] `tolokaforge_models` declares `__api_version__` (integer) and `minimum_engine_version` (PEP 440 spec) at module level
- [ ] Hatch custom builder produces the `tolokaforge-models` wheel from the monorepo; installable via `pip install tolokaforge-models`
- [ ] Engine `pyproject.toml` declares `tolokaforge-models >= 1.0.0` as a runtime dependency
- [ ] `tolokaforge/core/model_data.py` is the single accessor the engine uses to reach models-wheel content; nothing else in engine code paths imports `tolokaforge_models` directly
- [ ] Install-time validation lands: `__api_version__` compatibility check, `minimum_engine_version` runtime check (deliberately not `Requires-Dist`), and class-name-existence check at the earliest safe import point (after `presets.py` top-of-module imports, or in `prepare`) — not at `LLMClient` construction
- [ ] `tolokaforge/core/llm/__init__.py` ships a one-release re-export shim with `DeprecationWarning` for every per-model subclass name that moved
- [ ] Runner-subset builder gains `tolokaforge-models` as a pip dep; the runner-subset partition drops the two data files
- [ ] `.github/workflows/release-models.yml` and `.github/workflows/publish-tolokaforge-models.yml` land (publish stays dry-run until the PyPI project is registered — that manual step is owner-only)
- [ ] `docs/adr/README.md` and `docs/RELEASING.md` name the new artefact

## Acceptance Criteria
- New canonical `test_models_subset_partition.py` (mirrors `test_runner_subset_partition.py`) locks the `tolokaforge_models/` tree against a partition manifest
[...]

Issue #937, opened 2026-08-07
feat(downstream): migrate importlib.resources readers off tolokaforge.core.data before the cutover


CiroGamboa:
## Context

The known downstream consumer `benchmark-results-collector` reads pricing.json via `importlib.resources.files("tolokaforge") / "core/data/pricing.json"` and returns an empty dict on failure — a silent cost-zero regression when the resource moves. ADR-0030 § Downstream data-resource consumers explicitly drops the forwarding-stub idea (no attachment point for a `DeprecationWarning`; a build-time stub goes silently stale on the first independent models release) in favour of migrating consumers up-front and failing loud on empty.

## Deliverables
- [ ] `benchmark-results-collector` migrates to `tolokaforge.core.model_data.bundled_pricing_path()` and fails loud on missing/empty pricing rather than returning `{}`
- [ ] Codebase-wide sweep for `importlib.resources.files("tolokaforge")` readers touching data paths (including the runner-subset's documented lookup); each identified reader either migrates or is documented as intentionally untouched
- [ ] Migration guidance is added to `docs/RELEASING.md` for any future downstream consumer

## Acceptance Criteria
- Every reader of engine bundled pricing / preset data through `importlib.resources.files("tolokaforge")` has been identified; each has either migrated to the accessor or is documented as staying on the resource path with justification
- The migrated `benchmark-results-collector` fails loud (not empty) when pricing is unavailable — verified in its own test suite
- The sweep leaves zero silently-swallowing readers of `tolokaforge/core/data/pricing.json` or `model_presets.yaml`

## Design Decisions
- **Fix consumers first, then move; no stub.** A build-time forwarding stub in the engine would ship stale prices after the first independent models release — worse than today's loud-empty failure. Migrating consumers ahead of the move keeps the failure mode legible.
- **Fail loud on empty is the safer default.** Empty pricing surfaces as zero-cost columns in leaderboards, which is a silent-wrong outcome; a clear error at load time forces the operator to notice.

## Out of Scope
- The actual data relocation to `tolokaforge_models/data/` — cutover PR
- Retroactively fixing consumers outside the Toloka org — a courtesy note goes in release notes

## Reference
Part of #645. Design authority: [ADR-0030](https://github.com/Toloka/tolokaforge/blob/main/docs/adr/0030-tolokaforge-models-split.md) § Downstream data-resource consumers. @bberkes-toloka has volunteered to land the `benchmark-results-collector` migration.

[...]

Issue #936, opened 2026-08-07
feat(llm): promote module-private helpers to public API; land base-class hooks per-model subclasses need


CiroGamboa:
## Context

Every per-model subclass shipped since the OSS cleanup either reaches into module-private helpers (`_coerce_json_strings`, `_coerce_empty_containers`, `_find_additional_properties`) or requires a base-class hook to be added ad-hoc (`StrictSchema.carry_scalar_dict_map_value` for the Gemini recursive-schema case). Both are engine-code touches today, which is exactly what keeps those subclasses Bucket B. Per ADR-0030's design principle — *"everything a per-model adaptation needs is public, documented API with a compatibility guarantee"* — this ticket makes the promotion so the cutover PR can move the subclasses to the models wheel cleanly.

## Deliverables
- [ ] Every module-private helper reached by a currently-shipped per-model subclass becomes public API with docstrings and a compat guarantee
- [ ] `StrictSchema` (and any similar base whose hook was added for a specific per-model case) gains the hook as a documented named parameter with a default that preserves current behaviour
- [ ] An import-scope canonical test rejects `_`-prefixed imports of engine LLM-layer symbols from anywhere that will host per-model subclasses in the cutover
- [ ] `RefResolvingDictMapHints._build_hints` scope call is resolved — either the base's `@staticmethod → instance method` shape change lands as a companion here, or `RefResolvingDictMapHints` is explicitly kept engine-side with the reason documented

## Acceptance Criteria
- Every currently-shipped per-model class can be re-implemented against only the public API — verified by a canonical test that re-imports each subclass under a rewriting import hook rejecting private-symbol references
- Every existing per-model subclass continues to behave byte-identically (unit tests untouched)
- Adding a new per-model subclass that inherits from a stable base + uses only public API does not require any base-class edit

## Design Decisions
- **Promotion is engine work; the move is models work.** This PR lands the API contract; the cutover PR moves the subclasses. Landing the promotion first means the cutover diff is strictly a relocation, not a promotion-plus-relocation.
- **Base-class hook policy per ADR.** Hooks land with default values that preserve current behaviour, so the engine PR is observationally a no-op on every currently-shipped model.
[...]

PR #1058, opened 2026-08-11
feat(core): Milestone 29 — tolokaforge-models split (ADR-0030 delivery)

### LINKED ISSUE #645: [design] Decouple model integration from the engine

CiroGamboa:
## TL;DR

Milestone 29 implements ADR-0030 across ten sub-issues (#930–#939). The tolokaforge engine splits into two independently-versioned PyPI wheels: `tolokaforge` (base classes, runtime, orchestration) and `tolokaforge-models` (data, per-model policy subclasses, certificate registry). The engine no longer needs a release for a model-only change — new pricing, new preset, new certificate, and new per-model policy subclass all land in the models wheel, versioned independently. Behaviour-locked byte-identical across the cutover (`sha256({presets, pricing, certificates}) = 29b3ee7c…` invariant); TestPyPI publish verified end-to-end (run `31510609678`).

Closes #645, #930, #931, #932, #933, #934, #935, #936, #937, #938, #939.

**Compatibility surface:** No behaviour change to shipped models. The engine still ships every pre-cutover model on every existing preset; the same fingerprint payload subset (presets + pricing + certificates) hashes byte-identical across the cutover. The models wheel is now a transitive dep of the engine wheel — `pip install tolokaforge` also brings `tolokaforge-models>=1.0.0,<2.0.0` from PyPI once the first models-v1.0.0 tag publishes.

**Deprecation window (v0.18.0):**
- `from tolokaforge.core.llm import GeminiSchema` (and the seven other moved per-model subclasses) still resolves via a per-name `DeprecationWarning`-emitting shim. The shim removes in v0.18.0; downstream consumers should migrate imports to `tolokaforge_models.policies.<family>`.
- The `MODELS_PACKAGE_VERSION` and `MODELS_MINIMUM_ENGINE_VERSION` engine constants are gone; readers now import `tolokaforge_models.__version__` and `tolokaforge_models.minimum_engine_version` at runtime.

**Safety guards active from this release:**
- `_check_minimum_engine_version()` fires at every `presets.py` import. Missing models wheel → `RuntimeError("tolokaforge requires tolokaforge-models >= 1.0.0. Install with pip install tolokaforge-models.")` with chained `ImportError`.
- `_check_class_names_resolve()` walks the merged `_POLICY_REGISTRIES` against every preset's `default`/`presets`/`providers` block. Unknown class name → `RuntimeError` naming preset + slot + missing class + closest-match suggestion.
- Silent-fail paths eliminated: `_load_pricing` and `_load_bundled_presets` raise `FileNotFoundError` / `ValueError` instead of returning `{}` / falling back to defaults.

**Follow-ups filed:**
- Benchmark-results-collector migration (out-of-repo) — stays with @bberkes-toloka, lands ahead of first `models-v1.0.0` publish.
- Auto-integration workflow display-text sweep across ~12 docs outside the canonical Bucket-A/B path (low-value tidy, not urgent).
- Pre-existing hangs (`test_run_trial_composition.py`, `test_ungradeable_trial_accounting.py`) — surfaced during milestone validation, filed as separate follow-ups; not caused by this milestone.

## Design walkthrough

```mermaid
flowchart LR
    subgraph pre["Pre-cutover — engine wheel carries everything"]
        E1["<b>tolokaforge/</b><br/>core/data/pricing.json<br/>core/data/model_presets.yaml<br/>core/data/providers.yaml<br/>testing/certify/_registry.py (39 certs)<br/>core/llm/{schema_sanitizer,prompt_policy,<br/>response_policy,reasoning_codec}.py<br/>+ 8 per-model subclasses inline"]
    end

    subgraph post["Post-cutover — two independently-versioned wheels"]
        direction TB
        E2["<b>tolokaforge/</b> (v0.17.0)<br/>base classes + runtime<br/>_DATA_ROOT = tolokaforge_models/data<br/>load_policy_registrations() merge-loader<br/>_check_minimum_engine_version() gate<br/>_check_class_names_resolve() gate"]
        M2["<b>tolokaforge-models/</b> (v1.0.0)<br/>src/tolokaforge_models/data/*.yaml,*.json<br/>src/tolokaforge_models/certificates/registry.py<br/>src/tolokaforge_models/policies/{gemini,minimax,deepseek,inkling}.py<br/>[project.entry-points.tolokaforge.policies]"]
        E2 -- "importlib.metadata.entry_points<br/>+ importlib.resources.files" --> M2
    end

    pre -- "#938 the load-bearing<br/>one-line _DATA_ROOT flip" --> post
```

The eight prior sub-issues (#930–#937) built the seams that #938's one-line flip actually moves the data. #939 retargets the Slack-triggered auto-integration workflow to commit to `tolokaforge_models/` on Bucket A (data-only) changes and inverts `docs/ADD_NEW_MODEL.md` so Bucket A is the documented default.

## Key design choices

| Decision | Rationale |
|---|---|
| **`_DATA_ROOT: Final[Path]` module constant** as the single seam | One line changes in #938 — the three accessor bodies (`bundled_pricing_path`, `bundled_presets_path`, `bundled_providers_path`) stay stable across v0.17.x per compat guarantee. Monkey-patchable in tests via `monkeypatch.setattr`. |
| **Two-module split — `model_data.py` (light seam) + `model_data_fingerprint.py` (orchestrator-only sibling)** | The `test_subset_files_do_not_import_orchestrator_only_siblings` AST guard rejects orchestrator-only imports from any subset-included file. Adding `model_data.py` to the runner subset required its heavy first-party imports (`compute_models_fingerprint`, `_certificate_to_dict`, `_capability_key`) to move to a sibling. |
| **`Capability` enum + `ModelCertificate` dataclass stay engine-side** (revised ADR 2026-08-07) | The models wheel imports these types; the engine owns them. This is the load-bearing rule that lets Bucket A cert-only PRs skip an engine release without breaking type safety. |
| **`ModelCertificate` widened additively — four new default-empty fields** (`excluded_capabilities`, `known_unsupported_reasons`, `probe_params`, `capability_extras`) with `MappingProxyType` freeze + custom `__hash__` skipping Mapping fields | No existing cert edits required; `dataclass(frozen=True)`'s hashability contract survives the widening. |
| **`@register_probe` shipped as forward-compat seam without shared-body dispatch** | The strict-dispatch reading of ADR-0030 would have required refactoring ~30 shared bodies whose subtle bugs would break byte-identical acceptance. API is not dead code; behaviour-lock unit test exercises the registry. |
| **`ProviderBinding` Pydantic frozen schema for provider transport** | Nova's three hardcoded sites + UNROUTABLE_PROVIDERS + rate-limit regex + OPENROUTER_API_KEYS rotation all read from `providers.yaml`. Kills three model/provider-name conditionals. |
| **`LLMClient.classify_loop_error` public bound-method seam** | Replaces module-level `_is_rate_limit_exception` / `_is_rate_limit_retry_state`. Every runner/judge path passes `classify_error=client.classify_loop_error` directly; no `partial`, no reach into `_compiled_rate_limit_patterns` across module boundaries. |
| **9-slot `_POLICY_REGISTRIES`** (`schema_sanitizer`, `prompt_policy`, `content_policy`, `response_policy`, `reasoning_codec`, `cache_policy`, `params_policy`, `message_assembly_policy`, `assistant_text_policy`) | New slots (`assistant_text_policy`, `params_policy`, `message_assembly_policy`) turned Cohere/Nova/generation-params hardcodes into policy dispatch. `{name, params}` slot values (with bare `name` deprecated in v0.17.0 for v0.18.0 removal). |
| **AST-based public API boundary guardrail** at `tests/unit/llm/test_public_api_boundary.py` | Enumerates the 8 currently-shipped per-model subclasses, AST-walks each source, rejects any private-symbol reach into the engine. Prevents a moved subclass silently reaching back for a `_helper`. |
| **Deprecation shim with `_WARNED: set[str]` per-name once-cache** | Python's default warnings filter dedupes on `(message, category, module, lineno)` — insufficient. Per-name cache locks "exactly one warning per name per process"; different names still fire independently. |
| **Dropped `Requires-Dist: tolokaforge` from `tolokaforge_models/pyproject.toml`** | Runner-subset distribution name is `tolokaforge-runner-subset`; a pip-level pin refuses to resolve alongside it. Stage 5's runtime `_check_minimum_engine_version` gate is the authoritative pairing check — pip-level dep was redundant AND blocking. |
| **Engine version resolution tries both distribution names** (`tolokaforge` → `tolokaforge-runner-subset` → `tolokaforge.__version__` fallback) | Runner-subset image ships as a differently-named distribution; the gate must fire correctly in both install shapes. |
| **Class-name-existence gate widened to `default` + `presets` + `providers` blocks** | Matches the overlay validator's walk; drift in `default:` or `providers:` would slip through a `presets:`-only scope. |
| **No forwarding stub for pre-cutover import paths** — fail-loud on removed data files instead | ADR-0030 § "Downstream data-resource consumers (2026-08-07)" — a build-time stub would silently serve stale prices from the first independent models release onward. Plausible-but-stale is worse than loud-failure. |
| **Bucket A/B classifier hoisted to `tools/automation/src/automation/bucket_classifier.py`** as SSOT | The canonical replay acceptance test (`test_models_wheel_replay.py`) and the Slack-triggered workflow both consume the same classifier; splitting them across two SSOTs would drift the "Bucket A" definition. |

## Industry precedents

- **`black` / `black-primer` split** — same shape: a stable base tool + independently-versioned "data" wheel it consumes. Landing shape borrowed: two-wheel install, one is a hard dep of the other.
- **`django` / `django-<vendor>` extension packaging** — deliberately rejected. Django's extension mechanism is discovery-by-import (django's `INSTALLED_APPS`); we use `importlib.metadata.entry_points` for a stronger fail-loud contract (unknown slot / duplicate collision raise at import).
- **`pip` / `pip-tools` deprecation-shim pattern** — `_WARNED: set[str]` per-name once-caching borrowed directly. Alternative (Python's `DeprecationWarning` filter with `simplefilter("always")`) requires test-side ceremony that leaks into every consumer.

## Suggested review order

Follow this order to make the ten-PR story readable:

1. **[#1002](https://github.com/Toloka/tolokaforge/pull/1002) `68f6bc41`** — fail-loud entry-point registry (#930). Prerequisite hygiene; establishes the fail-loud pattern reused throughout the milestone.
2. **[#1008](https://github.com/Toloka/tolokaforge/pull/1008) `ceeb1f57`** — public `tolokaforge.testing.certify` seam (#931). Extracts certification and widens `ModelCertificate`.
3. **[#1013](https://github.com/Toloka/tolokaforge/pull/1013) `6784d878`** — replay acceptance test (#932). The "engine-releases-avoided" metric that every subsequent PR moves.
4. **[#1016](https://github.com/Toloka/tolokaforge/pull/1016) `a761108a`** — models fingerprint (#933). `engine_run_state.json` audit trail.
5. **[#1024](https://github.com/Toloka/tolokaforge/pull/1024) `8149c02c`** — three new extension slots (#934). The 9-slot registry lands here.
6. **[#1028](https://github.com/Toloka/tolokaforge/pull/1028) `b3a4e3e8`** — provider bindings as data (#935). Nova + OpenRouter + rate-limit patterns move to `providers.yaml`.
7. **[#1029](https://github.com/Toloka/tolokaforge/pull/1029) `162571b0`** — public API + base-class hooks (#936). AST boundary guardrail lands here.
8. **[#1033](https://github.com/Toloka/tolokaforge/pull/1033) `45fb61cc`** — importlib.resources reader migration (#937). Accessors + fail-loud conversion; the `_DATA_ROOT` seam.
9. **[#1048](https://github.com/Toloka/tolokaforge/pull/1048) `51489b3d`** — THE cutover (#938). The load-bearing one-line flip. Read `docs/adr/0030-tolokaforge-models-split.md` alongside.
10. **[#1051](https://github.com/Toloka/tolokaforge/pull/1051) `9b6736d0`** — workflow retarget + docs inversion (#939). Wraps up.

## Verification

- **All 10 sub-issue PRs merged on green CI.** Every PR passed lint / hygiene-review / wheel-install-gate / test-smoke lanes. `test-full` and `test-gate` are path-gated (skipped for smaller diffs; run on the consolidation PR).
- **TestPyPI publish verified end-to-end** — run `31510609678` published `tolokaforge-models 1.0.0` via OIDC Trusted Publishing. Real PyPI will accept `models-v1.0.0` at first tag push without config changes.
- **Byte-identity across the cutover** — `tests/canonical/test_models_fingerprint_flip_invariant.py` hardcodes `EXPECTED_SUB_PAYLOAD_SHA256 = "29b3ee7c769aeca01a50bfbcc656e2bc86183fef7edb8339851894a6483004e1"` — the pre-cutover `sha256({presets, pricing, certificates})`. Post-cutover recomputation matches, locking that no certificate reordering, whitespace tweak, or byte-copy drift slipped through.
- **Runner-subset partition invariant preserved** — `test_subset_files_do_not_import_orchestrator_only_siblings` AST walker still green post-flip. `model_data.py` remains subset-safe; heavy imports live in `model_data_fingerprint.py` sibling.
- **Public API boundary preserved** — `tests/unit/llm/test_public_api_boundary.py` enumerates the 8 moved per-model subclasses at their new `tolokaforge_models.policies.<family>` paths and re-verifies zero private-symbol reach.
- **Deliberately skipped:** full canonical sweep on the merge-commit tip surfaces two pre-existing hangs (`test_run_trial_composition.py`, `test_ungradeable_trial_accounting.py`) unrelated to any milestone stage; filed as follow-ups.

## Load-bearing assumptions on consumer side

**Consumer resolution shape.** The independent-cadence promise (models 1.0.0 → 1.1.0 without re-cutting the engine) holds for consumers that declare `tolokaforge>=0.17,<0.18` in their own `[project.dependencies]` (PyPI resolution). Consumers that **git-pin the engine by tag** — e.g. `{ git = "...", tag = "v0.17.0" }` — instead pull the whole engine repo and uv resolves `tolokaforge-models` from that tag's `[tool.uv.workspace]` declaration, pinning it to whatever the engine checkout carried. In that shape, `uv lock --upgrade-package tolokaforge-models` is a no-op and picking up a new model still means bumping the engine ref. The migration to a PyPI-based engine dependency is downstream work and is a prerequisite for the independent-cadence property to be available to that consumer.

**Publish ordering.** Engine `v0.17.0` declares `tolokaforge-models>=1.0.0,<2.0.0` in `[project.dependencies]`. If `v0.17.0` publishes before the first `tolokaforge-models` wheel exists on PyPI, `pip install tolokaforge==0.17.0` fails resolution. Correct order: cut `models-v1.0.0` **first**, verify on PyPI, then cut `v0.17.0`. Trusted Publishing has already been validated end-to-end on TestPyPI (run `31510609678`).

## What's next

- **First real cutover release:** cut `v0.17.0` (engine) and `models-v1.0.0` (models wheel) in either order after this PR merges. Once both tags publish, `pip install tolokaforge` transitively pulls the models wheel.
- **`benchmark-results-collector`** migration to `tolokaforge.core.model_data.bundled_pricing_path()` — stays with @bberkes-toloka, lands ahead of `models-v1.0.0` publish so no consumer sees the removed engine data dir.
- **Deprecation shim removes in v0.18.0.** The 8 `from tolokaforge.core.llm import <MovedSubclass>` re-exports and the `MODELS_*_VERSION` engine constants are scheduled for removal in the next minor.
- **First live Bucket A integration** after this merges tests the "engine version stays put across a lineup refresh" property — the acceptance-test metric from #932 will reflect it.


bberkes-toloka:
Reviewed against one question: can the arena onboard any model, including whatever customization it needs, without pulling a new engine version? Line references are against `85361f5`.

The mechanism is real and in places better than the ADR promised. A brand-new class name registers from the models wheel and merges into `_POLICY_REGISTRIES` with fail-loud on unknown slot and on shadowing an engine class. The helpers a per-model class actually needs are public now (`coerce_json_strings`, `coerce_empty_containers`, `find_additional_properties`), `StrictSchema.inline_refs_in_tool` is a real overridable hook, and all four models-wheel policy modules import public engine API only. Two things I want to call out because I underrated them in the ADR review: **provider bindings as data** is a bigger arena win than the policy split, since onboarding a whole new provider becomes one YAML edit instead of client branches; and the muse-class problem is genuinely gone, with `_UNRELIABLE_COLD_CACHE_REPORT_NAMES` replaced by `cert.excluded_capabilities`, which converts a historically Bucket-B integration into data.

Scope note for everything below: we are retiring the v1 dataset and its in-flight evals, so only **future** eval branches matter. That narrows some of this and sharpens the rest.

## 1. The split gives the arena nothing while the arena pins the engine by git tag

This is not a defect in the PR. It is an assumption the PR depends on that is written down nowhere, and it is currently false.

`tolokaforge-tasks` pins the engine directly by git tag (`pyproject.toml:35`, `tag = "v0.16.1"`), and uv honours a git dependency's own `[tool.uv.sources]`. With `tolokaforge_models` declared a workspace member (`pyproject.toml:529`) and sourced as `{ workspace = true }` (`:543`), the models package resolves out of the engine tag rather than from PyPI. I reproduced the real shape on a synthetic repo pair:

```
source = { git = file://…/enginerepo?subdirectory=mods&tag=v0.17.0#291a11b }
```

Then I bumped the member to 1.1.0 in that repo, left the engine ref at `v0.17.0`, and ran `uv lock --upgrade-package <member>` on the consumer. It stayed at 1.0.0. **The independent-upgrade path is a no-op for any consumer that git-pins the engine.**

Consequence for the arena today: picking up a new model, even a pure pricing entry, means editing the engine ref on the eval branch and relocking. Byte-for-byte the same motion as before this PR.

The fix is on our side, not yours: future eval branches pin the engine **from PyPI**, at which point the models wheel is an ordinary dependency with its own range and `--upgrade-package` works. I checked the alternative of dropping the workspace-member declaration and I do not think you should — that is how the monorepo is developed, and paying for it to work around a consumer's pin style is the wrong trade.

What I would like from this PR is two things:

- **State the assumption**, ideally in ADR-0030 § What "success" looks like and in the PR description: the independent-cadence property holds for consumers resolving the engine from PyPI, and does not hold for consumers git-pinning it. Without that written down, the milestone reads as "the arena is decoupled" and it is not.
- **Confirm the publish ordering.** `tolokaforge-models` is not on PyPI today (I get a 404), and engine v0.17.0 declares `tolokaforge-models>=1.0.0,<2.0.0`. If v0.17.0 lands first, `pip install tolokaforge==0.17.0` fails for everyone. Under the PyPI-pin plan above this stops being a corner case and becomes the only install path we have.

## 2. The automated path still writes new classes into the engine

This is the one I would fix before merge, and with the dataset scope narrowed it is the top item: every future integration goes through the workflow.

`resolve_agent.md:62-64` still instructs the agent to *"write a NEW small reusable adapter class in the right module (e.g. a response policy in `response_policy.py`), register it in the matching `_POLICY_REGISTRIES` slot AND export it in `tolokaforge/core/llm/__init__.py`"*, and `:87` sends it to grep `response_policy.py` + `schema_sanitizer.py` for prior art. `resolve_finalize.md:18-19` then says *"If the compose step wrote a new adapter class, it is already in the engine + `_POLICY_REGISTRIES` + `__init__.py`; leave it."* Neither prompt mentions `tolokaforge_models/src/tolokaforge_models/policies` anywhere.

The plumbing is retargeted — the finalize `git add` stages the models policies dir and `pyproject.toml`, and `bucket_classifier.py` allows the whole `tolokaforge_models/` prefix, so a new class plus its entry point would classify Bucket A. Only the instructions point the old way. Every data-shaped path was retargeted carefully (`cert.py`, `pricing.py`, `probes.py`, finalize steps 1-3); the code-writing case looks like the one that was missed.

The failure mode is uneven, which is what makes it worth fixing rather than leaving as a known gap:

- If the agent writes a **subclass** of a registered base, `test_public_api_boundary.py` test 4 reds in PR CI. Visible.
- If it writes a **composite** inheriting only `object` — the `MinimaxM3TagRecoveryResponse` pattern the same prompt holds up as the model to follow — nothing catches it. The in-workflow verify only runs `import tolokaforge.core.llm` plus two policy test files, so it commits, and test 4 skips it because `:312` only flags classes with an in-registry ancestor. It lands Bucket B with no signal at all.

Two smaller things in the same area:

- The finalize `git add` whitelist does not include `tolokaforge_models/tests/`, so a test the agent writes alongside a models-wheel policy is silently dropped from the commit.
- If leaving the prompts engine-pointed is deliberate for now — human-writes-the-class-first — that is a defensible call, but then it should route to needs-human rather than commit quietly and surface as a red PR later.

## 3. The public-API guard now polices the wrong side

`PER_MODEL_SUBCLASSES` (`test_public_api_boundary.py:61`) is a hardcoded 8-entry list, and the registry-completeness check at `:308` skips any class whose module is not under `tolokaforge.core.llm.`. Every entry in the list is a `tolokaforge_models.policies.*` module, so the completeness check can never grow the list, and every future models-wheel class is exempt from the AST audit.

Under the narrowed scope this matters more, not less: the eight listed classes are the past, and everything from here is models-side. The guard that exists to stop a moved subclass reaching back into engine privates will not see any of them. Inverting the filter — audit `tolokaforge_models.policies.*` and let the engine-side branch be the exception — would make it grow with the system.

## 4. Two stale texts that now say the opposite of the design

- `presets.py:578` still emits *"New policy classes require an engine release; overlays can only reference existing ones."* This one is user-facing at runtime, and after the cutover it is wrong.
- `docs/ADD_NEW_MODEL.md:174` says *"Adding a brand-new policy class still requires an engine release"*, contradicting lines 70, 133 and 333 of the same document. Separately, step 5 is advertised in the Bucket A list but its body (around `:366`) routes the contributor to extend `../tolokaforge/core/llm/reasoning_codec.py` with fixtures under `tests/unit/llm/fixtures/` — both engine paths, and precisely the deepseek-v4-flash shape. The new `test_add_new_model_paths_are_current.py` guard cannot catch that: it checks for three legacy path strings and that markdown links resolve, and `reasoning_codec.py` resolves fine.

## 5. My follow-up, restated with a firmer gate

The consumer migration is still mine and it now hard-gates the first repin rather than being a tidy-up. `benchmark-results-collector` reads `importlib_resources.files("tolokaforge") / "core/data/pricing.json"` and `eval-orchestrator` carries the same fallback; this PR deletes that resource, and `lineup.py` is fail-closed, so `configure-model` in the eval pre-flight raises on the first v0.17 repin instead of degrading. Retiring the v1 dataset simplifies it — I no longer need a back-compat path for older engine pins — so it reduces to reading through `bundled_pricing_path()` and failing loud on an empty table instead of returning `{}`. I will land it in the same change that repins, not after.

## Not raising

The base-class-hook ceiling and the new-`Capability` case are both settled in the ADR and I am not reopening them. `@register_probe` having no shared-body dispatch yet is fine as a forward-compat seam given the byte-identical acceptance constraint; worth a line in the follow-ups so it does not read as wired when it is not.

azorej:
Full-branch review of the consolidation diff (213 files, reviewed at `85361f55`, re-checked at `85b6b0eb` — the four colleague-review fixes don't overlap with anything below). The engine-side refactor is high quality: moved code verified byte-identical against `main`, model-name conditionals genuinely killed, fail-loud discipline consistent. The issues below sit almost entirely on the operational seam the milestone creates.

## Blockers

**1. `tolokaforge_models/**` is invisible to CI.**
Neither `on.push.paths` nor `on.pull_request.paths` in `.github/workflows/ci.yml` includes it. A Bucket-A PR — the *default* flow this milestone creates, and per the updated `docs/ADD_NEW_MODEL.md` one that now carries new production Python (policy subclasses, cert registry) — merges with zero lint, zero tests, zero wheel gate. Every sub-PR went green only because each also touched engine paths. Fix: add `tolokaforge_models/**` to both trigger blocks (and audit `benchmark.yml`/`release-gate.yml` for the same gap). We'd deliberately *keep* `claude-review.yml`'s path-ignore as-is — the hygiene review on data-only diffs isn't worth the tokens; lint/tests/wheel-gate are the gates that matter.

**2. The pricing updater writes into the deleted directory.**
`tools/pricing-updater/src/pricing_updater/fetcher.py:18-21` — `DEFAULT_PRICING_PATH` still targets `tolokaforge/core/data/pricing.json` (the docstring one line above was updated; the constant wasn't; `cli.py:30` help text also stale). `pricing-updater update` — the flow `docs/ADD_NEW_MODEL.md` documents and `tolokaforge/core/pricing.py:237`'s error message names — resurrects the deleted dir and writes a file nothing reads. Silent no-op; exactly the "plausible-but-stale" hazard this PR's own design table rejects. Worth a test pinning that the default target exists.

**3. The byte-identity ratchets deadlock the Bucket-A flow they certify.**
`tests/canonical/test_models_fingerprint_flip_invariant.py:25-29` and `test_certificate_registry_flip_byte_identity.py:40-56` hardcode hashes/counts (`EXPECTED_MODEL_COUNT = 39`) over live data the system is designed to change engine-free; `tests/canonical/snapshots/models_wheel_replay/metric.json` stales on every new `integrate:` commit. The first legitimate Bucket-A change turns three canonical tests red; the finalize `git add` whitelist can't stage the test files, and if a human stages them the diff classifies as Bucket B. Today this is masked only by Blocker 1 — it detonates on the next engine-side PR, on someone else's diff. Needs a decision: (a) declare the two flip tests cutover-acceptance, state that in their docstrings, and file a follow-up to delete them after the `v0.17.0`/`models-v1.0.0` tags (cleaner); or (b) keep them as deliberate ratchets and document the regeneration step in `ADD_NEW_MODEL.md` + whitelist those files in the classifier and finalize. As-is, (b) trains people to paste new hashes blind.

**4. Credential env reads moved into the shipped wheel, in a shape the secrets grep can't catch.**
`tolokaforge/testing/certify/fixtures.py:53` does `os.getenv(cert.env_key)` (same pattern in `suite/test_thinking_replay_roundtrip.py:57`, `test_thinking_emits_blocks.py:52`, `test_decimal_field_tool_call.py:71`, `test_unsigned_thinking_replay.py:56`). Legal while it lived under `tests/`, but the move puts it on the published import path, and `tests/unit/secrets/test_no_raw_secret_access.py` only matches string-literal credential names, so `cert.env_key` sails through (verified: the grep passes on this head). Separately, that grep's `PKG_ROOT` doesn't scan `tolokaforge_models/src/` at all. Fix: route the presence check through `SecretManager` (also gains `.env` support), and widen the grep to the models tree.

**5. `AGENTS.md` gotchas #21 and #25 still link moved classes at their old engine paths.**
`AGENTS.md:477` links `GeminiSchema` at `tolokaforge/core/llm/schema_sanitizer.py`; `AGENTS.md:481` links `ARRAY_SITES` at `response_policy.py`. Both now live in `tolokaforge_models/src/tolokaforge_models/policies/{gemini,minimax}.py` (verified absent from the old files). The sweep fixed gotcha #23 but missed these two.

## CI on this head is red for two concrete reasons

- **`test-smoke`**: `tests/unit/test_docker_build_context.py::test_runner_build_context_ships_source_tree_for_multi_stage_hatch_build` — the in-container models-wheel build added `tolokaforge_models/` to the runner build context but the ADR-0025 lock test still expects the old set.
- **CodeQL** (1 high, `py/clear-text-logging-sensitive-data`, `tolokaforge/core/logging.py:325`): `_sanitize_extra` only covers the `logger.log(..., extra=...)` path. `self.logs.append(log_entry)` stores the **unredacted** `full_context`, and strict mode interpolates raw values into the `RuntimeError` message — so the docstring's "single choke point — callers cannot bypass it" doesn't hold yet. Sanitize once up front and use the sanitized dict for all three sinks.

## Major

- **Certify suite silently left the CI integration lane**: it moved from `tests/integration/llm/` to `tolokaforge/testing/certify/suite/`, which is under neither `pytest tests/integration/` (ci.yml) nor `testpaths` in `pyproject.toml`. If dropping it from the lane is intended (cost), say so in `tests/README.md`; if not, add `--pyargs tolokaforge.testing.certify.suite`.
- **`release-models.yml` `cz bump` is not commit-scoped** despite the header comment: commitizen has no path filtering — `working-directory` changes which config it reads, not which commits it scans, so the models increment/changelog derive from *every* conventional commit since the last `models-v*` tag. Also: with version already 1.0.0 and no `models-v*` tag, the first run bumps *past* 1.0.0 — `models-v1.0.0` must be hand-tagged, which `docs/RELEASING.md` should state.
- **Release bookkeeping**: `pyproject.toml` pre-bumped to 0.17.0 with a hand-written dated CHANGELOG section, outside the `cz bump` flow. If the next engine release lands as v0.18.0 instead, the models wheel's `minimum_engine_version = ">=0.17,<0.18"` refuses the freshly released pair at boot. Either document that v0.17.0 is hand-tagged at the pre-set version, or revert the manual bump and let `cz bump` produce it.
- **`TODO(#645)`** at `tolokaforge/core/llm/presets.py:265` points at an issue this PR closes — re-point it at a live follow-up or drop #645 from the Closes list.

## Worth discussing (judgment calls, not violations)

1. **Publish ordering is not "in either order"**: engine 0.17.0 pins `tolokaforge-models>=1.0.0,<2.0.0`, so `pip install tolokaforge` is unresolvable between the engine publish and the first models publish. Models must go first (or same run) — worth a line in RELEASING.md.
2. **`minimum_engine_version` carries an upper bound** (`<0.18`), making every engine minor a coordinated two-wheel release. Documented honestly, but consider whether the ceiling should live only in the engine→models pin direction.
3. **`kwargs_pin_transport` pins `api_base`/`api_key` outside the gateway-route guard** in `client.py` (unlike the slug/provider rewrites). Latent while nova is unroutable; the first routable pin-transport provider fights the gateway. Move the pin under the guard or assert `unroutable` for pin-transport bindings.
4. **`client.py:775` special-cases the literal `"OPENROUTER_API_KEYS"`** — a magic data value gating the key-file fallback, the pattern this PR kills elsewhere. A `key_file_env` field on `ProviderBinding` would finish the job.
5. **`_rotate_key` now republishes any provider's key into `os.environ`** — sanctioned by the subprocess carve-out, but the surface widened; a note in LLM_LAYER's secrets section would help.

Minor (batchable): entry-point `ep.load()` runs before `_check_minimum_engine_version()` fires (`presets.py:188` vs `:1217`) so version-skew can die with a raw `ImportError` instead of the actionable message; `ProviderBinding` lacks a validator requiring `endpoint`/`api_key_env` when `kwargs_pin_transport` is set; `ParamPolicy = ParamsPolicy` alias shipped with no deprecation shim while the eight moved subclasses got one; future-tense docstrings in `certify/probes.py:3`, `tolokaforge_models/__init__.py:4`, `assistant_text_policy.py:39`; `git_walk.py:104-109` hard-fails on merged `integrate:` subjects lacking `(#N)` — confirm the squash-title setting; `_check_class_names_resolve._walk_block` silently skips malformed slot shapes while claiming to be the sole gate; the 4 default rate-limit regexes are copy-pasted into 5 `providers.yaml` entries alongside `DEFAULT_RATE_LIMIT_PATTERNS`.

**What we verified clean**: flip-invariant/gate/registry canonical tests, unit llm/core/loop, models-wheel unit, classifier, and shim/certify suites all pass locally on the PR head; moved-code fidelity byte-identical vs main; nova three-site mapping wire-locked; Docker multi-stage rules followed; no "previously/now" doc violations. Not verified: the TestPyPI run and live-provider evidence.

Overall: with Blockers 1–5 fixed (1, 2, 4, 5 are small diffs; 3 needs a decision) plus the two CI reds, this is mergeable.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — review by Claude on behalf of @azorej

azorej (approved):

Issue #1107, opened 2026-08-13
chore(examples): bundled pack backstories still author success-only ###STOP### exits


azorej:
## What

18 bundled task packs author a `###STOP###` exit condition in their user
backstory that fires only when the agent succeeds — "once the report is
delivered", "once the agent reports the confirmation number", "once the value is
written to a submission file".

Since #1078 the simulator's own termination rule is outcome-based: `###STOP###`
is sent once every part of the request has reached an outcome, carried out **or
turned down by the agent**. But the same rewrite made the task's `Instruction`
outrank every rule below it, so an authored success-only exit **wins** — by
design. In those packs a scenario the agent correctly refuses still runs to
`max_user_turns` instead of ending with a gradeable transcript.

`docs/TASKS.md` § the `###STOP###` exit bullet now steers authors to
outcome-based phrasing, and the bundle should follow its own guidance.

## Scope

`examples/native/multi_service_helpdesk_workflow/dataset/tasks/helpdesk_01/task.yaml`
was already rephrased in #1078 because it is the doc's worked example. The
remaining 18:

- `examples/native/mock_web_booking/dataset/tasks/booking_01/task.yaml:25`
- `examples/native/persistent_tools/dataset/tasks/persistent_tools/persistent_tools_public_example_01/task.yaml:18`
- `examples/native/coding/dataset/tasks/coding/coding_public_example_01/task.yaml:20`
- `examples/native/coding/dataset/tasks/coding/coding_public_example_02/task.yaml:20`
- `examples/native/browser_task/dataset/tasks/browser/browser_public_example_01/task.yaml:22`
- `examples/native/browser_task/dataset/tasks/browser/browser_public_example_02/task.yaml:22`
- `examples/native/tool_use/dataset/tasks/tool_use/tool_use_public_example_01/task.yaml:19`
- `examples/native/tool_use/dataset/tasks/tool_use/tool_use_public_example_02/task.yaml:19`
- `examples/native/rag_search/dataset/tasks/kb_lookup_01/task.yaml:26`
- `examples/native/custom_checks/dataset/tasks/reconcile_ledger/task.yaml:27`
- `examples/native/multi_service/dataset/tasks/multi_service/multi_service_example_01/task.yaml:31`
- `examples/native/multi_service_advanced/dataset/tasks/multi_service/orders_customers_join_01/task.yaml:37`
- `examples/native/multi_service_postgres/dataset/tasks/multi_service/support_triage_01/task.yaml:44`
[...]

Issue #1106, opened 2026-08-13
user simulator: the tool-guidance prompt segment hardcodes mobile-app vocabulary; make it task-declarable


azorej:
## What

`UserSimulator._build_system_prompt` (`tolokaforge/core/llm/client.py`) appends a
four-bullet guidance block whenever the simulator holds tool schemas. Those four
bullets are written in mobile-app vocabulary:

- "tools to check device status and perform actions … when the agent asks you about your device state"
- 'if the agent asks "what does your status bar show?", you must call `check_status_bar` tool first'

No bundled task pack registers a `check_status_bar` tool, and no bundled
user-side tool schema is device-shaped. For every task in this repo the segment
therefore instructs the simulator about a tool it does not have, and names one
task family's vocabulary inside a prompt shared by all of them — the tension
`AGENTS.md` Core Rule 2 (no task-specific logic in the harness) describes.

## Why it was not fixed in #1078

The segment is pre-existing and was kept byte-for-byte on purpose while the
`Rules:` block was rewritten:

- it is pinned by the `with_tools` digest in
  `tests/canonical/test_simulator_prompt_generation.py`, whose manifest is
  hand-edited and has no regeneration mechanism;
- it is doc-referenced from `docs/LLM_LAYER.md` § The prompt body;
- editing it is a prompt-body revision, so it bumps
  `Trajectory.simulator_schema_version` and opens a new `_PROMPT_DIGESTS`
  generation — a change that re-baselines difficulty for every stamped bundle
  and belongs in its own commit, not folded into a rules rewrite.

## Proposed fix

Make the guidance segment declarable rather than hardcoded: a task/config-level
field on the user actor carrying the guidance text, with a tool-name-agnostic
default ("call the tools you hold rather than describing or inventing their
results"). Ship the generation bump plus its `_PROMPT_DIGESTS` row in the same
commit, and update `docs/LLM_LAYER.md` § The prompt body so the four bullets are
no longer described as fixed text.

Issue #1105, opened 2026-08-13
docs(output): user_system_prompt migration history survives in OUTPUT_FORMAT and artifacts.py comments


azorej:
## Summary

`docs/OUTPUT_FORMAT.md:159` and `:1450`, and `tolokaforge/core/output/artifacts.py:229` / `:373`, still reference `Trajectory.user_system_prompt` as "legacy"/"historical" to explain where `prompts.yaml`'s key names came from. Core Rule 8 forbids migration history in source-of-truth docs — the sites should justify the key names as current state (or not at all) rather than by appeal to a removed field.

Pre-existing; surfaced during #1078 (which fixed the same dead reference in `_build_system_prompt`'s docstring and the capture test). Small, but rewriting means deciding how to justify the key names without the history.

Priority: P3.

Issue #1104, opened 2026-08-13
external-harness capstone errors at setup on main — the golden cannot be regenerated


azorej:
## Summary

Both tests in `tests/integration/test_external_harness_e2e.py` error at fixture
setup on `main`. The capstone probe subprocess exits 1 before either test body
runs, so the external-harness capstone — the load-bearing matrix for #536 / #537
/ #538 — has been asserting nothing.

```
ValueError: These selected tasks cannot be graded as written, so no trial was run:
* capstone — [Errno 9] Bad file descriptor
```

Reproduce (no Docker, no keys):

```
scripts/with_env.sh uv run pytest tests/integration/test_external_harness_e2e.py -q
```

## Root cause — two defects, one behind the other

**1. A `MagicMock` adapter is read as a file descriptor.**
The probe replaces the orchestrator's adapter with a bare `MagicMock`, stubbing
only `to_task_description`, `docker_stack_requirements` and
`trial_grader_name` (`test_external_harness_e2e.py:109-113`). `Orchestrator.run()`
now calls `_reject_ungradeable_packs()` (`orchestrator.py:1698`), whose
`_grading_rejection` reads `self.adapter.get_task_dir(task_id)` — unstubbed, so
it returns a `MagicMock`. The derived grading path flows into
`validate_grading_yaml`, and `open()` on a `MagicMock` takes its default
`__index__` of `1`: the call opens **file descriptor 1** and reads stdout,
raising `OSError: [Errno 9] Bad file descriptor`. The gate catches `OSError`,
so the message surfaces as an authoring finding against the pack — pointing an
author at a `grading.yaml` that is fine.

**2. The fixture pack's grading block does not pass the gate.**
Stub `get_task_dir` and the probe still fails, now with the gate's real verdict:

```
* capstone — Grading file .../grading.yaml cannot be graded as written:
  - combine.weights.state_checks: combine.weights declares a weight for
    state_checks, which this pack does not configure: its state_checks section
    is absent or opted out and the pack configures [], so no substrate produces
    a state_checks component and the weight weighs nothing.
```

The fixture pack (`test_external_harness_e2e.py:160-172`) writes a
combine-only `grading.yaml` weighting `state_checks: 1.0` with no `state_checks`
section. That was legal when the fixture was written and is not now.

Neither defect is in the code under test: `run_trial` itself completes. Only the
orchestrator-baseline half of the probe fails, and it takes the whole payload
fixture down with it.

## Why it went unnoticed

[...]

PR #1109, opened 2026-08-13
feat(core): user simulator fidelity — the simulated user says what the task author intended

### LINKED ISSUE #1075: `first_message` is declared in the runner schema and populated by an adapter, but **nothing ever reads it**
## Summary

`first_message` is declared in the runner schema and populated by an adapter, but **nothing ever reads it**. A task cannot pin its opening message, even though nearly every conversational task authors one. This is a generic engine gap, not adapter-specific.

## The three layers

| layer | has `first_message`? | populated? | consumed? |
|---|---|---|---|
| `core/models/task_config.py:UserSimulatorConfig` — the task.yaml model | **no** | n/a | n/a |
| `runner/models.py:RunnerUserSimulatorConfig` (`:244`) | yes | only by `tolokaforge-adapter-tlk-mcp-core` (`adapter.py:1680`, `:1770`) | **never** |
| `UserSimulator` (runtime) | no parameter | — | — |

`core/conductor.py:662` builds the simulator from `task.resolve_user_simulator()` — the **task-level** config — passing `mode`, `llm_config`, `persona`, `backstory`, `scripted_flow`, `tool_schemas`, `rate_limit_probe`. The runner-level field is therefore a dead end: it is not the config the conductor reads for the simulator, so an adapter populating it changes nothing.

Two consequences worth calling out:

- The task-level model sets `model_config = {"extra": "ignore"}`, so a `task.yaml` that declares `user_simulator.first_message: "…"` is **silently dropped** — no error, no warning.
- Downstream tooling assumes this works. A pre-run gate in our repo documents "the engine sends `fixtures/user_simulator.json:first_message` verbatim as the user's first turn" and keeps those fixtures in sync with the backstory. That statement is false against v0.16.1, and the fixtures have no runtime effect.

## Why it matters

In our corpus, **1,689 of 1,690** task backstories carry an authored opening line under a `Your reason for contacting support:` marker — the literal first message the scenario owners wrote, review, and audit. Today the simulator regenerates that line from the backstory on every episode, which:

- spends an LLM call to reproduce text we already have;
- makes the opener non-deterministic, so the audited wording is not necessarily what the agent sees;
- exposes the highest-stakes turn to simulator failure modes. Tasks that instruct the user to withhold a fact until asked (43 in our corpus) depend entirely on the model honouring that instruction, and a scratchpad-polluted opener dumps every withheld fact at once — see the companion issue.

A pinned opener removes all three by construction.

## Proposed fix

1. Add `first_message: str | None = None` to `core/models/task_config.py:UserSimulatorConfig` so a task can declare it.
2. Pass it through `conductor.py:662` into `UserSimulator`.
3. When set and the conversation has no prior turns, emit it verbatim as turn 1 and skip the generation call. Every later turn is unchanged.
4. Decide what the runner-level field means now: either route it through the same path, or drop it so it stops implying support that does not exist.

Backwards compatible — tasks that set nothing keep generating turn 1 exactly as today.

azorej:
# Milestone 30: User simulator fidelity

Milestone: https://github.com/Toloka/tolokaforge/milestone/30
Umbrella: https://github.com/Toloka/tolokaforge/issues/1079

## TL;DR

`UserSimulator` now delivers what task authors wrote. An authored opener reaches the agent verbatim with the generation call skipped; a generated turn follows rules that defer to the backstory instead of fighting it; and no turn is ever silently mutated between generation and delivery — a defective reply is discarded whole and regenerated, or the trial fails loudly as our defect. Compatibility impact is deliberate and gated: `ENGINE_PROTOCOL_VERSION` moves 1 → 2 (older engines are refused at the registration gate with the rebuild-the-image remedy), `simulator_schema_version` moves 2 → 3 (difficulty re-baselines; consumers comparing runs across the boundary gate on the stamp), and rate-limit-probe configs whose episode budget sits in the 5711–8420 s window are refused at load with migration text. Everything else is additive: three new `trajectory.yaml` keys, three new bundle `task.yaml` keys, one new detector seam. Closes #1075, #1077, #1076, #1081, #1078.

- **Today, clean runs are unchanged.** Tasks that set nothing keep exactly today's behaviour: unset openers still generate turn 1; simulators that never break frame or leak delimiters never meet the reply guard; every recorded bundle keeps loading (all bundle changes are additive with defaults).
- **Today, three fail-fast conversions.** A blank `initial_user_message` is refused at load (it silently meant "generate anyway"); every spelling of `first_message` is refused with an error naming `initial_user_message` (it was silently dropped); probe configs in the 5711–8420 s episode window are refused at load (the queue-lease bound they claimed was false once the guard could retry). Zero in-tree occurrences of any of these; each carries CHANGELOG migration text.
- **On upgrade, two version gates.** A runner image built from this branch refuses older engines at `RegisterTrial` (protocol 2) — `make docker-build-core` is part of the upgrade for every runner-image consumer. And `simulator_schema_version: 3` marks the prompt-rewrite boundary: **difficulty re-baselines**, with two deltas pushing in opposite directions (dropped correction rules make tasks harder; outcome-based `###STOP###` lets partial-block scenarios terminate gradeable). The umbrella carries the re-baseline task with the per-delta measurement plan; do not infer direction from aggregates.
- **Longer-term commitments filed, not implied:** the untagged-scratchpad detector needs a stance classifier and a labelled corpus (#1095); agent-path delimiter stripping belongs in `AssistantTextPolicy` (#1094); the mobile-vocabulary tool-guidance segment becomes task-declarable behind its own generation bump (#1106); 18 bundled backstories still author success-only exits (#1107).

## Design walkthrough

The milestone is one pipeline seen at three layers: what the simulator is told (input contract), what it says (generation), and what the engine does with what it says (delivery guard + record).

```mermaid
flowchart TB
    subgraph authored["Task pack (authored)"]
        A1["initial_user_message<br/>(pinned opener — #1075)"]
        A2["actors.user.backstory<br/>(the Instruction)"]
    end

    subgraph prompt["Input contract (#1078)"]
        P1["Rules: 12 rules, precedence clause first —<br/>'Your Instruction above is authoritative'"]
        P2["digest manifest guards the prompt body<br/>simulator_schema_version 2→3"]
    end

    subgraph gen["Generation"]
        G1["UserSimulator._llm_reply"]
    end

    subgraph guard["Delivery guard (#1077 seam, #1076 extension)"]
        D1["FourthWallDetector<br/>(frames, not words)"]
        D2["ScratchpadDetector<br/>(line-leading think tags)"]
        D3{"defect?"}
        D4["discard whole reply,<br/>regenerate (≤3 attempts)"]
        D5["UserReplyRefused →<br/>trial fails as HARNESS_ERROR"]
    end

    subgraph record["Bundle record (#1075/#1077/#1081)"]
        R1["trajectory.yaml:<br/>first_user_message_source,<br/>user_reply_guard_events"]
        R2["task.yaml:<br/>interaction_mode, initial_user_message,<br/>user_actor (resolved)"]
    end

    AGENT["Agent receives the turn"]

    A1 -- "verbatim, generation skipped" --> AGENT
    A2 --> P1
    P1 --> G1
    G1 --> D1 --> D2 --> D3
    D3 -- clean --> AGENT
    D3 -- hit --> D4 --> G1
    D4 -- "budget spent" --> D5
    guard --> R1
    authored --> R2
```

Key mechanism: the pinned opener (top edge) bypasses generation entirely — which removed the highest-risk surface for every downstream defect before the guard even existed, since the measured scratchpad exposure was an opening-message rate. The guard never edits: the one text transformation left on the user path is the pre-existing `###STOP###` partition, which consumes a harness-protocol token the simulator's own prompt instructs the model to emit — not words the model committed to as the user.

## Key design choices

| Decision | Rationale |
|---|---|
| #1075: canonical pinned-opener field stays `initial_user_message`; `first_message` spellings refused, not aliased | The field already shipped end-to-end; a second spelling would owe a precedence rule forever — the actual bug was silent dropping |
| #1075: `ENGINE_PROTOCOL_VERSION` 1→2 rides with the dead-wire-field removal | Old engines emit `"first_message": null` (no `exclude_none`); the registration gate is the designed refusal point with the correct remedy, and a canonical guard now pins doc == constant |
| #1075/#1077: enum values reach YAML as `.value` / `model_dump(mode="json")` | A `str`-Enum member in `yaml.dump` emits `!!python/object/apply:` and makes the whole file unreadable to `safe_load` consumers |
| #1077: regenerate, never repair | Text surgery synthesizes turns the model never committed to — the failure mode the milestone removes; after bounded retries the trial fails visibly as our defect, counted in the denominator |
| #1077: detector precision outranks recall; every recall loss recorded | A false positive is fatal (3 generations + a failed trial) where it used to be a silent word deletion; the corpus grew 47→89 rows through adversarial review and every deliberate miss is in the residual list |
| #1077: `UserReplyRefused` carries reason codes, never reply text | `classify_loop_error` has a prose tier matching `"API"`; quoting a reply would reclassify our defect as the agent's (`HARNESS_ERROR` → `MEASURED`) |
| #1077: retry cost wired into the probe-budget invariant | `turn_budget_s`/`turn_overshoot_s` read `USER_REPLY_MAX_ATTEMPTS`; without this the queue-lease bound a config-load validator enforces becomes silently false |
| #1076: no `strip_reasoning` port on the user path | Where regeneration is available, curing an artifact by editing spends the delivered-verbatim guarantee to save one call; stripping belongs on the agent path, which cannot regenerate (#1094) |
| #1076: untagged-scratchpad families deleted, not narrowed | Two adversarial measurement rounds proved the distinction is stance, not surface form — every bound variant fell to fresh false positives; the attack evidence seeds #1095 |
| #1081: bundle key is `user_actor`, not `user_simulator` | A `TaskConfig(**bundle)` reload lifts the deprecated `user_simulator` spelling with a warning; `user_actor` is inert — locked by a DeprecationWarning-as-error reload test |
| #1081: the writer serialises the caller's snapshot whole | `write_task_info` rebuilt the payload from a hard-coded key list, silently dropping unknown keys — the root cause that made a conductor-only fix a no-op |
| #1078: a precedence clause instead of per-conflict patches | Enumerating conflicts doesn't scale; the first rule makes the family self-healing and is the only guarantee for backstories that author their own corrections |
| #1078: the digest manifest is hand-edited by design | A regenerable golden would let a prompt edit be blessed without the stamp bump — the exact failure the guard exists to prevent; its transcription-checksum property caught a real bug in flight |
| #1078: no in-repo LLM smoke | The one bundled pack exhibiting a measured conflict needs Docker + live tokens; a single run cannot produce the two per-delta measurements; n=1 against a stochastic simulator is not evidence about a precedence bet |

## Suggested review order

1. **#1075 — PR #1087 (`4c24760e`)** — the input contract's foundation: one pinned opener, delivered verbatim and visibly. Read the premise correction first (the field already existed; the issue's proposed `first_message` would have been a second spelling).
2. **#1077 — PR #1093 (`8bf49403`)** — the seam everything later plugs into: `UserReplyGuard`, `FourthWallDetector`, the probe-budget widening, the trajectory record. The largest PR; the detector corpus and its falsifier discipline are the heart.
3. **#1076 — PR #1097 (`d3176363`)** — the seam contract proven: one detector class, one registration, zero guard edits. Read the design-decision section for why stripping was rejected and the untagged half deleted on evidence.
4. **#1081 — PR #1102 (`1045cdad`)** — the measurement surface: bundle records the resolved user actor; the writer pass-through fix. Small; read the reload-inertness test.
5. **#1078 — PR #1108 (`6ca5f5c6`)** — the payoff: the Rules rewrite, the stamp's forcing function, and the #1104 repair the critique loop surfaced. Read the digest-manifest module first, then the prompt diff.

## Verification

- Per-PR CI: all five PRs green on the branch lanes (lint, hygiene-review, wheel-install-gate, test-smoke); the single red across the milestone was a network flake on #1087's wheel-install-gate, green on one re-run.
- Post-merge validation ran on the integration branch after every merge (targeted unit+canonical modules; 96/186/188/51/23-test spot lanes, all green).
- Full lanes at the final head: 7566 unit+canonical green. The 8 persistent failures in `tests/unit/test_persistent_shell_local.py` are environmental (bracketed-paste PTY escapes, filed as #1098) and reproduce on an untouched tree.
- The integration capstone (`tests/integration/test_external_harness_e2e.py`) — a lane CI never runs on PRs — was repaired (#1104) and run locally at every #1078 stage: 2 passed at the final head, wire golden regenerated by its own mechanism with the diff accounted per stage.
- Deliberately not run: real-LLM smoke (argued on the record in #1078's plan — no bundled pack exercises the measured conflicts without Docker + tokens); the downstream per-delta re-baseline (post-merge work, tracked on the umbrella).

## What's next

- The umbrella (#1079) carries the mandated **difficulty re-baseline task**: two per-delta measurements on the downstream corpus, partitioned by `simulator_schema_version` 2 vs 3, using #1081's bundle record to identify per-trial user configuration.
- Follow-up issues filed during the milestone, by theme: detection (#1094 agent-path stripping, #1095 stance classifier), prompt (#1103 persona, #1106 tool-guidance vocabulary, #1107 pack exit sweep), infrastructure (#1080 dead wire config, #1085 Trajectory extra-forbid, #1086 tbench description predicate, #1091 requires-python floor, #1098 PTY escapes, #1099 untyped bundle payload, #1100 writer output parity, #1101 xdist log leakage, #1104 capstone on main, #1105 legacy doc history).


bberkes-toloka:
**`FourthWallDetector`: one FP class missing from the residual list**

The `named_the_exercise` copula rule (`this <exercise noun> (is|was|tests|…)`) accepts a bare `is`/`was` with no attributing continuation, so an ordinary support turn from a customer of simulation software matches:

- `"This simulation is crashing every time I open it."` -> hit
- `"This simulation keeps crashing on my laptop."` -> passes

The split is the verb, not the meaning. The module docstring enumerates the false positives kept deliberately and this shape is not among them; the neighbouring pattern already solves it with `_EXERCISE_END`, which requires a continuation. Same class: `"I'm not a real customer, I just want a quote before I book."` matches `denied_being_human`, because `_HUMAN_END` is satisfied by the comma in `_CLOSE`.

Given the stated cost of a false positive (3 generations plus a trial counted in the denominator as `harness_error`), worth either extending the residual list or requiring the same attributing continuation on the copula branch.

Related: `UserReplyGuard.enforce` re-invokes the identical closure (same system prompt, same context, fixed `temperature=0.2`), so a defect driven by the backstory reproduces on all three attempts and the turn is refused. Was varying the retry considered, e.g. raising temperature or appending a one-line frame reminder to the *prompt*? That would not edit the reply, so it seems compatible with the "regenerate, never repair" rule.

azorej:
Both findings confirmed live and fixed on this branch in `b3b2d86f` (PR #1120) — thank you, both were real and your examples are now MUST_PASS corpus rows verbatim.

**Copula branch**: we went one step further than requiring the attributing continuation — the verb branch (`this <exercise noun> (is|was|tests|measures|…)`) is deleted outright. Mapping every MUST_DETECT row to its owning pattern showed the branch locked **zero** detections, and its non-copula verbs carried the identical defect (`This simulation measures heat transfer across the wall.` was also a hit). Its one legitimate shape (`This simulation is a roleplay exercise.`) is absorbed by widening the demonstrative-article pattern's head, which already carries `_EXERCISE_END`.

**`denied_being_human`**: split along the adjective×noun cross product. `person`/`human` denials stand alone (no support turn denies personhood — `I'm not a real person.` stays detected). Role-noun denials (`customer`/`user`/`caller`) now require the denial to *name the exercise* in its continuation — `I'm not a real customer, this is a benchmark.` stays detected; your prospective buyer passes. One measured surprise: sentence-final does **not** stay detected for role nouns — `I'm not a real customer. I just want a quote…` is the same FP with a full stop, so the discriminator is the continuation, not the punctuation. The standalone role-noun denial is recorded as deliberately given up in the module docstring's residual list.

Corpus: 56 pass / 33 detect → 65 / 38; all 33 original detections keep their reason codes; each narrowing has a falsifier that reds exactly its own rows.

**On varying the retry**: considered during the guard's design and deliberately deferred — the #1077 plan's risk section records the escalation path (a corrective nudge appended to the retry *request* is prompting, not mutation, so it stays inside the regenerate-never-repair rule) and the two reasons it wasn't built: it changes the request mid-trial and therefore needs its own recording on the trajectory (a varied attempt is a different measurement than a plain retry), and there was no refusal-rate evidence yet to justify it. Your framing (temperature bump and/or a one-line frame reminder) is now tracked as #1113 with exactly those constraints; the trigger condition is refusals showing up in practice once this ships.

balazs-toloka (dismissed):

CiroGamboa:
Re-read the PR against head `b3b2d86f` after the two follow-up commits — nice narrowing of the copula and role-denial branches. A batch of things I'd still flag, grouped by area and skipping the retry-diversification point already on the thread.

### `FourthWallDetector` — three residual leak/FP shapes

Manual regex trace, corpus not re-run:

1. **`_CLOSE` catches the possessive apostrophe** (`reply_guard.py:132`), so `self_identified_as_model` fires on ordinary customer speech about owning/using an AI product:
   - `"I'm an AI's owner and my system is down."` → HIT `"I'm an AI"`
   - `"I was the LLM's user last week."` → HIT `"I was the LLM"`
   Not in the docstring's residual list. Same 3-generation cost as the shapes `b3b2d86f` closed.

2. **`_MACHINE_END` participles fire on third-party AI mentions** (`reply_guard.py:146`) — `would`, `following`, `talking`, `designed` combine with the two-word filler slot to hit a customer describing an AI, not identifying as one:
   - `"I was told an AI would help me."` → HIT `"I was told an AI"`
   - `"I was hoping an AI assistant would answer."` → HIT `"I was hoping an AI assistant"`
   - `"I was told your AI system was designed for that."` → HIT `"I was told your AI system"`
   The participle list is broader than the `chatbot following a script` shape it was justified by.

3. **`named_the_exercise` misses `we're`** (`reply_guard.py:216`). The optional group `(?:'m|\s+(?:am|are))?` covers `'m` (which never appears with `we`) but not `'re`, so a clear plural self-identification leaks:
   - `"In this simulation, we're playing the role of a customer."` → MISS
   - `"In this roleplay, we're representing a customer."` → MISS
   `we am`/`we are` still hit. Reads as an oversight; not in the accepted-miss list.

### `_sanitize_user_text` removal delivers bare AI-adjacent tokens verbatim (`llm/client.py:2450`)

`FourthWallDetector`'s docstring (`reply_guard.py:39-42`) explicitly declines bare `ai / model / prompt / benchmark / simulation / llm`, and the old sanitiser is gone. `"Sorry, the AI is thinking about this."` and `"Just check the prompt I sent earlier."` now ship into the agent transcript intact. Not one of the two deltas the difficulty re-baseline is scoped to measure — worth an explicit residual-list entry either way.

### Scratchpad — `\r`-only line endings bypass `_THINK_TAG` (`scratchpad.py:67`)

`re.MULTILINE` `^` anchors on `\n`, not on a bare `\r`. `"plan\r</think>\rreply"` → MISS; `\n` and CRLF hit. Any provider whose streaming/chunk boundaries surface `\r`-only content bypasses the detector — the exact leak the module exists to catch. Cheap fix: normalise `\r` → `\n` first, or extend the anchor to `(?:^|(?<=[\r\n]))`.

### Reply-guard runtime

- **`exc.add_note()` is Python 3.11+ (`reply_guard.py:372`), floor is 3.10.** `pyproject.toml` still declares `requires-python = ">=3.10"`. On 3.10, an `except Exception` around `generate()` that fires after ≥1 discard raises `AttributeError` from inside the handler; the original exception's text is lost, and a real RATE_LIMIT is misclassified as HARNESS_ERROR.
- **`max_attempts` ctor knob defeats the probe-budget invariant (`reply_guard.py:334`).** `UserReplyGuard.__init__` accepts `max_attempts=…`, but `validate_rate_limit_probe_budget` reads only the module-level `USER_REPLY_MAX_ATTEMPTS`. Any test double or future integration constructing `UserReplyGuard(max_attempts=6)` silently doubles the per-turn simulator-call budget while the queue-lease invariant still thinks 3 is the ceiling.
- **Non-refusal exceptions after discards drop the defect list (`reply_guard.py:376`).** `enforce` re-raises with a note, but `rejected` stays local. The runner's `except UserReplyRefused` branch doesn't run and `_record_user_reply_guard` is never called on the non-refusal path, so `user_reply_guard_events` misses those discards — they survive only in WARNING logs, not the trial artifact.

### Bundle telemetry

- **`first_user_message_source=None` when bootstrap is REFUSED** (`runner.py:559`). `_bootstrap_via_simulator` records `REFUSED` at :656 then re-raises; the assignment `_first_user_message_source = SIMULATOR` at :560 never runs. Trajectory writes `first_user_message_source: null` (documented as "trial never bootstrapped") next to a guard event at index 0 saying the simulator was dispatched.
- **`DELIVERED` recorded before the empty-text check** (`runner.py:636`). `_record_user_reply_guard(outcome=DELIVERED, message_index=len(self.messages))` runs before the empty-text `RuntimeError` at :645-649. On empty text the trial terminates with no user message appended, but the trajectory carries a `DELIVERED` event at index 0 pointing at what ends up being a SYSTEM initialization-error message.
- **Guard retries triple-bill the observation (`llm/client.py:2429`, :2435).** Same `observation` threaded across up to `USER_REPLY_MAX_ATTEMPTS` retries; `cost_usd`/`total_tokens`/`latency_s` accumulate with no attempts marker on the observation itself. In rate-limit-probe mode `_record_probe_success` fires per successful generation, inflating the probe-throughput census up to 3× — the measurement the mode exists to produce.

### Boundary schema

- **`write_task_info` no longer defaults missing keys to `{}`** (`output_writer.py:69`). The old signature filtered a fixed key list and defaulted missing entries; the new one `yaml.dump`s `task_config` whole. Callers that omit `grading_config`/`tools`/`policies` now write a `task.yaml` where those keys are absent, so any downstream reader that used to do `task_yaml["grading_config"].get("trace_checks")` now `TypeError`s on `NoneType`.
- **`_refuse_first_message` only runs on dict input** (`models/task_config.py:126`). The before-validator returns `data` unchanged when it isn't a `dict`, so a `UserSimulatorConfig` / `ActorSpec` instance revalidated (`model_validate(model)`, `model_copy(update=...)`) bypasses the retired-key refusal. The docstring claims every spelling reaches the refusal, but the model-instance path is a live bypass.
- **`simulator_schema_version 2→3` conflates two versioning axes (`models/trajectory.py:568`).** The stamp's own docstring says it exists for the simulator prompt shape / message-trace shape. The bump moves it for both a Rules-prompt rewrite AND two new trajectory fields (`first_user_message_source`, `user_reply_guard_events`). Downstream analytics gating on `>= 3` can't tell whether a run got the new prompt, the new fields, both, or a future orthogonal change. Splitting into separate stamps (or renaming to make one of the axes explicit) keeps each monotonic.

azorej:
Thanks — every item was probed live before triage (each detector claim re-run against head, the runtime/schema claims traced through the call sites). Verdicts below in your grouping. The fixes landed via PR #1126 → `31d84981` on the integration branch.

### Fixed in `31d84981`

**1 — possessive apostrophe.** Confirmed by probe. The machine families now refuse a possessive noun (`(?!'s\b)` after the noun), scoped to those families only — `your system prompt's wording` still names the agent's prompt, so `named_a_party_prompt` keeps the apostrophe. Both of your shapes are locked MUST_PASS rows.

**2 — report verbs bridging into a third party's machine.** Confirmed, and the fix is structural rather than a verb blocklist: the premodifier slot must now be entered through its own determiner (`I am` reads as a copula only while the noun phrase heads with `an/the/my/your`). Every FP in this family reaches the noun across a verb with the determiner *after* it, so the ordering refuses them all — including two bare-noun variants a blocklist would have missed (`I was told AI is unreliable.`, `I'm hoping an AI can call me back.`), found while probing and also locked. The slot carried zero MUST_DETECT rows, so this was pure unmeasured recall funding two confirmed FP families. Compensating widenings (determiner list, downplaying-adverb list) hold today's recall flat and each preserved shape is now its own locked+falsified row.

**3 — `we're`.** Confirmed oversight; `'re` added to the copula alternation in both noun branches, one locking row per branch (a single row would leave the other copy unmeasured — falsifiers confirm each reds alone).

**4 — sanitizer-removal residual entry.** Agreed it deserved an explicit entry. The residual list (and `docs/LLM_LAYER.md`) now state it in substance: no sanitizer remains on the path, bare nouns are deliberately unmatched, your two example sentences reach the transcript intact backstopped only by the prompt rule — and this is *outside* the two deltas the `simulator_schema_version` 2→3 re-baseline is scoped to measure.

**5 — `\r`-only scratchpad bypass.** Confirmed by probe. Took your anchor-extension option, expressed once: `(?:\A|(?<=[\r\n]))` with `re.MULTILINE` dropped (the two candidates measured identically on every row). Deliberately no normalisation — the excerpt still slices the original text, so a `\r` shape quotes `\r` back honestly.

Fourth-wall corpus: 65 pass / 38 detect → 74 / 43; scratchpad 6/6 → 6/7. No locked detection lost; eight falsifier patches each red only their own rows.

### Tracked

**6 — `add_note` vs the 3.10 floor.** Real, and pre-dates this branch as a class: the floor is stale repo-wide. Tracked in #1091 (raise `requires-python`); if that issue resolves the other way, the `add_note` call sites go with it.

**8 — non-refusal exceptions drop the discard record.** Confirmed and filed as #1125 with your trace. The discards survive only in WARNING logs today; the fix belongs in the runner's generic exception path, not in `enforce` (the guard can't write trajectory records), which is why it's an issue rather than a batch fix.

**7 — `max_attempts` ctor knob.** Half-accepted: the knob is test-facing and production only ever constructs the default — `tests/unit/test_rate_limit_probe_wiring.py:154` locks `UserReplyGuard().max_attempts == USER_REPLY_MAX_ATTEMPTS`, so a drifted default fails CI. What was missing was the contract being stated; `__init__` now carries a docstring saying a non-default value is outside the probe-budget invariant. If an integration ever needs a non-default value, that's the moment to thread the instance value into `validate_rate_limit_probe_budget` — speculative today.

### Where the probe says otherwise

**9 — `first_user_message_source: null` on refused bootstrap.** Behaviour is as designed: the field records the source of a message that *exists*; a refused bootstrap produces no message, and the guard event at index 0 with `outcome: refused` is the record of the attempt. What was genuinely missing was the pairing being documented — `docs/OUTPUT_FORMAT.md` now states that `null` + a refused index-0 event is the signature of a guard-refused opening (in `31d84981`).

**10 — `DELIVERED` before the empty-text check.** Designed ordering. `DELIVERED` is the *guard's* verdict — the reply passed inspection and the guard released it. The empty-text refusal at `runner.py:645` is a different seam (a transport/provider defect, not a content defect), and folding it into the guard record would misattribute it. A trajectory with a `DELIVERED` event and no user message at that index is the honest record of "the guard passed it; delivery then failed downstream" — and the trial fails loudly at that point, so the state can't be mistaken for a clean run.

**11 — retry "triple-billing" the observation.** The accumulation is the point, not a leak: each retry is a real wire call with real cost/tokens/latency, and the observation exists to record what the trial actually spent (per the #1077 plan — retry cost is why `turn_budget_s` is sized as `per_call + USER_REPLY_MAX_ATTEMPTS × simulator_per_call`). Same for the probe census: `_record_probe_success` counts successful generations because throughput-under-rate-limits is measured in real calls, and a guard retry *is* load. An attempts-count on the observation would be a fine addition — the per-attempt record already exists in `user_reply_guard_events`, so the join is available today.

**12 — `write_task_info` no longer defaulting keys.** Deliberate, and adjudicated during #1081's review: the hard-coded key list was the root cause of the bug that PR fixed (silently dropping unknown keys made a conductor-only change a no-op). The conductor — the sole production caller — always supplies the full mapping, so no shipped `task.yaml` loses a key. A reader doing `task_yaml["grading_config"].get(...)` over hypothetical partial payloads is reading a file only the conductor writes; re-adding writer-side defaults would reintroduce the masking behaviour.

**13 — `_refuse_first_message` model-instance bypass.** Probed, and it's inert: `model_copy(update={"first_message": ...})` sets a stray attribute on a config that has no such field — it's excluded from `model_dump()`, never serialises, and never survives `TaskConfig` intake or actor resolution, so nothing downstream can ever read it. `model_validate(instance)` round-trips through the declared fields only. The refusal exists to catch the authoring surface (YAML/dict), which is the only path where a human can misspell the key; a dict-only before-validator is the standard pydantic shape for that. Happy to tighten the docstring wording from "every spelling" to "every authored spelling" if you find it overclaims.

**14 — stamp conflation.** The premise doesn't hold: the two trajectory fields did *not* ride the 2→3 bump. `first_user_message_source` and `user_reply_guard_events` landed while the stamp was still 2 (PRs #1087/#1093 — additive fields, within the stamp's documented scope), and the 2→3 bump (PR #1108) moved for exactly one thing: the prompt-body rewrite, enforced by the digest manifest keyed per generation. So `>= 3` means precisely "the new prompt"; the fields are present on stamp-2 runs too. One stamp, one axis — the monotonicity you're asking for is what shipped.

(The retry-diversification point stays on its own thread — tracked as #1113.)

CiroGamboa:
Re-checked against `31d84981` — the machine-noun and residual-list fixes close five of the fourteen points from #issuecomment-5279536077 (the three `FourthWallDetector` regex shapes, the bare-token residual-list entry, and the `\r`-only `_THINK_TAG` leak). Nice.

Nine still stand. Ranked by whether they're worth blocking on:

**Worth doing before merge — cheap and load-bearing**

- **Guard retries triple-bill the observation** (`llm/client.py:2429`/`:2435`). Same `observation` is threaded into up to `USER_REPLY_MAX_ATTEMPTS` retries, so on any guarded turn where the first N discards then the (N+1)th succeeds, `cost_usd` / `total_tokens` / `latency_s` accumulate silently, and `_record_probe_success` fires per successful generation — inflating the rate-limit-probe throughput census up to 3×. #1127 is reshaping the merge gate around exactly the `wire.infra` numbers this feeds; cleanest to close from this side.
- **`first_user_message_source=None` when bootstrap is REFUSED** (`runner.py:559`). Assignment at :560 sits *after* the call that raised, so every refused-bootstrap trial writes a bundle where the source field says "never bootstrapped" and guard event 0 says the simulator was dispatched. Two-line fix.

**Nice-to-have same PR, cheap**

- `exc.add_note()` on Python 3.10 floor (`reply_guard.py:416`) — one-line guard or drop the floor if 3.10 isn't in the CI matrix.
- `DELIVERED` recorded before the empty-text `RuntimeError` check (`runner.py:636`) — reorder.
- `write_task_info` no longer defaults missing keys to `{}` (`output_writer.py:69`) — small restore, but shim tests explicitly relied on it.

**Follow-up issues — latent, rare, or design call**

- `UserReplyGuard(max_attempts=…)` ctor knob vs module-level `USER_REPLY_MAX_ATTEMPTS` invariant (`reply_guard.py:378`) — no in-tree bypass today.
- Non-refusal exceptions after ≥1 discard drop the `rejected` list on the way out (`reply_guard.py:420`).
- `_refuse_first_message` `isinstance(data, dict)` guard bypassable by `model_validate` on an instance / `model_copy(update=…)` (`models/task_config.py:126`).
- `simulator_schema_version 2→3` gating both prompt-shape and trajectory-shape (`models/trajectory.py:568`) — worth splitting into two stamps before the axes diverge further, but not a bug.

azorej:
Went through the ranked nine against the sources — the two blocking-tier items dissolve under the code they cite, so nothing here changes `31d84981` before merge. Evidence per item:

### The two "worth blocking on"

**Retry billing → #1127.** The link doesn't connect: the probe census never reaches `wire.infra`. Those numbers are assembled in `tools/automation/src/automation/observe.py:237-246` from **per-trial** trajectory counters — `termination_reason` tallies (`rate_limit`, `api_error`, `api_timeout`, `stuck`) and `status == "error"` — and #1127's rate denominator is the trajectory-file count (`observe.py:214`). `probe_successful_calls` / the Little's-law census is not read by observe at all. A guard retry therefore changes neither the numerator nor the denominator of the gate #1127 reshapes — with one honest exception: a guard-*refused* trial terminates HARNESS_ERROR and counts once in `status_error`, which is correct, because that contamination is ours. Within the trial artifact, the accumulation on the observation is also as-designed: each retry is a real wire call, and recording its real cost/tokens/latency is why `turn_budget_s` is sized as `per_call + USER_REPLY_MAX_ATTEMPTS × simulator_per_call` (#1077). The per-attempt breakdown already exists in `user_reply_guard_events`, so the join is available today; if #1127's successor ever starts consuming probe numbers, an `attempts` field on the observation is the two-line addition to make *at that point*, keyed to an actual consumer.

**`first_user_message_source` on refused bootstrap.** Holding the design: the field records the source of a message that exists, and a refused bootstrap produces no message — stamping `SIMULATOR` would assert a `messages[0]` that isn't in the file. The record of the attempt is the index-0 guard event with `outcome: refused`, and since `31d84981` `docs/OUTPUT_FORMAT.md` documents exactly that pairing as the signature of a guard-refused opening — a reader cannot mistake it for "never bootstrapped". If you want the field itself to be self-describing, the coherent change is a third enum value (`refused`), not `simulator` — that's a schema addition, and I'm happy to file it as a follow-up if you think the pairing isn't enough.

### The three "nice-to-have"

**`add_note` on the 3.10 floor.** The floor is false repo-wide, not at this call site: CI pins 3.12 (`.python-version`) and never runs 3.10, and #1091 is already titled "requires-python >=3.10 is a false floor — the tree needs 3.11+" (`typing.Self` alone appears across `task_config.py`, `trajectory.py`, `run_config.py`). Guarding one `add_note` while the tree needs 3.11 elsewhere is a fix that only pretends; the real fix is #1091 raising the declared floor.

**`DELIVERED` ordering.** Holding: `DELIVERED` is the guard's verdict on *content*; the empty-text refusal at `runner.py:645` is a transport seam. Reordering deletes the fact that the guard ran and passed the reply — strictly less information in exactly the trial you'd be debugging — and the trial fails loudly at that point, so the state can't be read as clean.

**`write_task_info` defaults.** Not a regression to restore — the current behaviour is deliberately locked by `tests/unit/test_output_writer.py::test_write_task_info_invents_no_key_the_caller_omitted` (#1102). The old key-list-with-defaults *was* the root cause that PR fixed: it silently dropped unknown keys, which made a conductor-only change a no-op. The shim tests that relied on defaulting were updated in that adjudication, and the conductor — the only production caller — always supplies the full mapping.

### The four follow-ups — agreed, and mostly already filed

- `max_attempts` ctor knob: agreed no in-tree bypass; the contract is now stated in `__init__`'s docstring (in `31d84981`) and the default is CI-locked. Threading the instance value into `validate_rate_limit_probe_budget` becomes real the day a non-default constructor call exists.
- Non-refusal exceptions dropping the `rejected` list: **#1125**, filed with the full trace.
- `_refuse_first_message` instance path: probed inert — the stray attribute set by `model_copy(update=…)` is excluded from `model_dump()`, never serialises, and never survives `TaskConfig` intake or actor resolution, so no downstream reader can observe it. The refusal targets the authoring surface, which is the only path a human can misspell.
- Stamp split: the axes haven't diverged — both trajectory fields landed while the stamp was 2 (#1087/#1093, additive, within the stamp's documented scope), and 2→3 moved for the prompt rewrite alone, enforced per-generation by the digest manifest. The stamp is single-axis today; splitting it buys a second stamp with no second axis to put on it. The day trajectory shape moves independently is the day the split earns its keep.

CiroGamboa (approved):

PR #1110, opened 2026-08-13
feat(llm)!: param_value_rules — declare a value a route will not take

balazs-toloka:
Providers refuse individual **values** of parameters we send. Each such gap has so far cost a bespoke constructor kwarg on `GenerationParams` plus a use site — an engine release per gap. This is the general form.

Two cases exist and they sit at **different layers**, which is what makes a general form worth having rather than a third kwarg:

| declaration | layer | what the claim is about |
|---|---|---|
| `reasoning_effort: medium` | `providers: gemini:` | a litellm defect on **one route**; OpenRouter is unaffected |
| `tool_choice: auto` | a preset | Cohere's Chat API has **no AUTO at all**; true on every route |

The same block is legal wherever `params:` is, so choosing the layer is choosing what the claim is about.

```yaml
params:
  param_value_rules:
    tool_choice:
      auto:
        action: drop
        evidence: "2026-08-12, Cohere Chat API: no AUTO; omission is its documented equal"
```

## Three actions, no gatekeeping

- **`reject`** refuses to build the request, naming the remaining choices and the evidence.
- **`drop`** omits the parameter and lets the provider default apply.
- **`override`** sends a declared replacement, named by a required `with:`.

All three work on both rulable parameters. The engine does **not** decide which combination is sensible — that is a configuration choice, and the operator making it knows their own tolerance. What it guarantees is that a declaration is never accepted and then ignored: every action has a consult site for every parameter.

`RULABLE_PARAMS` lists the parameters a rule can reach. A rule on anything else is refused, because nothing would ever read it — that is a typo, not a decision.

> [!WARNING]
> `override` silently satisfies a call the provider would have refused, and nothing in the response says so. `drop` can change the request too — omitting `reasoning_effort` is not free.
>
> Anything derived from such a call is **not directly comparable** with a call that sent the requested value, across models, providers, or time. The engine logs a `WARNING` on every substitution naming the requested value, the value sent, and the evidence. `docs/LLM_LAYER.md` carries the full warning and the rule of thumb: prefer `reject` when you can act on the failure, reach for `override` when getting an answer at all is worth more than comparing it.

`evidence` is required and non-blank on every rule. A value gap is a claim about a provider on a date; the Gemini rule is conditional on an open upstream bug and nothing else records when to re-check it.

## Breaking: `unsupported_effort_levels` is removed

It carried no evidence, covered one parameter, and could only ever mean "refuse". The bundled Gemini declaration is migrated and verified behaviour-identical. An overlay still using the key fails loud at overlay **load**, with a `ValueError` naming the file and the legal keys.

That migration puts a `param_value_rules` block in the models wheel, so the wheel now needs an engine that understands it. `minimum_engine_version` cannot be raised here — a floor can only name a version that exists, and this feature is not cut yet — so `docs/RELEASING.md` gains the ordering: **cut the engine, raise the floor, then cut the models wheel.** Until that happens, a `pip install -U tolokaforge-models` against an older engine passes the import gate and then raises `TypeError` at the first Gemini-route capability build.

## Layering

`params:` merges one level at four sites and `param_value_rules` is nested inside it, so a shallow merge let an overlay declaring **one** rule delete **every other** rule — an operator adding a `tool_choice` rule silently disarmed the Gemini guard, and the run proceeded with the quirk unguarded and no error. `_merge_params` now merges per parameter and per value; the innermost spec still replaces. One pre-existing exception is documented: an overlay `presets:` entry that shadows a bundled preset **by name** replaces it wholesale.

## Verification

- `tests/canonical` + `tests/unit`: **7369 passed**, 1 failed — `test_docker_build_context.py`, which needs a Docker daemon this sandbox lacks and fails identically on `main`.
- 28 tests. The client ones drive `LLMClient._build_kwargs` through a real operator overlay, not a re-implementation of its branch logic.
- `pre-commit run` clean on every touched file.

## Review history

Two independent adversarial passes. The first found three blockers — a models-wheel floor break, the layering loss, and an action that constructed cleanly then did nothing on the wire — all reproduced and fixed. The second ran ≈12.5k `adapt()` calls per tree across every preset, six providers, eleven reasoning configs and both dispatch paths: **zero kwargs diffs, zero raise flips**, the only caller-visible delta being the Gemini error-message text, which nothing in this repo or the sibling repos matches on. It also found six stale references to the removed key and two behaviour/doc disagreements (`tool_choice: drop` not logging; the reject message reporting non-rejections as refused), all fixed.

Two findings deliberately **not** acted on: the commit-history footnote in `docs/GEMINI_QUIRKS.md` naming the old key is a historical reference, and `tests/canonical/test_certificate_registry_flip_byte_identity.py` carries the same "Scheduled for deletion" docstring as the fingerprint lock this PR retires — same argument applies, but it does not fail today and it is not mine to remove on a hunch.

The last commit retires `tests/canonical/test_models_fingerprint_flip_invariant.py`, whose own docstring scheduled it for deletion once the cutover shipped. It is a separate commit so it can be dropped on its own.


docs/LLM_LAYER.md
```diff
@@ -1113,6 +1113,107 @@ that silently drop `additionalProperties` parameters).
...
+it describes a **model** (the Cohere entry — true on every route because it is
+the vendor's API contract). Choosing the layer is choosing what the claim is
+about.
+
+`unsupported_effort_levels` was the older shorthand for a `reasoning_effort`
+rejection. It is **removed**: it carried no evidence, covered one parameter, and
+had no way to say anything but "refuse". The shipped Gemini declaration is
+migrated. An operator overlay still using the old key fails loud at overlay *load*, with
+a `ValueError` naming the file, the block, and the keys that are legal — before
+any model resolves.
```
azorej: 🔴 **Blocker — doc freshness (AGENTS.md Core Rule 8):** this paragraph is a migration-history block — "`unsupported_effort_levels` was the older shorthand … It is **removed** … The shipped Gemini declaration is migrated" is removed behaviour described in past tense, which source-of-truth docs must not carry. The removal rationale already lives in the PR description / changelog.

Keep only the current-state sentence:

```suggestion
`unsupported_effort_levels` is not a `params:` key. An operator overlay
carrying it fails loud at overlay *load*, with a `ValueError` naming the
file, the block, and the keys that are legal — before any model resolves.
```

tolokaforge/core/llm/params_policy.py
```diff
@@ -278,30 +411,90 @@ def _resolve_budget(self, cfg: ReasoningConfig) -> int:
...
                 f"ReasoningConfig(effort_hint={effort!r}) is declared "
                 f"unsupported for this provider+model combination "
-                f"(unsupported_effort_levels={sorted(self._unsupported_effort_levels)}). "
+                f"(refused: {sorted(refused)}). Evidence: {evidence}. "
                 f"Use one of {list(supported)!r}, or route through a transport "
                 f"that supports this effort level (e.g. OpenRouter rather than "
                 f"the direct provider, when available). See "
                 f"tolokaforge_models/data/model_presets.yaml for the declarations."
             )
+        if action == "drop":
```
azorej: 🟠 **Major — test coverage (missing behaviour-lock):** `drop` on `reasoning_effort` is the one untested cell of the 2-param × 3-action matrix the docs guarantee ("each action has a consult site for each parameter"). The other five cells are locked (`reject`/`override` on `reasoning_effort` via `adapt`, all three on `tool_choice` through the real `_build_kwargs`), but nothing exercises this branch — which must both suppress the `reasoning_effort`/`extra_body` emission *and* log the substitution warning. If someone removes the early `return`, the dropped effort still emits and ships silently.

Suggested test (in `tests/unit/llm/test_param_value_rules.py`): build `GenerationParams(param_value_rules=…("reasoning_effort", "medium", "drop", evidence))`, call `adapt` with `effort_hint="medium"` under `caplog.at_level("WARNING")`, then assert `"reasoning_effort" not in kwargs and "extra_body" not in kwargs` and that the substitution warning fired.

tolokaforge/core/llm/params_policy.py
```diff
@@ -132,6 +157,121 @@ def __getattr__(name: str) -> Any:
...
+            )
+        for value, spec in values.items():
+            if not isinstance(spec, dict):
+                raise ValueError(
+                    f"param_value_rules[{param!r}][{value!r}]: expected a "
+                    f"mapping with 'action' and 'evidence', got "
+                    f"{type(spec).__name__}."
+                )
+            action = str(spec.get("action", "")).lower()
+            if action not in VALID_RULE_ACTIONS:
```
azorej, line 214: 🟡 **Minor — fail-fast:** `action`, `evidence`, and `with` are validated here, but an *extra* key in a rule spec — `note:`, `until: 2027-01`, `expires:` — passes silently. The overlay load validator stops at the `params:` top level (`_reject_unknown_params` doesn't descend into `param_value_rules`), so nothing else catches it. That's inconsistent with every other guard in this function, all of which raise — and with its own docstring ("a preset that would otherwise look like it does something and quietly do nothing").

Fix: in the spec loop, `unknown = set(spec) - {"action", "evidence", "with"}` → raise naming the unknown keys and the legal ones.

tolokaforge/core/llm/params_policy.py
```diff
@@ -132,6 +157,121 @@ def __getattr__(name: str) -> Any:
...
+#: lets the provider default apply. ``override`` sends a declared replacement.
+#:
+#: ``drop`` and ``override`` both change what the request carries, and only the
+#: caller knows whether that matters — omitting ``tool_choice`` is the
+#: provider's own spelling of ``auto`` and costs nothing, while omitting
+#: ``reasoning_effort`` yields the provider's default budget rather than the
+#: level asked for. The engine does not adjudicate: it applies the declaration
+#: and logs a WARNING naming both values, so a caller that compares results can
+#: see the deviation. ``docs/LLM_LAYER.md`` carries the warning in full.
+VALID_RULE_ACTIONS: Final[frozenset[str]] = frozenset({"reject", "drop", "override"})
```
azorej: 🟡 **Minor — type-system fit (AGENTS.md type table: named values → `str, Enum`):** `"reject"` / `"drop"` / `"override"` are compared as bare string literals at three consult sites across two modules (here at `_emit_effort_kwargs`, and in `client.py`'s `_build_kwargs`). A typo at a future consult site (`action == "overide"`) is a silent no-op — precisely the failure class this feature exists to eliminate.

Fix: `class RuleAction(str, Enum): REJECT = "reject"; DROP = "drop"; OVERRIDE = "override"`, type `_ValueRule.action: RuleAction`, derive `VALID_RULE_ACTIONS` from it. (`_ValueRule` as a frozen dataclass is the right choice per the table — keep that.)

tests/unit/llm/test_param_value_rules.py
```diff
@@ -0,0 +1,382 @@
...
+    @staticmethod
+    def _with_overlay(overlay: dict):
+        import tempfile
+        from pathlib import Path
+
+        import yaml
+
+        from tolokaforge.core.llm import presets
+
+        path = Path(tempfile.mkdtemp()) / "overlay.yaml"
```
azorej: 🔵 **Nit — test hygiene:** `tempfile.mkdtemp()` (here and at line 325) leaks directories across runs — the repo convention is pytest's `tmp_path` (see `test_preset_overlay_validation.py`). Also `_with_overlay` and `_kwargs` both re-implement the same write-yaml-then-`set_overlay_path` scaffolding with in-method imports; one module-level helper (or fixture) taking `tmp_path` would cover both classes, with the imports hoisted to the top of the file.

docs/GEMINI_QUIRKS.md
azorej, line 1: 🔴 **Blocker (second half of the doc-freshness finding):** the reference list at line 487 still labels commit `8b1511d67` with the now-removed `unsupported_effort_levels` key — the `rg <old-name>` cleanup missed it. (This line isn't in the diff, hence the file-level comment.) Reword the commit label so it doesn't name the removed key, e.g. `` `8b1511d67` (effort-level guard) ``.

azorej (changes requested):
Reviewed at `0f216c81` against AGENTS.md + the code-review skill rules (worktree checkout; new + adjacent unit suites 60 passed, full canonical lane 1197 passed — the fingerprint-test retirement and preset-data migration break nothing mechanical).

**Verdict: request changes** — one doc-freshness Blocker (mechanical, two small edits: `docs/LLM_LAYER.md` + `docs/GEMINI_QUIRKS.md`) and one Major missing test for a contract cell the docs explicitly guarantee (`drop` on `reasoning_effort`). The mechanism itself is sound, well-guarded, and correctly wired at both consult sites. Two Minors and a Nit are non-binding — take or leave.

Verified rather than trusted from the PR description:

- Overlay old-key rejection: generic `_reject_unknown_params` names file, block, allowed keys + closest-match hints — locked by `tests/unit/llm/test_preset_overlay_validation.py`.
- Preset-name shadowing stays wholesale (`_merge_overlay`, `presets.py:693-710`) — doc claim accurate.
- `_merge_params` deep-merges per parameter and per value, innermost spec replaces, explicit-null clears — all deliberate and doc'd.
- `GenerationParams` is the only `ParamsPolicy` subclass, so the `KNOWN_KEYS` migration is complete; `tool_choice` is typed `str | None` so the `.lower()` is safe.
- Retirement of `test_models_fingerprint_flip_invariant.py` follows its own docstring's schedule; nothing references it.

Clean categories: secrets, harness/task-pack boundary, model-name conditionals (the feature is itself the sanctioned alternative), silent-failure handling (every invalid declaration raises; every substitution warns with both values + evidence), compat-surface migration (loud at overlay load, `feat!`, docs + CHANGELOG), tests-of-code (`TestThroughTheRealClient` drives the real `_build_kwargs`), test duplication, structure/root hygiene.

Open question (non-binding): the "Concretely: `param_value_rules` landed in the data while the floor still read `>=0.17`…" example in `docs/RELEASING.md:190-193` is accurate today (floor verified still `">=0.17,<1.0"` on this branch) but becomes stale history the moment the release checklist executes — consider making the example generic so it doesn't need a follow-up edit at release time.

azorej (approved):

PR #1124, opened 2026-08-13
integrate: azure_ai/cohere-command-a-plus-05-2026

balazs-toloka:
**Disposable test branch — do not merge.**

First live exercise of the auto-integration since the models split landed. Nothing has run through it yet: the last `integrate:` commit on `main` is 2026-08-04, and the split merged on 08-12.

The Cohere `tool_choice: auto` gap is **pre-declared as data**, using the mechanism that merged in #1110:

```yaml
cohere_command_a_plus:
  params:
    param_value_rules:
      tool_choice:
        auto: { action: drop, evidence: "2026-08-01, verified live: ..." }
```

That is a precondition, not part of the test. `tool_choice` is refused by the `azure_ai` route before the model is reached, so all 31 capability probes and all 6 shape variants fail at once and the resolve agent never sees anything. A previous attempt (#1070) died exactly there. Notably this declaration needs **no engine code** — which is the ADR-0030 claim, exercised for the first time on a real model.

**What is actually under test:** whether the agent handles the `<|START_TEXT|>…<|END_TEXT|>` markers by writing an `assistant_text_policy` subclass into `tolokaforge_models/.../policies/` with an entry point, rather than into `tolokaforge/core/llm/`. That case is why the ninth slot exists and it has never run end to end.

Also being watched: whether the derived `PER_MODEL_SUBCLASSES` audit picks a newly written class up without an engine-repo edit.

Not covered: the auto-release, and the Bucket B refusal — both live in #1067, still open. On `main` today a Bucket B result still commits, under a different label.

### DIFF
--- a/observation/TEST_RUN.md
+++ b/observation/TEST_RUN.md
@@ -1,0 +1,15 @@
+# Disposable test branch
+
+First live exercise of the post-split auto-integration, on top of PR #1110
+(`param_value_rules`).
+
+The Cohere `tool_choice: auto` gap is pre-declared as DATA in the models wheel,
+because it blocks the observe stage outright — every tool probe fails at the
+gateway before the model is ever reached, so the resolve agent would never see
+anything to work with. It is a precondition, not the thing under test.
+
+What IS under test: whether the agent handles the `<|START_TEXT|>` markers by
+writing an `assistant_text_policy` subclass into
+`tolokaforge_models/.../policies/`, rather than into the engine.
+
+**Never merged.** Delete after the run.
--- a/tolokaforge_models/pyproject.toml
+++ b/tolokaforge_models/pyproject.toml
@@ -40,9 +40,11 @@ dependencies = []
 
 [project.entry-points."tolokaforge.policies"]
 # key = "<slot>.<policy-name>" ; value = "<module>:<class>"
+"schema_sanitizer.cohere_recursive" = "tolokaforge_models.policies.cohere:CohereRecursiveSchema"
 "schema_sanitizer.gemini" = "tolokaforge_models.policies.gemini:GeminiSchema"
 "schema_sanitizer.gemini_recursive" = "tolokaforge_models.policies.gemini:GeminiRecursiveSchema"
 "prompt_policy.dict_map_hints_ref" = "tolokaforge_models.policies.inkling:RefResolvingDictMapHints"
+"response_policy.cohere_root_key_repair" = "tolokaforge_models.policies.cohere:CohereRootKeyRepairResponse"
 "response_policy.scalar_array_dict_map" = "tolokaforge_models.policies.gemini:ScalarArrayDictMapResponse"
 "response_policy.minimax_m3_tags" = "tolokaforge_models.policies.minimax:MinimaxM3TagRecoveryResponse"
 "reasoning_codec.openai_summary_replay" = "tolokaforge_models.policies.deepseek:OpenAISummaryReplayReasoningCodec"
--- a/tolokaforge_models/src/tolokaforge_models/certificates/registry.py
+++ b/tolokaforge_models/src/tolokaforge_models/certificates/registry.py
@@ -2645,6 +2645,108 @@ _ALL: list[MC] = [
             }
         ),
     ),
+    # -----------------------------------------------------------------
+    # Cohere Command A+ 05-2026 (the gateway ``azure_ai`` route, reached
+    # through OpenRouter) — landed via auto-resolve (Slack-requested
+    # integration, PR #1124). Routes through the model-specific
+    # ``cohere_command_a_plus_05_2026`` preset (see model_presets.yaml),
+    # declared ahead of the shared ``cohere_command_a_plus`` preset and a
+    # strict superset of it: ``schema_sanitizer: cohere_recursive``,
+    # ``response_policy: cohere_root_key_repair``,
+    # ``reasoning_codec: openai_summary_replay``, and the shared preset's
+    # ``tool_choice``/``auto`` drop rule carried over verbatim.
+    #
+    # The observe (default) baseline surfaced three preset-fixable clusters,
+    # each 0/15 and each reprobed 5/5 under the preset above:
+    #
+    #   * ``RECURSIVE_REF_TOOL_CALL`` — ``strict`` raised "$ref resolution
+    #     exceeded depth 16" IN-ENGINE on the recursive ``_TreeNode``, so the
+    #     model never saw ``submit_tree`` (0/15 on all four shapes). Cycle-
+    #     tolerant ``$ref`` inlining fixes the engine-side raise; the residual
+    #     ``simple`` / ``deep_chain`` / ``wide_tree`` misses were a root-key
+    #     mangling artifact (``{"submit_treeroot": …}`` carrying a correct
+    #     tree), repaired by the depth-0 rename in the response policy. Stays
+    #     ``required`` — the recursion itself was never wrong.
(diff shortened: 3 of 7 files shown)


bberkes-toloka:
## Review

Two findings about the fix, one about the certificate, and two about the pipeline that produced them. I did not re-verify the recursive-ref or dict-map recoveries end to end; I went at the gate the run itself flagged as the risky one.

### 1. The root-key repair fires far wider than its docstring says (blocker)

`policies/cohere.py:153` states the gate as: the emitted key must be **"byte-equal to some tool name concatenated with the declared parameter name"**. The code at `policies/cohere.py:212` checks something much weaker:

```python
if len(emitted_key) <= len(param_name) or not emitted_key.endswith(param_name):
    return arguments
```

Any non-empty prefix matches, not the tool name. The tool name cannot be checked because `parse_arguments` never receives it, although the call site has it in hand (`tc.function.name`, `tolokaforge/core/llm/client.py:2073`).

Run against the branch:

| emitted arguments | declared param | result |
|---|---|---|
| `{"submit_treeroot": {...}}` | `root` | renamed (intended) |
| `{"metadata": {...}}` | `data` | **renamed** |
| `{"line_items": {...}}` | `items` | **renamed** |
| `{"uuid": {...}}` | `id` | **renamed** |

Failure scenario: a single-object-parameter tool whose parameter is a common short word (`id`, `data`, `items`, `value`, `name`). A model emits the wrong parameter name, the repair silently renames it to the declared one, and the call is graded as correct. That is a scoring change, not a transport recovery, and it is invisible in the trajectory.

The conjunction still bounds it (exactly one root parameter, declared `object`, exactly one emitted key), so this is narrow, not wild. But the widening is unnecessary: thread the tool name into `parse_arguments` and require `emitted_key == tool_name + param_name`, which is what the docstring already claims and what the evidence supports.

### 2. The certificate gates on `OPENROUTER_API_KEY` for a model OpenRouter does not serve

`certificates/registry.py:2717`. Any contributor holding an OpenRouter key opens this cert's live suite against a route OpenRouter cannot serve, and every failure reads as a model problem. #738 reasoned through exactly this and chose a dedicated gate naming the deployment that serves the route. Worth keeping that conclusion.

### 3. The handoff manifest misreported the baseline (pipeline, not the agent)

`.github/workflows/integrate-model.yml:377` writes `"preset": "default"` unconditionally, and the observation comment repeats it. A preset **was** active: the pre-seed commit added `cohere_command_a_plus` with `schema_sanitizer: strict`, `response_policy: array_dict_map` and the `tool_choice` rule.

The agent worked the real baseline out anyway (its docstrings cite `strict` correctly), so nothing here is wrong in the output. But an agent that trusted the manifest would credit the preset's recoveries to the bare model and write a cert that overstates it. The field should report what actually resolved.

### 4. The run could not answer the question it was set up to ask

`observation/TEST_RUN.md` names the test: does the agent handle the `<|START_TEXT|>` markers by writing an `assistant_text_policy` subclass into the models package? No such policy exists in the diff, and it could not have:

- `test_basic_completion` passed. It asserts non-empty text, and `<|START_TEXT|>pong<|END_TEXT|>` is non-empty (confirmed live against the gateway on 2026-08-13). Nothing failed, so the agent had no signal.
- There is no `assistant_text_policy` axis. `_POLICY_REGISTRIES` (`tolokaforge/core/llm/presets.py:133`) carries `schema_sanitizer`, `prompt_policy`, `content_policy`, `response_policy`, `reasoning_codec`, `cache_policy`. There was no slot to register one into.

Both need to exist before this hypothesis can be tested: a probe that fails on wrapped text, and an axis to hold the fix.

### 5. Pricing was derived rather than declared, and that deserves a decision

`ensure-pricing` reported the miss loudly and the finalize agent solved for the rates against the run's own billing (zero-residual fit over 1317 `cost_source: litellm` calls), landing on 0.8 / 3.2, which matches the Azure list price. Measured rather than asserted, and better than a guess.

The hazard is that it encodes whatever the gateway bills. A gateway charging a discounted rate would put that discount on the board as the model's list price. There is now a second mechanism in flight for the same gap (an operator-declared price, #1064). One of them should own this, or the docs should say when each applies.

### What worked

Worth recording, since the run is the artifact under test: observe went clean and chained automatically, the fix loop took three iterations with the first two red, and iteration 2's **wrong** hypothesis (a dropped envelope, "restored" by re-nesting) is what exposed the real mechanism. The class docstring records that honestly rather than presenting the final answer as if it arrived first. The data-scope guard also fired on its own fix instead of merging it.

bberkes-toloka:
## Fixes for the review findings: `fix/cohere-integration-review`

Branch off this one, [`c0b24eb5`](https://github.com/Toloka/tolokaforge/commit/c0b24eb5). Merge it in or cherry-pick, whichever you prefer.

**1. The root-key rename is removed.** Narrowing it correctly needs the tool name threaded through the response-policy protocol, which couples the engine and models-wheel release axes for one gate. By this integration's own evidence the quirk is intermittent per tool (`submit_document` never showed it), so `recursive_ref_tool_call` is recorded as a genuine ceiling instead, with the mechanism written down. Everything else the response policy did is inherited unchanged from `scalar_array_dict_map`.

**2. The certificate gate is now `TF_COHERE_A_PLUS_GATEWAY_LIVE`** rather than `OPENROUTER_API_KEY`, so holding an OpenRouter key no longer fires the live suite at a route OpenRouter cannot reach.

**3. The certificate count and id-list pins move to 40.** The run that added the fortieth certificate left them at 39.

---

## The bigger finding: this PR's CI never ran, and nine canonical guards are red

`gh pr view 1124 --json statusCheckRollup` returns three checks: `Analyze (actions)`, `Analyze (python)`, `CodeQL`. No `test-smoke`, no `lint`, no `hygiene-review`. The integration commit was pushed with `GITHUB_TOKEN`, and a `GITHUB_TOKEN` push does not trigger workflows: the same Actions recursion guard `integrate-model.yml` documents at the top for the label path. So the pipeline's own output is never tested on the PR it writes to.

Running `pytest tests/canonical` against the **untouched** branch gives nine failures:

```
test_certificate_registry_flip_byte_identity  (count, id-list hash, and both seam mirrors)  4
test_certify_excluded_capabilities            (ratchet targets: re2, implicit_prompt_caching) 2
test_certify_suite_collection                 (collection snapshot)                          1
test_models_wheel_replay                      (bucket_a_count 4 != 3)                        1
test_policy_registry_merge                    (entry points: cohere_recursive is unexpected) 1
```

(plus `test_runner_subset_install_smoke`, which may be environmental here.)

Every one of them is a baseline or snapshot that a legitimate single-model integration necessarily moves, and **every one lives outside `tolokaforge_models/`**, so a Bucket-A-only pipeline cannot update them. That is the same coupling #1067's second change describes for `PER_MODEL_SUBCLASSES`, and #1124 is the empirical case it predicted: `test_public_api_boundary` fails here with

```
CohereRecursiveSchema inherits from registered slot base(s) [...] but is not in PER_MODEL_SUBCLASSES
```

Deriving that one tuple fixes one of the seven. The other six need the same treatment, or the allow-list needs to cover them, or the integrator has to bump them by hand every time and that step needs to be written down.

**Why this matters beyond housekeeping.** `main` has no required status checks and no required approving reviews, and `ARENA_AUTOMATION_AUTO_MERGE_ENABLED` is `true`. So `gh pr merge --squash` succeeds on a red PR. With #1067 merged, that same step then dispatches the models release, and the `release` environment has no reviewer gate. The path from an agent-authored policy to a published PyPI version has no human and no green CI in it. I have written this up on #1067 as well.

PR #1127, opened 2026-08-13
fix(automation): score observe contamination as a rate, not a boolean

balazs-toloka:
The observe cleanliness gate refused a run if **any** of `rate_limit`, `api_error` or `status_error` was non-zero:

```python
"dirty" if (not capability_ran) or any(infra[k] for k in (...)) else "clean"
```

## What that cost

Run [31694131208](https://github.com/Toloka/tolokaforge/actions/runs/31694131208), the first live exercise of the auto-integration since the models split: `status_error=1` out of **200** wire trials — 0.5% — discarded the whole observe and sent the candidate to a human.

That run was healthy by every other measure: 22 of 31 capability probes green, 200 wire trials with **0** tool-arg rejections across 911 tool calls, `rate_limit` and `api_error` both zero.

So a full-cost observe was thrown away and the resolve stage — the part that actually produces the integration — never ran. The re-run on this branch scored `status_error=0` on the same model over the same route, confirming the one error was noise rather than a systematic problem.

## The fix

Contamination is a **rate**. The gate exists so the agent never analyses failures the transport caused rather than the model, and a systematic problem arrives in bulk. A single transient hiccup cannot bias which quirks the agent sees.

| scenario | verdict |
|---|---|
| 1 error / 200 trials (0.5%) | clean |
| 3 / 200 (1.5%) | clean |
| 10 / 200 (5.0%) | clean |
| 11 / 200 (5.5%) | dirty |
| 1 / 10 (10%) | dirty |
| every trial timed out (200 / 200) | dirty |
| systematic throttling, 40 / 200 | dirty |
| capability suite never ran | dirty, regardless of rate |
| 0 trials with any noise | dirty, no division by zero |
| `max_turns=196`, `stuck=1` | clean — model behaviour, not transport |

**5% rather than something tighter** because the two ways to be wrong cost very differently. A false "dirty" throws away a paid observe run and puts a clean candidate in a human queue. A false "clean" hands the agent slightly noisy data — which is work it does anyway, since separating model quirks from transport failures is its job, and `findings.json` carries the infra counters for it to read. Above 5% the word "transient" stops being credible.

The number is a judgment call from one data point, not a measurement. That is why the rate is echoed to the run log and the Slack message: the next few runs will show where the real noise floor sits.

## `api_timeout` (from review)

The gate summed three counters but the producer emits four transport ones. `api_timeout` sits beside `api_error` in `observe.py` specifically so a wire run that died on provider 5xx/timeouts cannot read as clean — and the gate ignored it, so that run scored zero noise and chained to resolve. Pre-existing, but this PR rewrote that expression and its rationale claims to catch outages, which usually arrive as timeouts.

Nothing locked the counter list, which is how it survived. A canonical test now reads the producer's infra block and asserts the gate consumes every transport counter it emits, and that `max_turns` / `stuck` stay excluded.

## Visibility

The verdict, a reason token, the raw counts and the rate reach the run log and the needs-human Slack. Previously a reader was told "infra-dirty" and had to download the artifact to learn whether that meant one blip or an outage. The reason token also disambiguates `dirty … 0 200 0.0`, which otherwise reads as a contradiction.

## Verification

Exercised against the real `findings.json` from the run above plus the scenarios in the table. Five canonical tests lock the shape: rate not boolean, `capability_ran` absolute, numbers surfaced, the counter list matched against the producer, and the reason token. `pre-commit` clean including actionlint. 14/14 in the affected suite.

### DIFF
--- a/.github/workflows/integrate-model.yml
+++ b/.github/workflows/integrate-model.yml
@@ -460,14 +460,22 @@ jobs:
           # and `findings.json` carries the infra counters for it to read. Above
           # 5% the word "transient" stops being credible.
           #
+          # The four summed counters are the transport ones. `api_timeout` sits
+          # beside `api_error` in the producer (observe.py) precisely so a wire
+          # run that died on provider 5xx/timeouts cannot read as clean, and
+          # "an outage" -- which this gate claims to catch -- most often
+          # arrives as timeouts. `max_turns` and `stuck` are excluded on
+          # purpose: those are the model failing to finish, which is exactly
+          # what the agent is here to analyse.
+          #
           # `capability_ran` stays absolute -- a suite that never ran is not a
           # rate question. Zero trials with any noise counts as fully dirty
           # rather than dividing by zero. Counts are echoed so a reader can see
           # how close the call was, and so the real noise floor becomes visible
           # over the next few runs rather than staying a guess.
-          GATE="$(uv run python -c 'import json; f=json.load(open("observation/findings.json")); w=f.get("wire",{}); i=w.get("infra",{}); t=w.get("trials") or 0; n=sum(i.get(k,0) for k in ("rate_limit","api_error","status_error")); r=(n/t) if t else (1.0 if n else 0.0); print("dirty" if not f.get("capability_ran") else ("dirty" if r>0.05 else "clean"), n, t, round(r,4))')"
+          GATE="$(uv run python -c 'import json; f=json.load(open("observation/findings.json")); w=f.get("wire",{}); i=w.get("infra",{}); t=w.get("trials") or 0; n=sum(i.get(k,0) for k in ("rate_limit","api_error","api_timeout","status_error")); r=(n/t) if t else (1.0 if n else 0.0); print("dirty" if not f.get("capability_ran") else ("dirty" if r>0.05 else "clean"), "capability-suite-did-not-run" if not f.get("capability_ran") else "infra-noise", n, t, round(r,4))')"
           CLEAN="${GATE%% *}"
-          echo "observe gate: $GATE  (verdict noise trials rate)"
+          echo "observe gate: $GATE  (verdict reason noise trials rate)"
           if [ "$CLEAN" = "clean" ]; then
             gh pr edit "$PR" -R "$REPO" --remove-label 'automation:integrate-running' \
               --add-label 'automation:resolve-running' || true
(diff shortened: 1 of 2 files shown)


azorej:
Reviewed at `bfc29ed0` in a local worktree — read the full diff, verified every consumed field against the producer in `tools/automation/src/automation/observe.py`, and ran `tests/canonical/test_integrate_model_bucket_flow.py` locally: **12/12 pass**.

The core change is right: rate over `wire.trials` instead of a boolean, `capability_ran` stays absolute, no division by zero, verdict + counts surfaced to the run log and Slack. Shell wiring is correct (`CLEAN="${GATE%% *}"`; `$GATE` interpolates inside one quoted argument). One real gap before merge, though:

### 1. `api_timeout` is missing from the noise sum

The gate sums `("rate_limit", "api_error", "status_error")`, but the producer's infra dict (`observe.py:242`) adds two counters with this comment:

> Without these, a wire run that died on provider 5xx/timeouts on every trial reads as "clean" (no rejections, all-zero infra).

— introducing `api_error` **and** `api_timeout`. The gate picked up `api_error` but never `api_timeout`, so a run where every trial died on a provider timeout still scores `n=0 → clean` and chains to resolve — exactly the scenario that comment says the counters exist to catch. The gap is pre-existing (the old `any(...)` used the same three keys), but this PR rewrites that exact expression, and its new rationale explicitly claims the gate catches "an outage" — which very commonly manifests as timeouts. One-word fix: add `"api_timeout"` to the tuple; the new tests don't lock the key list, so nothing else needs touching. (`max_turns`/`stuck` are correctly excluded — model behaviour, not transport.)

### 2. PR description is stale

The body's verdict table still describes the 1% threshold from the first commit (`2/200 clean`, `3/200 dirty`, `1/10 dirty`); the shipped threshold is 5%, under which `3/200` is clean. The workflow comment is correct — only the PR body drifted. Worth updating so the description matches what the tests lock.

### Minor, non-blocking

- When dirtiness comes from `capability_ran=false`, the Slack message reads `dirty 0 200 0.0 … threshold 5% of trials` — a reader must infer that zero rate + dirty means the capability suite never ran. A reason token would remove the head-scratch; the counts do disambiguate, so fine as is.
- `capability_ran=true` + `trials=0` + `n=0` scores clean (empty wire chains to resolve). Identical to the old code's behaviour in that state, so not a regression.

The 5% threshold itself is a judgment call and the inline rationale (asymmetric cost of false-dirty vs false-clean; the agent reads the raw counters anyway) defends it adequately. Nice test hygiene — `_strip_comments` guarding the `"any(" not in body` assertion is a good touch.

CiroGamboa (dismissed):

azorej (approved):

PR #1136, opened 2026-08-13
fix(config): keep DockerConfig out of the orchestrator-only package

balazs-toloka:
`main` has been red since #1082 merged. Six canonical tests fail on `b0dc238` in [its own CI run](https://github.com/Toloka/tolokaforge/actions/runs/31704765596), and every branch rebased on it inherits them. This fixes all six. Two causes, one of them a real runtime break.

## 1. The runner container would have failed at import

#1082 added `docker: DockerConfig | None` to `RunConfig`, imported from `tolokaforge.docker.config`:

```python
# tolokaforge/core/models/run_config.py
from tolokaforge.docker.config import DockerConfig
```

`RunConfig` lives in `tolokaforge/core/models`, which `RUNNER_SUBSET_PACKAGES` ships **wholesale in the runner subset**. `tolokaforge/docker` is base-wheel only, and is named explicitly in the runner's forbidden-import roots. So the runner's boot closure grew all thirteen `tolokaforge.docker.*` modules, none of which exist inside the slim image:

```
tolokaforge.docker -> tolokaforge/docker/__init__.py
tolokaforge.docker.config, .container, .health, .image, .logging, .mount,
.network, .policy, .ports, .registry, .stack, .wait_for_services
```

Not a lint failure. `core/models/__init__.py` imports `run_config` eagerly, so the blast radius is the whole package, not just the runner entry point. Installing `main`'s subset wheel and importing it gives the receipt:

```
ModuleNotFoundError: No module named 'tolokaforge.docker'
  tolokaforge/core/models/__init__.py:43
  tolokaforge/core/models/run_config.py:26
```

That wheel ships 111 files and zero `tolokaforge/docker/*`. The partition tests caught a genuine defect.

`DockerConfig` is pure pydantic, importing only `logging`, `typing`, and `pydantic`. Nothing about it needs to sit behind the Docker foundation layer, and as a `RunConfig` sub-config it belongs beside the others. It moves to `tolokaforge/core/models/docker_config.py`; `tolokaforge.docker.config` becomes a re-export so all thirteen existing call sites, `tolokaforge/docker/__init__.py` included, keep working unchanged.

One behavioural repair rides along: `configure_logging()` announced itself through this module's own logger, which the move took outside the `tolokaforge.docker` hierarchy whose level that call sets, so the record was silently dropped. It now logs through the logger it just configured.

Making `tolokaforge/docker/__init__.py` lazy would not have worked. `FORBIDDEN_PREFIXES` in `test_runner_import_boundary.py` contains `tolokaforge.docker` and matches on `startswith(f + ".")`, so `tolokaforge.docker.config` is forbidden to the runner however cheap the import becomes; the path-based partition test would also still have required shipping a file under `tolokaforge/docker/`.

## 2. The canonical reuse test predates the daemon check

#1082 gated reuse on an image-identity check that queries the daemon via `_inspect_running_image`. The canonical test stubs only `_try_reuse_existing`, so the lookup found no daemon, returned `([], None)`, and the "cannot verify, recreate" arm destroyed the container and called `Container.create` - the one call the test asserts never happens.

The test now stubs the lookup too, mirroring `tests/unit/test_stack_container_reuse_tag_ordering.py`, which #1082 added. Recreate-on-unverifiable is deliberate and stays covered there by `test_destroyed_when_daemon_lookup_fails`, so this narrows the test to what it is about rather than weakening it.

## Verification

`tests/canonical` + `tests/unit`: **7735 passed, 1 failed**. `main` at `b0dc238` scores **7729 passed, 7 failed** on the same machine, so all six regressions are gone and nothing else moved. The remaining failure is `test_docker_build_context.py`, which needs a Docker daemon and fails identically on `main`. It is marked `unit` despite that, which predates #1082; tracked separately.

An independent review pass verified the claims above rather than taking them on trust: it rebuilt both wheels (subset 112 files with the new module and no `tolokaforge/docker/*`; base 291 with both module and shim), proved no new import cycle in four import orderings with class identity preserved, confirmed `RunConfig.model_json_schema()` is byte-identical across the two branches, and mutation-tested the reuse gate three ways to show the stubbed canonical test still fails when the gate is broken. It found the dropped log record, a stale ADR-0031 pointer, a missing submodule bullet, and a miscount in this body; all four are fixed above.

### DIFF
--- a/docs/adr/0031-pull-vs-build-default-for-service-images.md
+++ b/docs/adr/0031-pull-vs-build-default-for-service-images.md
@@ -315,8 +315,10 @@ run-start crash trace.
     (v0.18.0) that made the published images available in the shape
     this ADR relies on.
 - Related code:
-  - `tolokaforge/docker/config.py` — `DockerConfig.image_source`,
-    `ImageSource` literal.
+  - `tolokaforge/core/models/docker_config.py` — `DockerConfig.image_source`,
+    `ImageSource` literal. Lives under `core/models` rather than beside the
+    Docker layer because `RunConfig` carries it and ships in the runner
+    subset.
   - `tolokaforge/docker/image.py` — `Image.pull`, `ImagePullError`,
     `_cached_image_matches_platform`.
   - `tolokaforge/docker/image_source_policy.py` —
--- a/tests/canonical/test_engine_stack_component_snapshots.py
+++ b/tests/canonical/test_engine_stack_component_snapshots.py
@@ -222,6 +222,15 @@ def test_start_service_wires_component_id_on_reused_container(
         "_try_reuse_existing",
         lambda self, container_name, svc: spy_existing,
     )
+    # Reuse is gated on an image-identity check that queries the daemon. There
+    # is none here, and an unverifiable image is destroyed by design (see
+    # tests/unit/test_stack_container_reuse_tag_ordering.py), so report a match
+    # to reach the reuse path this test is about.
+    monkeypatch.setattr(
+        EngineStack,
+        "_inspect_running_image",
+        lambda self, container_name: ([stack._images[svc.name].full_tag], "img-reuse"),
+    )
     monkeypatch.setattr(
         stack_mod.Container,
         "create",
--- a/tests/integration/deploy/test_run_pull_default.py
+++ b/tests/integration/deploy/test_run_pull_default.py
@@ -318,7 +318,7 @@ class TestBuildModeSkipsPull:
             import json
             from unittest.mock import MagicMock, patch
 
-            from tolokaforge.docker.config import DockerConfig
+            from tolokaforge.core.models.docker_config import DockerConfig
             from tolokaforge.docker.image import Image
             from tolokaforge.docker.stack import EngineStack, ServiceDefinition
 
--- a/tests/unit/test_docker_config_image_source.py
+++ b/tests/unit/test_docker_config_image_source.py
@@ -11,8 +11,8 @@ from __future__ import annotations
 import pytest
 from pydantic import ValidationError
 
+from tolokaforge.core.models.docker_config import DockerConfig
 from tolokaforge.core.models.run_config import RunConfig
-from tolokaforge.docker.config import DockerConfig
 
 pytestmark = pytest.mark.unit
 
--- a/tests/unit/test_orchestrator_docker_config_forwarding.py
+++ b/tests/unit/test_orchestrator_docker_config_forwarding.py
@@ -32,8 +32,8 @@ from unittest.mock import MagicMock
 
 import pytest
 
+from tolokaforge.core.models.docker_config import DockerConfig
 from tolokaforge.docker import stacks as stacks_module
-from tolokaforge.docker.config import DockerConfig
 
 pytestmark = pytest.mark.unit
 
--- a/tests/unit/test_stack_container_reuse_tag_ordering.py
+++ b/tests/unit/test_stack_container_reuse_tag_ordering.py
@@ -19,7 +19,7 @@ from unittest.mock import MagicMock
 
 import pytest
 
-from tolokaforge.docker.config import DockerConfig
+from tolokaforge.core.models.docker_config import DockerConfig
 from tolokaforge.docker.container import Container, ContainerStatus
 from tolokaforge.docker.image import Image
 from tolokaforge.docker.stack import EngineStack, ServiceDefinition
--- a/tests/unit/test_stack_pull_vs_build.py
+++ b/tests/unit/test_stack_pull_vs_build.py
@@ -17,7 +17,7 @@ from unittest.mock import MagicMock
 
 import pytest
 
-from tolokaforge.docker.config import DockerConfig
+from tolokaforge.core.models.docker_config import DockerConfig
 from tolokaforge.docker.image import Image, ImagePullError
 from tolokaforge.docker.stack import EngineStack, ServiceDefinition
 
--- a/tolokaforge/core/models/__init__.py
+++ b/tolokaforge/core/models/__init__.py
@@ -6,6 +6,8 @@ Per-concern submodules hold the actual definitions:
 - :mod:`.grade` — :class:`Grade` and the rubric-judge audit records
 - :mod:`.trajectory` — :class:`Trajectory`, :class:`Message`,
   :class:`Metrics`, and the rate-limit probe accounting rows
+- :mod:`.docker_config` — :class:`DockerConfig`, the ``docker:`` block
+  ``RunConfig`` carries
 - :mod:`.model_config` — :class:`ModelConfig` +
   :class:`OpenRouterConfig`
 - :mod:`.run_config` — :class:`RunConfig` and every orchestrator /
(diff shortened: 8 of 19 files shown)


azorej (changes requested):
Reviewed at head 58a89961 against `AGENTS.md` / the code-review skill. The diagnosis, the module's new home, the partition coverage, and the canonical-test repair all check out end-to-end — verified rather than taken on trust: `RUNNER_SUBSET_PACKAGES` ships `tolokaforge/core/models` wholesale and the partition test derives its locks from the live boot closure, so the new placement is genuinely covered; the recreate-on-unverifiable arm stays locked by `test_destroyed_when_daemon_lookup_fails` in the #1082 unit file; the five affected test files pass 43/43 locally. One Blocker and one Minor.

### 🔴 Blocker — the re-export shim violates "internals refactor cleanly"

`tolokaforge/docker/config.py:1-24`

AGENTS.md's compatibility rule cuts both ways: compat surfaces (task contracts, run-config schemas, CLI, the published Python API) change only with migration — but **internals refactor cleanly in one commit: no duplicate exports, no deprecation wrappers**. This shim lands on the internals side:

- The published Python API is the top-level `tolokaforge` names only (`docs/PYTHON_PACKAGE.md`); `tolokaforge.docker.config` is not on it.
- The run-config YAML surface is unchanged (consistent with your byte-identical `model_json_schema()` check).
- Every consumer of the old path is in-tree — zero hits in `external_adapters/`, `tools/`, `examples/`.

The stated reason for the shim — "so all thirteen existing call sites keep working unchanged" — is migration convenience, exactly what the rule forbids; the module's own docstring self-identifies as a "Historical import path". The lazy-`__init__` argument in the PR body justifies the *move*, not the shim.

**Fix (mechanical, ~15 sites, no external consumers to protect):** delete `tolokaforge/docker/config.py` and repoint imports to `tolokaforge.core.models.docker_config`:

- `tolokaforge/docker/__init__.py:83`, `tolokaforge/docker/stack.py:39`, `tolokaforge/docker/wait_for_services.py:27` (+ docstrings at 8, 222 and the error-message text at 243), `tolokaforge/docker/stacks/{core,full,test}.py`, `tolokaforge/dx/cli/main.py:72`, `tolokaforge/core/orchestrator.py:1789`, `tolokaforge/docker/image_source_policy.py:59` (docstring `:attr:` target)
- tests: `test_stack_pull_vs_build.py:20`, `test_stack_container_reuse_tag_ordering.py:22`, `test_orchestrator_docker_config_forwarding.py:36`, `test_docker_config_image_source.py:15`, `tests/integration/deploy/test_run_pull_default.py:321`

Then drop the two sentences that exist only to describe the shim: `tolokaforge/core/models/docker_config.py:9-10` ("``tolokaforge.docker.config`` re-exports every name here…") and the re-export clause in ADR-0031 line 321. The ADR's new "lives under `core/models` because `RunConfig` carries it and ships in the runner subset" rationale is current-state and should stay.

(Keeping the `DockerConfig` export on `tolokaforge/docker/__init__.py` — imported from the new home — is fine: that package `__init__` is a pre-existing aggregation surface whose stack constructors take a `DockerConfig`. The Blocker is the standalone shim module.)

### 🟡 Minor — past-tense migration phrasing in an otherwise keep-worthy comment

`tolokaforge/core/models/docker_config.py:158-160`

The comment earns its keep (a refactorer would "fix" `docker_logger.debug` back to a module logger and silently drop the record), but "the model moved out of the tolokaforge.docker hierarchy" anchors it to a migration event that means nothing to a future reader:

```diff
-        # Emitted through the logger just configured, not this module's own:
-        # the model moved out of the tolokaforge.docker hierarchy, so a record
-        # from __name__ would fall outside the level this call just set.
+        # Emitted through the logger just configured, not this module's own:
+        # this module lives outside the tolokaforge.docker hierarchy, so a
+        # record from __name__ would fall outside the level this call sets.
```

### Clean

Correctness (no import cycle; new module imports only `logging`/`typing`/`pydantic`), testing (the canonical-test change is a legitimate narrowing, not a weakening — the `Container.create` → `pytest.fail` sentinel and router-attach assertions still lock the reuse path), type-system fit (`frozen` + `extra="forbid"` Pydantic, moved verbatim), doc freshness elsewhere (ADR-0031 was the only doc referencing the old location). No CHANGELOG entry owed — the YAML schema is unchanged and this fixes an unreleased regression. Noting out of scope: #1082's `docker.image_source` feature itself has no Unreleased CHANGELOG entry, a pre-existing gap on `main`.

balazs-toloka (commented):
Verified the rule before acting on it, and you are right. `.claude/agents/reviewer-hygiene.md:24` names "duplicate exports kept for migration, deprecation wrappers" as a Blocker in so many words, `docs/PYTHON_PACKAGE.md` does not put `tolokaforge.docker.config` on the published surface, and `external_adapters/`, `tools/` and `examples/` have zero hits. So the shim was protecting nothing but my own diff size.

Deleted `tolokaforge/docker/config.py` and repointed all 15 sites, including the three prose references in `wait_for_services.py` (docstrings at 8 and 222, error-message text at 243) and the `:attr:` target in `image_source_policy.py:59`. Dropped the self-referential shim sentence from the module docstring and the re-export clause from ADR-0031 line 321, keeping the current-state rationale as you suggested. `DockerConfig` stays exported from `tolokaforge/docker/__init__.py`, now imported from the new home.

Minor applied verbatim.

Suite unchanged at **7735 passed, 1 failed** (`test_docker_build_context.py`, needs a daemon, fails identically on `main`); ruff and black clean.

On your Slack question about being more aggressive with moving configs: I would rather not widen this PR, because it is a hotfix for a red `main` and every extra file makes it slower to review and riskier to revert. `DockerConfig` had a forcing function that the others do not: `RunConfig` carries it as a field, so the runner subset imports it whether or not anyone wants that. Worth checking whether any other config has the same property, and if the answer is "the partition tests are already green", then moving the rest is a design cleanup that can stand on its own PR and its own argument.

azorej (approved):
Re-reviewed at head 9dd6135e. Both findings from the previous review are fully addressed:

- **Blocker (re-export shim): resolved.** `tolokaforge/docker/config.py` is deleted and all ~15 in-tree sites now import from `tolokaforge.core.models.docker_config` — engine code, stacks, `dx/cli`, the orchestrator's lazy import, docstrings/error text in `wait_for_services.py`, the `:attr:` target in `image_source_policy.py`, and all five test files. The two migration-history sentences (module docstring, ADR-0031 re-export clause) are gone; the ADR keeps only the current-state rationale. A repo-wide grep finds zero remaining references to the old path (the only `docker/config` hits left are the Docker CLI's `~/.docker/config.json`, unrelated).
- **Minor (past-tense comment): resolved.** The logging comment now reads as current-state ("this module lives outside the tolokaforge.docker hierarchy…").

Verified locally at the new head: the affected unit files plus the canonical partition/boundary and container-reuse tests — **65 passed**.

One non-blocking leftover: the PR body still says `tolokaforge.docker.config` "becomes a re-export so all thirteen existing call sites … keep working unchanged", which no longer describes the branch. Worth a one-line edit before merge so the squash-commit message doesn't record a shim that doesn't exist.

PR #1160, opened 2026-08-14
integrate: google/gemini-3.7-flash

balazs-toloka:
# Integrate `google/gemini-3.7-flash` (openrouter)

Auto-resolve integration of a new candidate model.

- **Candidate:** provider `openrouter`, name `google/gemini-3.7-flash`,
  model_id `openrouter__google_gemini-3.7-flash`
- **Engine ref:** `9f8470a40faacca11e429b24d610dc6b81309b2b`
- **Observe baseline:** 2026-08-14, `default` preset, 31 capability probes × 15 reps
  + 200 wire-probe trials (0 tool-argument rejections)

## Policy

One new preset entry, `gemini_37_flash_recursive`, matching this model's two exact
globs only (no shared glob broadened, no new adapter class, no change under
`tolokaforge/core/llm/`):

```yaml
reasoning_codec: gemini
schema_sanitizer: gemini_recursive      # GeminiRecursiveSchema
response_policy: scalar_array_dict_map  # ScalarArrayDictMapResponse
```

Both policy classes already ship in `tolokaforge_models`. The observe signature was
identical to the 3.5/3.6 Flash precedent: all four `recursive_ref` shapes 0/15 from
an in-engine `$ref resolution exceeded depth 16` crash, plus the `dict_map`
`scalar_values` and `nested_in_object` variants 0/15. `gemini_recursive` makes the
wire schema finite for a genuinely cyclic `$ref`; `scalar_array_dict_map` unwraps the
synthetic `{"value": X}` map entry and pivots a dict-map nested one level inside an
object param. Reprobe: **5/5 on all six fix-targets**.

Also landed: the candidate certificate in
`tolokaforge_models/src/tolokaforge_models/certificates/registry.py`
(20 `required` capabilities, `env_key=OPENROUTER_API_KEY`) and the
`google/gemini-3.7-flash` pricing entry ($0.375 / $1.875 per 1M in/out).

Three canonical locks are regenerated because the registry genuinely gained a
certificate — the cert-count/sha256 cutover lock (39 → 40), the
`implicit_prompt_caching` ratchet baseline (the new model declares that ceiling), and
the certify-suite collection snapshot (+38 nodeids, all this model, zero removals).
Canonical suite is green at 1247 passed.

`THINKING_REPLAY_ROUNDTRIP`, `PROMPT_CACHING`, `IMPLICIT_PROMPT_CACHING` — genuine
limits of the OpenRouter `google/gemini-*` routes (Flash emits
`reasoning.encrypted` only, so there is no signed block to replay; the route exposes
no cache surface). No policy axis can reach them, so they are declared honestly
rather than treated as fix-targets.

---

Integrated via **auto-resolve**: the observe → fix → reprobe loop converged in one
iteration, and the proven overlay was folded into the shipped presets.

⚠️ **This is a disposable test branch — it must NOT be merged.**


bberkes-toloka:
## Review + green-up (human pass over the auto-resolve output)

### Verified, independently of the pipeline's own evidence

- **Pricing** matches the live OpenRouter API on all four fields: input `0.375`, output `1.875`, cache_read `0.0375`, cache_write `0.020833`. Exactly half of `gemini-3.6-flash`.
- **Preset routing** simulated first-match-wins over all 21 presets: `google/gemini-3.7-flash` matches only `gemini_37_flash_recursive` (exact glob, wins) and the generic `gemini` (`google/gemini-*`). The new entry's globs carry no wildcard, so it cannot shadow another model.
- **`model_id`** `openrouter__google_gemini-3.7-flash` agrees with `model_id_slug(provider, name)`; the ongoing ratchet in `tests/canonical/test_capability_registry.py` passes.
- **All six fix-targets re-run LIVE on this branch**, not replayed — the gap the still-open #882 self-referential gate would otherwise cover:

```
test_recursive_ref_tool_call[simple-openrouter__google_gemini-3.7-flash]            PASSED
test_recursive_ref_tool_call[deep_chain-openrouter__google_gemini-3.7-flash]        PASSED
test_recursive_ref_tool_call[wide_tree-openrouter__google_gemini-3.7-flash]         PASSED
test_recursive_ref_tool_call[nested_in_object-openrouter__google_gemini-3.7-flash]  PASSED
test_dict_map_tool_call[openrouter__google_gemini-3.7-flash]                        PASSED
test_variant_dict_map[dict_map__scalar_values-...]                                  PASSED
test_variant_dict_map[dict_map__nested_in_object-...]                               PASSED
test_variant_dict_map[dict_map__wide_map-...]                                       PASSED
```

(The `test_variant_dict_map` cases need `TF_RUN_VARIANTS=1`; they skip silently without it.)

### Changed to make CI green

`test-smoke` was red for two unrelated reasons, neither of them a defect in the integration.

**1. Pre-existing on main.** Ten tests in `test_pinned_opener.py` / `test_user_reply_guard.py` failed on `first_user_message_source is None`. Confirmed failing on main's own CI (run 31790914096) before this PR existed. Fixed upstream by #1170 while this PR was open — resolved here by merging current `main`.

**2. Registry-shape pins that a new certificate invalidates.** Four of them:

- `tests/canonical/test_certificate_registry_flip_byte_identity.py` — **deleted**. Its docstring scheduled its own deletion "in the same PR that lands the first post-cutover model integration", predicting that "the hardcoded count / sha256 here would then red every legitimate integration". This is that PR: the file landed 2026-08-12 with the split, `models-v1.0.0` / `v0.18.x` shipped the same day, and the previous integration on main is 2026-08-04. Its one non-cutover invariant — that the certify re-export IS the wheel's tuple, asserted nowhere else — moved to `tests/unit/testing_certify/test_public_api.py`.
- `_RATCHET_TARGETS_BASELINE` — investigated before regenerating, as its comment requires. The delta is exactly one addition, `openrouter__google_gemini-3.7-flash` to the implicit-prompt-caching ratchet (21 -> 22); the other two ratchet modules are untouched.
- certify collection snapshot — regenerated with `--update-canon`; all 38 added node ids belong to the new model, zero unrelated lines.
- models-wheel replay metric — regenerated; `bucket_a_count` 3 -> 4, the new row being this commit with `engine_paths: []` and `reason: "all touched paths in Bucket-A allow-list"`, which is an independent confirmation of the Bucket A classification.

Full `tests/unit/ tests/canonical/` locally: **7886 passed, 0 failed**. `ruff check` and `black --check` clean.

### Note for the docs

`docs/ADD_NEW_MODEL.md` still points the capability suite at `tests/integration/llm/`; post-split it lives in `tolokaforge/testing/certify/suite/`, where the per-model parametrisation actually collects.

Not merging — that call is the owner's.

azorej (approved):

PR #1161, opened 2026-08-14
integrate: x-ai/grok-4.6

balazs-toloka:
# Integrate `x-ai/grok-4.6` (openrouter) — auto-resolve

## Candidate

| | |
|---|---|
| provider | `openrouter` |
| name | `x-ai/grok-4.6` |
| model_id | `openrouter__x-ai_grok-4.6` |
| env key | `OPENROUTER_API_KEY` |
| engine ref | `20ebe66` (`v0.18.1-9-g20ebe66`) |
| pricing | $2.00 in / $6.00 out / $0.50 cache_read per 1M (live OpenRouter API) |

## The policy

One new preset entry, `xai_grok_4_6`, declared before the generic `xai_grok` preset so
first-match-wins routing claims this route only — grok-4 / 4.3 / 4.5 still resolve to
`xai_grok`, and no shared glob was broadened. It is a **superset** of `xai_grok`: each axis
is a subclass of the policy it replaces, so the bundled preset's behaviour is preserved by
inheritance.

| axis | `xai_grok` | `xai_grok_4_6` |
|---|---|---|
| `schema_sanitizer` | `strict` | `xai_grok_recursive` |
| `response_policy` | `array_dict_map` | `scalar_array_dict_map` |
| `reasoning_codec` | `openai` | `openai_summary_replay` |

One new adapter class, `XaiGrokRecursiveSchema` (`tolokaforge_models/policies/xai.py`,
registered as the `schema_sanitizer.xai_grok_recursive` entry point). It subclasses
`GeminiRecursiveSchema`, inheriting cyclic-`$ref` tolerance, scalar dict-map value carriage
and `oneOf`/`discriminator` flattening verbatim, and resets only the two Gemini-specific
stripping flags so Grok's `strict` stripping stays intact. The other two axes reuse
already-shipped classes. **Nothing under `tolokaforge/core/llm/` is touched.**

This closed four fix-targets, each **0/15 → 5/5**: `recursive_ref[deep_chain]` (an
in-engine `$ref resolution exceeded depth 16` ValueError raised *before* the request was
sent), `dict_map__scalar_values`, `discriminated_union__union_in_dict_map`, and
`unsigned_thinking_replay`.

`THINKING_REPLAY_ROUNDTRIP` (needs signed turn-1 blocks; none are emitted and no signature
is synthesised — the unsigned sibling **is** required), `PROMPT_CACHING` (ephemeral
`cache_control` not wired, not honoured), `IMPLICIT_PROMPT_CACHING` (no upstream cache
surfaced on OpenRouter `x-ai/grok-*`). The other 20 capabilities are `required`, including
`COST_USD_POPULATED`.

## Notes

- Integrated via **auto-resolve**: the fix loop converged on iteration 5, and the proven
  `observation/resolve/overlay.yaml` was folded into the bundled preset table unchanged.
- Two refuted hypotheses are recorded in-tree so they are not re-tried: `reasoning_codec:
  gemini` (wiped all four targets 5/5 → 0/5 — the Gemini codec *raises* on Grok's
  `reasoning.summary` entry type, crashing while parsing every response) and a fenced-JSON
  prompt directive (10/15 → 3/15).
- `data_scope_review: true` — `scalar_array_dict_map`'s nested-object pivot keys on the
  synthetic `key`-field item signature one level into object params, whose cross-domain
  reach is data-bound. Worth a reviewer's eye.
- ⚠️ `tests/canonical/test_certificate_registry_flip_byte_identity.py` is, by its own
  docstring, scheduled for deletion in the first post-cutover model integration — both
  triggers have now fired. Its count/digest were regenerated rather than the module
  retired, since deleting a canonical test is a human call. Follow-up.
- `tolokaforge_models/tests` + `tests/canonical`: 1316 passed, 19 skipped. Lint clean.
  Pre-existing and unrelated: `tests/unit/test_user_reply_guard.py` failures, which
  reproduce with these changes stashed.


bberkes-toloka:
## Review — verified against the code, not just the write-up

Rebased onto current `main` (`b1196cc4`) and pushed one fix. Everything below was measured on the PR head, not inferred.

### Verified correct

| Claim | How it was checked | Result |
|---|---|---|
| pricing lands where the loader reads it | parsed `pricing.json` | `models["x-ai/grok-4.6"]` = $2.00 / $6.00 / cache_read 0.50 per 1M, 418 models — not stranded in `_meta` |
| all three policy names resolve | entry points + `test_policies_reachable` | 10/10 |
| cert invariants | `tests/canonical/test_capability_registry` | 9/9 |
| **"superset of `strict`"** | flag matrix + differential `sanitize()` on the probes' own tools | root-description and RE2-pattern stripping **identical to `strict`**, while `gemini_recursive` keeps both — the claim holds |
| the cyclic-`$ref` fix is real | same tool through both sanitisers | `strict` **raises** `$ref resolution exceeded depth 16`; `xai_grok_recursive` returns a finite schema, 0 residual `$ref` |
| scalar dict-map carriage | `set_quantities` (`Dict[str,int]`) | `strict` emits no value slot; the new sanitiser does |
| **the two axes must move together** | wire args through both response policies | old `array_dict_map` on the new sanitiser yields `{"SKU-A": {"value": 3}}` (**corrupt**); `scalar_array_dict_map` yields `{"SKU-A": 3}`; the structured, non-scalar case is byte-identical to the baseline → no regression |
| no routing regression | `resolve_effective_preset` over all 40 certified routes | only `x-ai/grok-4.6` moves; the whole grok family stays on `xai_grok`; non-grok routes untouched |
| tests + lint | `ruff check` + `tests/unit` + models tests on the rebased head | clean, **6664 passed / 0 failed** |

The two refuted hypotheses documented in `xai.py` and the preset comment (fenced-JSON prompt directive; `reasoning_codec: gemini`) are the kind of falsifiable record `docs/ADD_NEW_MODEL.md` asks for — worth keeping.

### Fixed in 38c3fb71

The third match glob `*grok-4.6*` matched anywhere in a route name and claimed models this preset was never evidenced on — `some-vendor/not-grok-4.6-really`, `x-ai/grok-4.60`, `x-ai/grok-4.6-fast`. Narrowed to `*/x-ai/grok-4.6*` + `*/xai/grok-4.6*`: gateway-prefixed routes (`openrouter/…`, `litellm_proxy/…`) still route, the foreign match falls back to `default`.

### Left for a human — deliberately not changed here

1. **The pipeline's own data-scope warning** (comment above): the fix targets a data-related array / free-form-object recovery class whose scope the pipeline did not fully see. This is the formal blocker.
2. **`UNSIGNED_THINKING_REPLAY` as `required`.** `test_unsigned_thinking_replay` captures turn 2 at `litellm.completion` and never sends it, then asserts on our own outgoing payload (`reasoning_details` present, turn-1 text inside). Once `openai_summary_replay.encode_for_replay` runs, green is structural — so this pins an **engine contract**, not a model capability. The cert already says "contract conformance on the unsigned surface only"; worth making that explicit at the `required` line. Note #852, which existed to gate exactly this, was closed unmerged.
3. **`RECURSIVE_REF_TOOL_CALL` as `required`** rests on `deep_chain` 5/5; `simple`, `nested_in_object` and `wide_tree` still miss and are attributed to Grok's intrinsic no-tool-call rate. The reasoning is plausible and well-evidenced, but the capability suite will read flaky on those shapes.

Cert `required`/`known_unsupported` sets were left untouched: those are labelling decisions for the owner.

### Unrelated

`hygiene-review` fails repo-wide with `API Error: 404 not_found_error: model: claude-sonnet-4-20250514` — `claude-code-review.yml` pins no model and the action's default was retired. The review never runs; worth a separate one-line fix.

azorej (dismissed):

CiroGamboa (approved):

PR #1164, opened 2026-08-14
fix(tbench): gemini-cli hardenings — TRUST_WORKSPACE + settings.json pre-pin

CiroGamboa:
## Two silent-failure knobs the shipped default currently misses

Both are prerequisite for gemini-cli to actually work in headless `--prompt` mode when an operator supplies `GEMINI_API_KEY`. Without them the trial degrades silently — the CLI runs but blocks tool calls (case 1) or 400s on auth resolution (case 2).

Empirically derived from reference-invocation practice observed in an out-of-tree reference host we compared against, and cross-checked against the shipped `@google/gemini-cli@0.55.1` bundle.

### The two knobs

1. **`container_env.GEMINI_CLI_TRUST_WORKSPACE: "true"`** — without this, `--yolo` gets silently demoted to "default" mode in an untrusted workspace (the task container is not a git repo, so gemini-cli treats it as untrusted). The demotion blocks every tool call. Reference-invocation practice always sets this env.

2. **`config_files["/root/.gemini/settings.json"]`** writing `{"security":{"auth":{"selectedType":"gemini-api-key"}},"experimental":{"skills":true}}` before the CLI runs. Headless `--prompt` mode 400s with "Invalid auth method selected" without a pre-pinned `selectedType` — the CLI's auto-detection breaks in the non-interactive path. Reference-invocation practice always emits this settings file.

### YAML comment sharpened with citations for the OpenRouter impossibility

Anyone reading the entry will now see the specific evidence chain:

- **OpenRouter's API reference** declares only OpenAI-compat and Anthropic-compat surfaces — no `generateContent` endpoint.
- **gemini-cli 0.55.1's bundle** — `dist/main.mjs` grep: `chat/completions` → 0 hits, `openai.com` → 0 hits, `openrouter.ai` → 0 hits, `generateContent` → 558 hits. All six auth types (`USE_GEMINI`, `USE_VERTEX_AI`, `LOGIN_WITH_GOOGLE`, `COMPUTE_ADC`, `LEGACY_CLOUD_SHELL`, `GATEWAY`) call `new GoogleGenAI(...)` which hardcodes the Google wire protocol. The `GATEWAY` mode reached via `GOOGLE_GEMINI_BASE_URL` still speaks Google's schema — only the endpoint changes.
- **[gemini-cli issue #1515](https://github.com/google-gemini/gemini-cli/issues/1515)** — Google maintainer explicitly declining OpenRouter support upstream.

Ruled out escape hatches: self-hosted `generateContent`-to-OpenAI shim (still needs a Google credential upstream — same operator burden, extra hop), community forks (unmaintained, out of policy).

## Test updates

Three unit tests pinned the pre-change gemini command shape (no `config_files` → no `mkdir && printf && ...` preamble). Updated to the new shape:

- `test_gemini_argv_shape` — uses `.rpartition(" && ")` to strip the preamble and pin the CLI portion. Same pattern the existing `test_codex_writes_config_toml_and_auth_json_before_the_cli` test uses.
- `test_gemini_has_no_preamble_no_stdin` → renamed to `test_gemini_uses_argv_instruction_not_stdin`. The preamble now exists (config_files emits), but the intent — no stdin pipe into the CLI — is preserved and pinned by asserting no `" | gemini"` in the command.
- `test_space_style_is_two_argv_words` — same `.rpartition(" && ")` pattern.

**Reviewer nit follow-up** (commit `e4775611`): added `test_gemini_writes_settings_json_before_the_cli` locking the settings.json preamble mechanism (mirrors the codex `config.toml + auth.json` pattern), tightened `test_synthesised_compose_carries_container_env` to parametrise across every shipped harness with a non-empty `container_env` so a regression on any entry surfaces here rather than in a live matrix cell, and swapped the fragile `endswith(" go")` tail-bytes check for a parsed-argv assertion. Seeded `GEMINI_API_KEY` in the `env_backed_secrets` fixture — gemini-cli's wire protocol isn't OpenAI-compat, so it can't share the OpenRouter key.

## Verification

- [x] 198/198 unit + canonical tests pass locally.
- [x] No canonical snapshot ripple — snapshots pin claude-code, not gemini.
- [ ] Empirical validation deferred until `GEMINI_API_KEY` is available. The hardenings prevent silent failure but can't be positively verified without a real Google key.

**Note on CI test-smoke**: main-branch CI has been failing on unrelated `test_pinned_opener` + `test_user_reply_guard` tests on the last three runs — same 10 failures reproduce on `main` locally. Not introduced by this PR; unblocking requires a separate fix on the frame-guard subsystem.

## What this does NOT do

- Does NOT enable gemini via OpenRouter — confirmed impossible.
- Does NOT change any other harness's behavior — YAML edit is scoped to `gemini-cli`.
- Does NOT add proxy shims or community forks.

Follow-up: when `GEMINI_API_KEY` arrives, we run one matrix cell (`gemini-cli × gemini-2.5-flash × fix-billing-holds`) to confirm the harness runs end-to-end.

### DIFF
--- a/external_adapters/tolokaforge-adapter-terminal-bench/src/tolokaforge_adapter_terminal_bench/data/harnesses.yaml
+++ b/external_adapters/tolokaforge-adapter-terminal-bench/src/tolokaforge_adapter_terminal_bench/data/harnesses.yaml
@@ -123,16 +123,43 @@ harnesses:
     # gemini-cli's model catalog uses bare names.
     strip_vendor_namespace: true
     # gemini-cli speaks Google's native Gemini schema (`generateContent`),
-    # not OpenAI-compat, so an OpenRouter key does not authenticate its
-    # requests — pointing `GOOGLE_GEMINI_BASE_URL` at OpenRouter returns 400.
-    # The reference vendor-CLI invocation reads `GEMINI_API_KEY` and 401s otherwise
-    # (empirically verified: no matrix run of any pipeline ever produced a
-    # real gemini trajectory without a real Google AI Studio key).
-    # The operator supplies `GEMINI_API_KEY=<google-key>` in `.env`; the
-    # secret ref below is honest about what is required. Without the secret,
-    # SecretManager fails at trial start (loudly), which is the right shape:
-    # a run measuring a model against a laptop the operator can't
-    # authenticate would silently degrade to the baseline container score.
+    # not OpenAI-compat, so the CLI cannot be pointed at OpenRouter:
+    #
+    # 1. OpenRouter exposes only OpenAI-compat (`/chat/completions`) and
+    #    Anthropic-compat (`/messages`) surfaces. There is no
+    #    `POST /v1beta/models/{model}:generateContent` endpoint (confirmed
+    #    against OpenRouter's own API reference, and explicitly declined
+    #    upstream at github.com/google-gemini/gemini-cli/issues/1515).
+    # 2. `@google/gemini-cli@0.55.1`'s bundle has zero code paths that
+    #    emit OpenAI-schema requests (grep on `dist/main.mjs`:
+    #    `chat/completions` → 0 hits, `openai.com` → 0 hits). All six
+    #    auth types — `USE_GEMINI`, `USE_VERTEX_AI`, `LOGIN_WITH_GOOGLE`,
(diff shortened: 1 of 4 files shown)


CiroGamboa:
Superseded by #1228 — all changes are in the consolidated matrix harness fixes PR.

Issue #1195, opened 2026-08-14
test: whole-tests/ collection breaks on twin terminal-bench fixture basenames (echo-hello vs echo-hello-skills test_outputs.py)


azorej:
Since the harness-mode fixtures landed (#1083), `tests/data/terminal_bench_tasks/echo-hello/tests/test_outputs.py` and `tests/data/terminal_bench_tasks/echo-hello-skills/tests/test_outputs.py` are twin basenames in non-package directories. Any pytest invocation that collects the whole `tests/` tree — `make test`, `make test-unit`, `make test-canonical`, and the dev MCP `run_tests` default paths (all of which rely on `testpaths = [\"tests\", ...]`) — dies at collection:

```
import file mismatch:
imported module 'test_outputs' has this __file__ attribute:
  .../echo-hello/tests/test_outputs.py
which is not the same as the test file we want to collect:
  .../echo-hello-skills/tests/test_outputs.py
```

CI never sees it because `ci.yml` passes explicit directories (`pytest tests/unit/ tests/canonical/`), which skip `tests/data`. So the tree is green in CI and uncollectable locally.

Priority: P2 — every local whole-tree run is broken, but CI and scoped runs are unaffected.

Fix options, one of:
- add `tests/data/terminal_bench_tasks/*/tests` to `norecursedirs` beside the existing `tests/data/projects/*/tests` entry (these fixture files are task-pack payloads executed inside containers by `run-tests.sh`, not host-collectable tests);
- or rename one twin (fixture layout permitting — terminal-bench's harness convention may pin the filename);
- or switch to `--import-mode=importlib`.

The `norecursedirs` entry matches the existing precedent and does not touch the fixture layout.

Issue #1171, opened 2026-08-14
feat(config): a run can ask for artifact-write redaction — expose SensitiveKeyRedaction through run config / CLI


azorej:
## Context

#694 moved redaction to artifact-write time. `OutputWriter` and `FileArtifactWriter` take a
`RedactionPolicy`; under a redacting one the writer replaces credential-named tool-call argument
values in `trajectory.yaml` and `tool_log.yaml`, withholds `judge_trajectory.yaml`, and stamps
`metrics.yaml` with a `redaction` block that every offline grading command refuses by name.

The default is `NoRedaction`, and **nothing in the shipped engine constructs anything else**.
Policy selection is a programmatic constructor argument, reachable from a test or an embedding
caller and from nowhere an operator can get to. The mechanism exists and is behaviour-locked at the
canonical tier (`tests/canonical/test_redaction_at_artifact_write.py`); it simply cannot be turned
on.

That was deliberate: #694 was a layering chore, and an enablement knob is a compatibility surface
(run-config schema and/or CLI) carrying design questions of its own.

**Priority: P3.** The mechanism is inert until this ships, and nothing else depends on it.

## Design questions to settle

- **Where the knob lives.** `RunConfig` (per run, recorded in the run's own config snapshot) versus
  a `tolokaforge run` flag versus both — and if both, which wins.
- **What the value is.** A policy name drawn from `RedactionPolicyName`, which extends cleanly to
  #1157's value-based pass, rather than a boolean that would need replacing the moment a second
  redacting policy exists.
- **What the operator is told, and when.** A run configured this way produces bundles that
  `retrace`, `rejudge`, `curate` and `reconcile` all refuse. That belongs in front of the operator
  before the run spends, not in the stamp they find afterwards.
- **Which entry points thread it.** `OrchestratorDeps.artifact_writer`
  (`tolokaforge/core/orchestrator.py:433`) is the seam a full run goes through;
  `tolokaforge/core/run_trial.py:165` is the single-trial path. `tolokaforge prepare` / `worker`
  need it too, since a worker writes its own bundles.

## Deliverables

- [ ] An enablement surface (run-config field and/or CLI flag), `extra="forbid"` clean
- [ ] Threaded to every `FileArtifactWriter` construction site that writes a run's bundles
- [ ] The operator is told at run start that this run's bundles will be refused offline
- [ ] `docs/CONFIG.md` and `docs/SECURITY.md` § Artifact-Write Redaction record the knob and its
      consequence
[...]

Issue #1149, opened 2026-08-13
test: the 8 test_persistent_shell_local failures are equality assertions on raw pty output, not environmental noise


azorej:
Reframed by #691's implementation. **Priority: P2.**

The 8 `tests/unit/test_persistent_shell_local.py` failures carried as 'known environmental' across weeks of runs (#1098/#596 territory) are equality assertions on raw pty output: this bash emits bracketed-paste escapes (\x1b[?2004h), so `result.output.strip() == "/tmp/…"` fails wherever the shell enables bracketed paste — and passes in CI, making the suite silently environment-dependent rather than the environment broken.

Fix: ANSI-normalise the 8 assertions (containment or normalised equality). #691 shipped the same discipline for its own bash_session locks (ANSI-normalised containment of a marker, per its critique) — the pattern to copy. Once fixed, the unit lane's baseline failure set drops to zero everywhere.

Refs: #691, #1098, #596.

azorej:
Additional evidence and a scope question, from #694 Stage 1 (where these 8 reds appeared in the branch's unit lane):

- Reproduced at `6d2e0ab4` in a clean `git worktree` — same 8 failures, same 6 passes. Confirms this is base-state, not branch damage, for every branch stacked on `feat/trace-checks-tail`.
- The failing assertion shape: `result.output.strip() == os.path.realpath(...)` receives `'\x1b[?2004h...\x1b[?2004l'`-wrapped output — the bracketed-paste enable/disable pair emitted by the devcontainer's shell around everything it reads.

Scope question this issue should settle before the fix: ANSI-normalising the 8 assertions (this issue's current framing) makes the *tests* environment-independent, but leaves `PersistentShell`'s output contract as "whatever the host shell chose to emit" — if a task pack drives this shell on a bracketed-paste host, the same escapes reach the agent and the grader in tool output. The stronger resolution is wrapper-side: strip terminal control sequences from captured output, or disable bracketed paste on the spawned shell (`set enable-bracketed-paste off` / pinned rcfile), after which the existing equality assertions pass everywhere *and* the contract belongs to the wrapper. Worth deciding which layer owns the fix rather than defaulting to the test-side patch.

Issue #1130, opened 2026-08-13
main CI red since #1082: run_config imports tolokaforge.docker at module level, breaking the runner import boundary and subset partition


azorej:
**Priority: P1 — main's CI (test-full) has been red for its last three pushes** (b0dc2387, 4d11e1fe, 09aeb4dd).

Reproduced on pristine `origin/main` @ b0dc2387, four canonical failures:

- `test_runner_import_boundary.py::test_runner_does_not_import_forbidden_first_party_surface` — `tolokaforge/core/models/run_config.py:26 imports tolokaforge.docker.config`, pulling the whole `tolokaforge.docker` package into the runner's first-party closure (chain: `core.models.run_config ← core.models ← core.actors.actor ← core.llm.client ← core.grading.judge ← runner.service ← runner.__main__`).
- `test_runner_subset_partition.py::test_runner_boot_closure_is_covered_by_subset` and `::test_subset_files_do_not_import_orchestrator_only_siblings` — same root cause; the boot closure now reaches files the subset wheel omits, so the slim runner image would fail at import.
- `test_engine_stack_component_snapshots.py::test_start_service_wires_component_id_on_reused_container` — also red on pristine main (not independently diagnosed here; surfaced by the same #1082 push).

Cause: #1082 (`docker.image_source`) gave `RunConfig` a `docker:` block typed by `DockerConfig`, imported at module level from `tolokaforge.docker.config`. `RunConfig` rides in the runner's model spine, so the orchestrator-side docker package now enters the runner closure.

Fix directions: a `TYPE_CHECKING`-guarded import + a plain re-validation shape at the boundary; or move `DockerConfig` into the models package the runner already ships; or an explicit lazy accessor. Whichever lands must make the four tests above green on main.

Found while rebasing milestone 28's integration branch onto main — these four failures reproduce identically on the branch and are annotated there as upstream, not branch damage.

azorej:
[...]

PR #1196, opened 2026-08-14
feat(grading): deterministic trace checks — the trace-checks tail (37 issues)

### LINKED ISSUE #900: fix(grading): RequiredAction is extra="ignore", so a typo inside a required_actions element silently changes matching semantics
## Context

Found while planning #685 (core transcript aggregation). Not in that ticket's scope — #685 changes how `transcript_rules` sub-checks are *aggregated*; this is about a config element accepting keys it does not declare.

Every other grading config block is `extra="forbid"`, and `docs/GRADING.md` § `tool_expectations` states the reason explicitly:

> `extra="forbid"` on the block means a misspelled key (`required_toolz`) fails at load rather than grading as an empty list.

`RequiredAction` is the exception. `tolokaforge/core/models/task_config.py`:

```python
class RequiredAction(BaseModel):
    """Required tool call that must appear in trajectory"""

    model_config = {"extra": "ignore"}
```

## Measured

A `compare_arg` (singular) typo is accepted silently and inverts the matching rule:

```python
raw = {"transcript_rules": {"required_actions": [
    {"action_id": "a1", "requestor": "assistant", "name": "cancel_order",
     "arguments": {"order_id": "O1", "reason": "x"}, "compare_arg": ["order_id"]}]}}

cfg = GradingConfig.model_validate(raw)
# core accepted the typo: action_id='a1' ... compare_args=None
# compare_args resolved to: None
```

`compare_args=None` means "compare **all** declared args" (`docs/GRADING.md`), so the author who wrote `compare_arg: [order_id]` — intending to match on `order_id` alone — gets exact-equality on `order_id` *and* `reason`. The check becomes strictly harder than authored, and fails trials that satisfied what the author wrote. No error at load, no error at grade time.

The adapter reads the raw YAML dict rather than the model (`tolokaforge/adapters/native.py:645`, `compare_args=action.get("compare_args")`), so the typo is dropped on the runner path too — both substrates silently agree on the wrong semantics.

The runner-side `RunnerRequiredAction` **is** `extra="forbid"` (verified: feeding it `name=` raises `extra_forbidden`), so the two models already disagree about strictness.

## Fix

Make `RequiredAction` `extra="forbid"`, matching every sibling block and its own runner-side counterpart. Have the authoring gate report the unknown key naming the element and the fields `RequiredAction` does declare, the way the `tool_expectations` refusal reads today.

## Acceptance

- A `required_actions` element carrying an undeclared key is refused at load, naming the key and the declared field set.
- `tolokaforge validate` reports it before a run is paid for, not at grade time.
- Existing packs are swept for elements that would newly fail; any hit is either a real typo (fixed) or enumerated in the PR.

azorej:
Milestone: [Deterministic trace checks](https://github.com/Toloka/tolokaforge/milestone/28) ·
Design authority: [#684](https://github.com/Toloka/tolokaforge/issues/684)

## TL;DR

The milestone's feature half — `trace_checks` as a fourth grading component, its authoring gate,
multi-path grading, argument correlation, zero-cost replay and the migration machinery — shipped in
the first consolidation PR. This iteration is the deferred tail: 36 merged issues that make the
grading pipeline tell the truth about itself at every station a trial passes through, from the
authoring gate to the offline replay commands. Three things run through all of it. First, a check
that has never been shown to fail is not a check: the parity guard rail is re-armed so every ledger
recording site is verified by driving a real `RegisterTrial → ExecuteTool → GradeTrial`, every parity
fixture is a truthful task, and the version-lock list is a census walked against the models rather
than a hand-written list. Second, a fault the harness caused must not be priced as agent behaviour:
a lost trial, an ungradeable trial, a dead search plane, a rag-service whose model never loaded, a
runner image that cannot boot, and a tool killed by the wrong backstop each get their own vocabulary
and their own accounting instead of a fabricated `0.0`. Third, a defect the author can fix should be
refused before the run pays for it: the authoring gate grew four resolvable layers and now reaches
`id_fields`, declared-but-absent grading files, lost hash-source fixtures, jsonpath targets neither
substrate composes, and binder references that can never be true. Compatibility impact is real but
narrow and every rejection names its remedy: `grading.yaml` closes two blocks and retires one key,
`tolokaforge validate` fails packs it previously passed (zero in-repo packs move), `run`/`worker`
exit non-zero on an ungradeable trial, `migration.yaml` moves `corpus` onto the entry, and one
deliberate behavioural re-baseline — the deleted completion-marker scan — makes conversational trials
run longer than they used to. Everything else is measured at zero verdict movement, with the single
intended exception of `transcript_rules` on the core substrate.

Closed by this branch: #685, #687, #688, #691, #693, #694, #695, #697, #708, #709, #711, #717,
#721, #727, #730, #742, #762, #767, #774, #781, #793, #869, #895, #896, #915, #916, #920, #921,
#922, #923, #926, #927, #940, #979, #990, #1062, #1130. (`transcript_rules`'s undeclared-key refusal
also closed #900 in flight.)

**Nothing in this branch silently changes a shipped pack's score, with two named exceptions.** Every
other PR measured verdict movement across the example corpus and the recorded bundles and reported
zero. The two exceptions are stated loudly because they are the point of their issues:

- **`transcript_rules` on the core substrate moves** (#685). The core engine had four always-present
  buckets averaged at one quarter each, so a `disallowed_tools` violation scored `0.75` and could not
  fail the component. Core now scores through the evaluator the runner already used. Six
  author-visible semantics changes are enumerated in the CHANGELOG, pinned by a 20-row canonical
  differential written before anything moved; four snapshot files moved, each with its arithmetic
  quoted.
- **Conversational difficulty is re-baselined** (#708). The agent's text is no longer scanned for
  `###STOP###`, which is the user simulator's exit token. A trial that used to end early because the
  agent merely quoted the token now runs until the simulator or the budget ends it — more turns, more
  tokens, and a tool call that used to be dropped on that turn now executes and is graded. Wire
  values, bundle contents and `TerminationReason` membership are untouched. Re-check reason
  distributions on the first calibration run after this lands.

### What a task pack must change

| Surface | What now happens | Migration |
|---|---|---|
| `state_checks.hash` | A closed block: an undeclared key inside it is refused at load on every read path, not only the one the gate happens to walk (#730) | The refusal names the file, the closest declared field and the accepted set. Zero in-tree packs migrate — the only undeclared key any pack carried was `description`, now declared |
| `state_checks.hash.expected_state_hash` | Refused at every pack-facing read (#693) | Declare `expect_initial_state: true` for a refusal task. The message names both replacements. Recorded bundles keep loading through the drop path |
| `state_checks.jsonpaths` | `path:` addresses the trial's database, `path_glob:` the filesystem; a path addressing state the runner cannot reach is refused at the gate and at `GradeTrial` (#896) | Zero packs are refused today: all 40 tracked `path_glob` entries were already `contains_ci`. Three shipped example packs were re-authored as part of the fix |
| `state_checks.id_fields` | `tolokaforge validate` runs the same cross-check the run path enforces — unknown table, component absent from every seeded record, non-unique key (#923) | The finding's own remediation text: fix the declaration, seed the field, widen the key, or set `relaxed_validation: true`. All 107 authored blocks in the tree pass |
| `grading: <path>` | A declared path with no file at it is refused by `validate`, `run` and `prepare` before any trial is scheduled (#869) | Correct the path or create the file. Under a non-native adapter it is reported honestly unchecked rather than refused |
| `transcript_rules` elements | `required_actions` and `communicate_info` refuse a key they do not declare; `required_actions[].name` must name a tool the pack declares, and `requestor` must match that actor's own set (#685, #688) | Addressed did-you-mean refusal at the gate. Three parity fixtures grading undeclared tools were genuinely defective and were repaired |
| `transcript_rules.tool_expectations` | Agent-scoped: a `required_tools` entry needs an agent call that succeeded, `disallowed_tools` is violated only by an agent call (#695) | No authoring change. Unreachable in every run today; the contract is pinned ahead of #688 activating it |
| `tools.user.enabled` | Reaches the wire and is honoured — the simulator is offered its own tools, executes them, and they are recorded and graded (#688) | A pack declaring user tools under `agent_only` or with a scripted simulator is refused at load with the remedy |
| `max_idle_turns` | Retired accept-and-warn, with the file named in the warning (#709) | The honest home for "the agent must act" is `transcript_rules.min_assistant_turns`, per task. No shipped configuration changes a verdict |
| Failed-call `result` text | Both substrates record the tool's own message; `Tool error: <Type>: …` and `Tool execution failed: …` wrappers are gone (#717) | A matcher written against the old wrappers must target the tool's own text. None exist in-repo — and the load-time guards that made failed-call matchers unwritable are now lifted, so a trace check may assert *why* a call failed |
| `migration.yaml` | `corpus` moves onto the entry and is required in every mode; a `by:` spanning two alternative routes is refused at load (#793) | Two-edit recipe in the CHANGELOG. An absolute corpus path is refused; resolution is relative to the declaration's directory |

### What stays compatible

Recorded corpora keep loading and keep their verdicts. The three preserved bundles carrying the
retired hash literal become the retired-key load corpus rather than being re-recorded — their
provenance is a deleted adapter and a re-record's faithfulness could not be attested (#693). Bundle
`call_id` values change on disk only for a provider that mints ids as model output and reuses them;
grading derives the join key per view from that view's own observation order, so **duplicate-id
bundles already on disk are gradeable and re-checkable with no rerun** (#979). The db-service
keyless-upsert refusal is additive safety: no in-tree writer, fixture, example or recorded bundle can
emit the shape (#920).

### Wire and CLI contracts

`ENGINE_PROTOCOL_VERSION` stays at 2. The standing rule is unchanged and load-bearing: **rebuild the
runner image before rolling the engine.** Three locks move within it — `required_actions[*].tool_name`
becomes `name` under a both-directions lock (#685); `search.plane` is a one-direction lock, so a new
engine's tasks are rejected by an older image (#927); the never-populated
`GradeTrialRequest.precomputed_expected_hash` is deleted with `reserved 3` and a descriptor-read guard
proven red against a stale generated module (#693). `ExecuteToolRequest.timeout_seconds` keeps its
field with `0` now meaning "the runner resolves the budget" — the skew-compat sentinel, stated
identically in the proto and the docs (#691).

Exit codes are a contract and three move: `tolokaforge run` and `worker` exit 1 when any trial is
`UNGRADEABLE`, with infrastructure aborts deliberately excluded and pinned by a negative lock (#979);
`rejudge --trial` over a grade-less bundle goes 1 → 0, because that bundle is now a named skip rather
than a failure; and an empty discovery goes 0 → 1, because a successful report about nothing is the
defect (#721). `db-service`'s `get_by_id` tightens: a lookup whose Python type disagrees with the
model's declared key type now returns `None` where quote-interpolation accident used to make it
succeed (#921).

### Safety guards

The pay-then-crash hole is closed and the closure was demonstrated, not assumed: with the gate
patched out, a run of the fixture scheduled both trials and spent the tokens before dying at grade
time; the lock asserts the conductor ran nothing (#869). Credential handling gained two
materialisation-time refusals — a pack whose compose declares the engine-owned secrets variable on any
service is refused by name, and a service bind-mounting the project context root or the compose file
itself is refused because the materialised file carries the payload — plus `0600` on the written file
and a fix for a pre-existing aliased-`volumes` leak that could hand the host docker socket to a
sibling service (#1062). The TypeSense API key travels only inside the SecretManager payload, with
runtime registration into the redaction set on both channels, closing a measured verbatim-key log
leak (#927).

### Follow-ups

The still-open milestone issues are listed under [What's next](#whats-next). The ones a reviewer of
*this* branch is most likely to ask about: #1032 (an unmakeable-comparison verdict is evidence about
the pack, and two consumers count it as evidence about the trial), #1019 (`docs/GRADING.md`'s per-key
enforcement prose is tied to the key manifest by nothing), #1018 (the hash family's
`DIFFERENTIAL_INTEGRATION` rows are not re-measured against the service-free path), and #1149 (the
eight environmental unit failures every lane in this branch reports).

## Design walkthrough

The picture worth holding is not the repo — it is the path one trial takes, and the five places this
branch inserted a refusal, a name, or a proof.

**Before the run pays.** The authoring gate had a shape problem: it could only check what it could
resolve, and it could resolve very little. This branch gives it four *layers* — a resolvable answer
plus an honest inability to answer. `HashSourceLayer` lets an adapter report what supplies a hash
source beneath the authored block, so a frozen pack whose golden-actions fixture was lost is refused
before tokens are spent (#940). `SeededTablesLayer` gives `validate` the seeded-table view the run
path already had, so an `id_fields` key that addresses no seeded record fails at authoring time
(#923). The grading-source resolver learned to `stat` the declared path (#869). And the gate reached
inside binders, flagging references that can never be true against a per-operator comparability
table rather than a naive type-difference test (#781). The discipline these share: **the layer
reports facts, the gate decides fatality** — the same division `grading_combine_layer` drew — and an
unresolvable layer draws a `?`, never a false refusal and never a silent clean bill.

**The stack, and registration.** A task-declared compose stack's runner now receives the same
credential payload the engine-built stack's runner already got, through one shared mechanism, so the
nine shipped tasks that pair their own stack with a judge rubric get a judge that can authenticate
(#1062). The TypeSense address becomes a property of the stack, injected once at container creation,
rather than a per-task fact rewritten after startup (#927) — and a plane that cannot be started or
bridged aborts the run before any trial, while a task whose declared plane cannot be made usable is
refused at registration for zero paid turns (#926). The rag-service carries its model in the image
and starts offline in a measured 3.5 seconds, and reports `503 degraded` naming the model where it
used to report `200 healthy` and serve keyword-only results (#697).

**The loop.** Tool-call ids become episode-unique by one rule applied at two altitudes — assigned at
ingestion so the runtime never mints a collision, derived per view so bundles already on disk are
repairable (#979). The user simulator gets its own executor and its own tools, so `requestor: user`
grading is satisfiable end to end for the first time (#688). The budget a tool declares is the budget
the call gets — the host-side constant that pinned every backstop to 30 seconds is deleted from the
whole seam — and a backstopped lifecycle tool is rebuilt from its stored registration context before
serving another call, closing a race that recorded empty output as `SUCCESS` in 8 of 10 measured
follow-ups (#691). Two heuristics that could not do their jobs are deleted with the measurement that
refutes them (#708, #709).

**The verdict.** The two substrates keep converging on shared evaluators. `transcript_rules` becomes
the second `BOTH_SCORE_PARITY` family on the `trace_checks` precedent — one config model, one
evaluator, one fold decision, called at both integration points (#685). `state_checks.hash` becomes a
closed typed block on every read path (#730), its stored literal retires onto a source both
substrates genuinely read (#693), its verdict's binariness becomes a type invariant (#742), and the
fact that the two substrates hash to *different digests while sharing one equivalence relation* is
finally written down as an invariant instead of contradicted in five places (#915). A
`custom_checks` verdict says so in `Grade.reasons`, in one sentence both substrates emit, and the
renderer stops enumerating components by hand (#895).

**And the guard rail over all of it.** The parity manifest is what made the earlier findings findable,
so the tail spends four issues making it unfakeable: lock 15 drives every one of the 26 ledger keys
through a real trial and asserts the recording site filed the outcome the manifest implies (#711);
every parity fixture becomes a truthful task with a real server and committed schemas, so the gate's
argument checks run at error tier instead of being advised about (#767); the version-lock list becomes
one anchored table beneath a 63-key census walked against the runner models in both directions (#762);
and the numeric-string folding promise is measured on both substrates by a six-cell differential whose
control rows are what make it a differential rather than a smoke test (#687).

**After the run.** A directory that records a trial is a bundle — identity is `trajectory.yaml` alone,
in one leaf module all three offline commands share — and every bundle a command is pointed at gets a
named disposition instead of being defined out of the batch (#721). A run directory becomes a
judge-labelled corpus by a rule that runs, with the composition written down and machine-checked, and
one command re-verifies every declared migration in CI (#793).

```mermaid
flowchart TD
    PACK["task pack<br/>grading.yaml · tools · initial_state"]
    LAYERS["resolvable layers<br/>grading source · hash source<br/>seeded tables · replay world"]
    GATE{"authoring gate<br/>validate + orchestrator pre-flight"}
    STOP["no trial — zero tokens spent<br/>finding names file, key, remedy"]

    STACK["stack materialisation<br/>credentials · TypeSense address · rag-service health"]
    REG{"RegisterTrial<br/>tool reconstruction · search plane<br/>id_fields · jsonpath reachability"}
    LOOP["agent + user-simulator loop<br/>episode-unique call ids · declared tool budgets"]
    EXEC["ExecuteTool → db-service / MCP / builtins"]
    LOST["trial_lost → HARNESS_ERROR<br/>counted, no grade"]

    CORE["core GradingEngine"]
    RUNNER["runner GradeTrial"]
    SHARED["shared evaluators and folds<br/>transcript_rules · state_checks.hash<br/>trace_checks · custom_checks"]
    LEDGER["accounted-keys ledger<br/>+ Grade.reasons"]
    PARITY["substrate-parity manifest<br/>lock 15 drives every recording site"]

    BUNDLE["trial bundle<br/>trajectory.yaml is the identity"]
    OFFLINE["rejudge · retrace · reconcile · curate<br/>every bundle gets a named disposition"]

    PACK --> GATE
    LAYERS --> GATE
    GATE -->|refused| STOP
    GATE -->|accepted| STACK
    STACK --> REG
    REG -->|refused| STOP
    REG --> LOOP
    LOOP <--> EXEC
    LOOP -.->|runner no longer holds the trial| LOST
    LOOP --> CORE
    LOOP --> RUNNER
    CORE --> SHARED
    RUNNER --> SHARED
    SHARED --> LEDGER
    LEDGER --> BUNDLE
    LOST --> BUNDLE
    BUNDLE --> OFFLINE
    PARITY -.->|asserts| SHARED
    PARITY -.->|asserts| LEDGER
```

## Key design choices

| Decision | Rationale |
|---|---|
| Redaction runs in the artifact writer over the dumped mapping, never at tool-call record time, and never over the model the grader reads (#694) | A matcher over `args.api_token` must score identically on both substrates; a redacted value cannot be asserted at all, so record-time masking corrupts grading input |
| A rewritten bundle stamps itself, the stamp is a property of the writer (not of `write_metrics`), and every offline grading command refuses a stamped bundle under its own name (#694) | A constraint over a replaced argument fails as *decided* — a confident wrong answer. Judge replay writes a grade and no metrics, so a method-scoped stamp would leave a withheld sidecar undeclared |
| Redaction is off by default; selecting a policy is a constructor argument, the config/CLI knob is #1171 | The corpus already carries raw values on both substrates, and a stamped bundle is offline-ungradeable by design — a silent default flip would cost every recorded run its re-gradeability |
| Ledger recording sites are verified by *driving* a real `RegisterTrial → ExecuteTool → GradeTrial`, and the claim matches the skip-note prefix in `grade.reasons` rather than RPC success (#711) | Deleting a real recording site left the whole canonical suite green, and downgrading a site to a `SKIPPED` record still does. "The RPC succeeded" is too weak a claim to catch either |
| The lock's driver table is hand-written and asserted total by set-equality against the ledger's own key list (#711) | Comprehending the table from `LEDGER_KEYS` would compare a set against itself — the tautology this milestone has now hit sixteen times |
| Parity fixtures become truthful tasks with real servers and committed schemas, not gate exemptions (#767) | Declaring tool names alone passes the gate and then hard-fails `RegisterTrial` for every parity pack. An exemption would make the gate blind exactly where the parity corpus lives, and let a new fixture *declare its way past* the check rather than be caught by it |
| `transcript_rules` becomes one config model, one evaluator and one shared fold decision called at both integration points (#685) | The `dict[str, Any]` seam and the copied logic are what let seven divergences accumulate invisibly; a shared helper makes an events-less trial fold identically by construction rather than by measurement |
| Runner semantics win on all eight transcript divergence classes, including two capabilities core loses (#685) | A rule about what the agent *said* must not be satisfiable by the user's own prompt. The capability that had to go is named in the CHANGELOG with its replacement |
| The wire renames `required_actions[*].tool_name` → `name`; the author key is untouched (#685) | `grading.yaml` spells `name:` today. Renaming the author surface would break external packs this repo cannot sweep, so the rename is version-locked in both directions instead |
| `state_checks.hash` becomes a closed model refused identically at every read path, with the gate delegating "is the hash a state source" to the model (#730) | Worst case measured: writing the runner's own documented field names into `grading.yaml` graded the trial `1.0/PASS` with "All checks passed" while the state was never checked. A raw `bool(.get("enabled"))` read also diverges from pydantic coercion on `"false"`/`"no"`/`"off"` — `validate` refused what loaded cleanly |
| `expected_state_hash` retires *onto* `expect_initial_state` rather than being deleted into a hole (#693) | The literal was the documented refusal-task shape, so plain refusal would make refusal tasks unauthorable. Each substrate hashes both sides in its own algebra, which is exactly what makes the new key portable where the literal was not |
| Resolvers return their basis, and the accounting consumes the returned value instead of re-deriving it (#693, #927) | It is what makes a read whose score coincides with a fallback falsifiable at all: delete the read, the populated key files nothing, `GradeTrial` fails, and lock 15's row reds |
| The two hash algebras stay separate, deliberately (#915) | `compute_stable_hash` backs persisted digests — ETags, snapshots, two wire fields — and core's algebra reproduces recorded-bundle literals, so changing either invalidates digests that already exist. Since #693 removed every crossing route, unification would pay a real migration for a property nothing uses |
| `hash_score` is *derived* from `hash_match` rather than validated (#742) | A `model_validator` is bypassable via `model_construct` / `model_copy(update=)` and stores one bit twice; a getter-only property makes the contradictory state unrepresentable everywhere — the issue's own stated end, reached more completely |
| An adapter reports facts about a hash source (`usable`/`missing`/`empty`); the gate keeps deciding what is fatal, via a classmethod hook (#940) | The same division `grading_combine_layer()` draws. A classmethod because `tolokaforge validate` constructs no adapters and must keep validating packs whose adapter package is not installed |
| A declared-but-absent grading path reuses the existing `WITHHELD` / `UNINTERROGABLE` decisions instead of adding a kind (#869) | Counterfactual-proven: a new `DECLARED_ABSENT` member would have fallen through both gates' non-`WITHHELD` arm into the *permissive* branch — failing open for native packs, recreating the very validate-vs-run drift the issue exists to close |
| The jsonpath classifier names the **divergence set** derived from core's own state composition, rather than denying a single root (#896) | Three of the four in-repo packs the issue's literal rule would have refused meant the filesystem. A root-blind fix would have converted three loud crashes into three silent wrong zeros — strictly worse, because a zero is attributed to the agent |
| The unmakeable-comparison record is per candidate event, three-valued, and windowed (#990) | Measured on a 12-cell matrix: the shipped evaluator is wrong in 6 cells, widening the guard under a union fold goes to 8, the per-candidate fold alone leaves 2, and fold + symmetric guard leaves 0. A two-state record would let a call that merely *omitted* the argument count as evidence, flipping an `absent` constraint to a silent pass |
| `tool_expectations` is agent-scoped, with no new `requestor`-style field (#695) | `docs/GRADING.md` already read agent-only — the doc decided this and the code never implemented it. The escape hatch for actor-explicit intent already exists as `trace_checks` matchers' `executor` field |
| Failed calls converge on the tool's own message, and an empty failure records a shared constant (#717) | Recording `""` is not an option: the wire substitutes it host-side into a third wording. With the texts portable, the two guards scoping `result` predicates to successful calls are lifted — a trace check may now assert *why* a call failed |
| Model-to-table registration reads declared `id_fields`; `get_id()` source parsing is deleted (#922) | Source parsing is the exact guess `id_fields` was introduced to retire, and it was the last source reader in the shipped package. Where the two readings disagree, the direction that preserves today's behaviour is mutant-pinned and the direction that intentionally changes it is documented as a divergence class |
| `get_by_id` verifies every query hit against the fallback scan's own predicate (#921) | The jsonpath dialect coerces symmetrically and its coercion artifacts sort first, so verification makes the query an index over the scan's semantics and never a second, looser definition of equality |
| Search-plane failure has two deliberately different tiers: run abort and registration refusal (#926) | Whole-plane failure costs a run zero trials; per-task failure costs a trial zero turns. Registration is the earliest point where the runner knows both the requirement and the client state |
| Runtime secret registration *replaces* the SecretManager singleton instead of mutating it (#927) | The log-redaction cache is identity-keyed, so a mutated manager keeps serving the stale redaction set and the leak survives. Driven both ways before choosing |
| Credentials are injected into the materialised compose document, not the compose subprocess environment (#1062) | The pass-through alternative needs the payload process-global in `os.environ` during concurrent provisioning and rests the boundary on compose's version-dependent resolution, which no free test can lock |
| The rag-service bakes its model and pins `HF_HUB_OFFLINE=1`, rather than warming a cache or lengthening timeouts (#697) | Measured: a warm cache still makes 32 model-hub round trips at startup. Only offline mode takes the network out of the timed path — 3.5s startup under `--network none` |
| The episode-unique call id is applied at two altitudes — assigned at ingestion, derived per view (#979) | Neither substitutes for the other: derivation repairs bundles already on disk, assignment protects the future and the provider-facing conversation. The rule is the identity on already-unique ids, so the whole corpus is byte-unchanged |
| The run gate counts `UNGRADEABLE`, not "produced no verdict" (#979) | An infrastructure-aborted trial also has no verdict, but was never measured and is already reported per reason. Gating on it would make a single rate-limited trial fail a several-hour run — and teach operators to route around the exit code |
| `TRIAL_NOT_FOUND` raises a typed fault at the client seam rather than returning a status-shaped `ToolResult` (#727) | The response is proof the call reached no tool: the runner answered, and its answer was "no such trial". A result object keeps the fault inside the tool-call record space, where `status: error` matchers can select it and a fabricated `0.0` follows |
| A lost trial is **counted** as `HARNESS_ERROR` but gets **no grade** (#727) | The party that would compute the verdict is the party that lost the trial, so a `0.0` would describe no agent performance. Our own defects stay in the denominator, so a non-zero count is a run-health signal rather than a hidden one |
| The runner resolves the budget the tool declares; the host stops naming a per-call budget at all (#691) | The host-side constant pinned every backstop to 30 seconds, making an ADR-locked 120s shell budget unreachable and inverting a 300s build budget by 270 seconds. The wire keeps the field with `0` as the skew-compat sentinel |
| A backstopped lifecycle tool is rebuilt from its *stored* registration context before serving another call (#691) | The old backstop abandoned the worker thread and the abandoned reader raced the next command — measured at 8 of 10 corrupted follow-ups, including empty output recorded as `SUCCESS`, which satisfies a `required_tools` check that did no work |
| The completion-marker scan is deleted rather than repaired (#708) | `###STOP###` is the simulator's exit token, and no agent-facing prompt anywhere instructs the agent to emit it — zero of the 29 loadable example packs, and all 64 in-repo occurrences simulator-side. `AGENT_DONE` keeps its one structural producer |
| The idle-turn heuristic is deleted rather than fixed the way the issue asked (#709) | The requested repair, replayed over the 34 recorded trajectories in the tree, would auto-fail 23 of 34 known-good trials — thirteen of them scoring exactly `1.0` — because the shape it catches is the ordinary close of a conversation. And a harness-global "the agent must act" rule fails refusal-style tasks by design |
| `tools.user.enabled` is wired rather than deleted, sliced at `num_agent_tools` with no proto change (#688) | The runner half was already live and generic, so deleting would break the `requestor`/`executor` task-pack vocabulary to remove a two-line fix. The skew gate protects only against older engines, so a new proto field would be unguarded in exactly the direction that matters |
| Bundle identity is `trajectory.yaml` alone, in one leaf module all three offline commands share (#721) | The original two-marker predicate would have kept dropping provision-failure bundles — production writes them with no `task.yaml` — repeating the issue's own defect one file over. Found by probing the production writer, not the fixture shapes |
| A one-sided or refuting judge corpus is committed as measured (#793) | The alternative — an arm whose prompt instructs the behaviour — buys a label and a false κ. A corpus that refutes its own candidacy is a finding about the constraint, not a failed stage |
| The container suite fails unless the running container's image tags include the tree's own content-hash ref (#774) | The fixture builds-when-absent and tags both the content hash and `:latest` onto one id, so the only reachable silent state was exactly the stale-`:latest` trap. A green run now *means* "this tree's image graded these trials" |

## Suggested review order

Merge order is roughly dependency order, but the story reads better grouped by what each change is
about. Squash SHAs are on `feat/trace-checks-tail`.

**The guard rail, made unfakeable.** Everything downstream is only as trustworthy as these four.

1. `af1ba2db` — **#899** (#711): parity lock 15 — every one of the 26 ledger keys with a runner field
   is driven through a real `RegisterTrial → ExecuteTool → GradeTrial`, with six committed fault
   injections and four production differentials.
2. `f006ef77` — **#985** (#767): every parity fixture declares the tools its grading names, backed by
   per-fixture servers and committed schemas; a new guard holds the 48 packs outside the gate walk to
   the whole gate.
3. `2e2d504c` — **#989** (#762): the runner-engine version lock becomes one anchored table under a
   63-key census walked against the runner models in both directions.
4. `eefdea6c` — **#1023** (#687): a listed numeric-string field folds alike on both substrates,
   measured by a six-cell differential whose control rows are the point.

**The scoring families converge.**

5. `542b147f` — **#906** (#685): `transcript_rules` becomes one config model, one evaluator and one
   fold decision; runner semantics win on all eight measured divergence classes.
6. `ba7fca30` — **#975** (#695): `tool_expectations` grades the agent's calls, not anyone's.
7. `6a050a2d` — **#977** (#717): a failed call's recorded result text is one text on both substrates,
   which unlocks `result` predicates and bindings on failed calls.
8. `5a6a9c42` — **#999** (#895): a `custom_checks` verdict says so in `Grade.reasons`, and the
   renderer stops enumerating components by hand.

**The hash family, end to end.**

9. `233dba8a` — **#913** (#730): `state_checks.hash` is a typed, closed block on every read path.
10. `67537dc6` — **#945** (#693): the stored expected hash retires onto `expect_initial_state`, a
    source both substrates genuinely read.
11. `5704b4c4` — **#958** (#915): the invariant is named — a hash comparison hashes both sides on one
    substrate, and a digest never crosses.
12. `09893568` — **#956** (#742): the runner's hash verdict becomes a type invariant.
13. `c310b65b` — **#955** (#940): adapters answer the hash-source question, so a frozen pack with a
    lost fixture is refused before tokens are spent.

**The authoring gate reaches further.**

14. `a8678478` — **#961** (#923): `validate` runs the `id_fields`-vs-`initial_state` cross-check with
    the run path's own sentences.
15. `a0983e6e` — **#971** (#869): a declared-but-absent grading path is refused before the trial is
    paid for.
16. `85aff3a8` — **#973** (#916): the custom-checks context resolves a path-declared
    `initial_state.json_db`.
17. `07b63bbd` — **#1012** (#896): a path names the state the trial can reach, and a pack that cannot
    reach it is refused by name.
18. `2ea21c6e` — **#993** (#781): the binder gate flags the never-true classes it holds the evidence
    for, plus a fourth the issue did not name.
19. `546e69a3` — **#1038** (#990): an unmakeable comparison fails the candidate it was read on, not
    the constraint.

**The state plane underneath.**

20. `be88cc1e` — **#965** (#920): an upsert whose record omits its key field is refused, not
    collapsed.
21. `074bb857` — **#968** (#921): `get_by_id` encodes its JSONPath key by type and verifies every hit.
22. `e555570a` — **#970** (#922): model-to-table registration consults declared `id_fields`, not
    `get_id()` source.

**The search plane.**

23. `0b5b8d1d` — **#949** (#926): the search plane fails loudly instead of degrading silently.
24. `1141d086` — **#953** (#927): the TypeSense address is a property of the stack, not of every task.

**Harness faults told apart from agent behaviour.**

25. `7d1e118b` — **#1031** (#979): the trial's tool-call id is episode-unique, and a trial nothing
    could grade fails the run.
26. `b109a4b3` — **#1155** (#727): a trial the runner no longer holds ends as `trial_lost`, not as an
    agent tool error.
27. `941e9544` — **#1151** (#691): the budget a tool declares is the one enforced, and a backstopped
    session is rebuilt rather than poisoned.
28. `2d705b12` — **#1140** (#708): the agent's text is no longer scanned for a completion marker.
29. `6f542861` — **#1146** (#709): the idle-turn stuck heuristic is deleted, with the measurement that
    refutes both it and its requested repair.

**Infrastructure that has to work before any of this is provable.**

30. `ca8b81e7` — **#1137** (#1130): the slim runner image boots again — `DockerConfig` moves out of
    the runner's import closure.
31. `44056c2c` — **#1069** (#1062): a task-declared stack's runner receives the credential payload its
    judge needs.
32. `bc1d5166` — **#1092** (#697): the rag-service carries its model, tells the truth about it, and
    the waits stop budgeting for a download.
33. `64f09f98` — **#1129** (#688): user-simulator tool calls are reachable end to end, and
    `requestor: user` grading is satisfiable.
34. `b7d12552` — **#1123** (#774): a trace-check gate fails a trial inside a real runner container,
    keyless.

**Offline replay and the evidence corpus.**

35. `a5a6029f` — **#1054** (#721): a recorded trial is a bundle, and every bundle a replay command
    sees gets a named disposition.
36. `70ee014f` — **#1084** (#793): a committed judge-labelled corpus behind every declared migration.

**Redaction moves to the artifact layer.**

37. `50a9b34f` — **#1192** (#694): secret redaction happens at artifact-write time — one
    credential-naming vocabulary, a writer-owned `metrics.yaml` stamp, offline commands refusing a
    stamped bundle under its own name (`RedactedBundleError`), off by default. Grading always reads
    the real values on both substrates; the enablement knob is #1171.

**Rebase reconciliation.**

38. `6fd6e204` — the rebase onto `main` (which had meanwhile shipped harness mode, #1083) surfaced
    one semantic collision: `run_harness` named a per-call budget the runner-resolved budgets of
    #691 had deliberately retired, and the orphaned kwarg was silently folded into the recorded
    tool arguments. The harness contract already required that value to equal the tool's registered
    `timeout_s`, so the wire transport was redundant — the harness RPC now names no budget of its
    own, and the transport lock asserts exactly that.

## Verification

**Rebase onto `main`.** The branch was rebased onto `main` at `b1196cc4` before this PR opened.
Two upstream features had shipped sibling implementations of tail work (`_finalise`, the DockerConfig
module move, the tool-budget tests), resolved in `main`'s favour; the tail-only semantics that had to
survive (agent-slice tool metrics, user-side tool tests, ADR renumbered 0033 → 0035 to avoid the
collision with `main`'s harness ADRs) were re-ported, and one genuine collision became commit
`6fd6e204` above. After the rebase: canonical 1533 passed, unit 7289 passed (the eight
`test_persistent_shell_local` failures stay quarantined as #1149). Discovered upstream and filed:
#1195 — whole-`tests/` pytest collection breaks on `main`'s twin `test_outputs.py` fixture basenames
(CI's explicit-directory invocations never see it). `main`'s `test-full` lane was already red before
this PR with six Docker-integration failures unrelated to this branch.

**Lanes at the pre-rebase branch head** (#1192's own verification plus its CI run): canonical
**1494 passed**; unit **7177 passed** plus the eight known environmental failures in
`tests/unit/test_persistent_shell_local.py`, identical at base and tracked as #1149. PR #1192's CI:
lint, wheel-install-gate, test-smoke, and the hygiene review all green. Lint clean; format clean
apart from the pre-existing ruff/black backlog outside these hunks. The canonical lane grew from
1195 at the start of this phase to 1494, and every PR reported its own delta. #1192 additionally
verified its boundary by whole-bundle sentinel probe: under a redacting policy the only artifact
carrying a planted credential is `prompts.yaml`, with all seven rewritten files and both withheld
sidecars named in the stamp.

**Container tier.** Three suites in this branch run against real images over real gRPC:

- #774's `test_docker_grading_trace_gate.py` drives a trace-check gate's satisfying and violating
  trials through `RegisterTrial → GradeTrial → CleanupTrial`, asserting the whole verdict surface
  (`binary_pass`, `score`, the component, `gate_failed`, `failed_gate_ids`, per-constraint tuples and
  the reasons sentence) in ~13s with no API key and no LLM. It was verified in *both* guard states,
  live: a stale `:latest` errors both tests naming the image, its tags, the expected ref and the build
  remedy; with the expected ref tagged, 2 passed in 12.65s.
- #688 adds the first suite in the tree where a real container executes a user-actor tool over gRPC
  and grades a record it wrote itself. Both container suites pass against an image built from that
  tip, with ids printed.
- #1130's boot fix was proven at the image tier: a rebuilt runner container starts and the review read
  the moved module's source back out of the running image, byte-identical to the tree.

**Live service tiers.** #697's integration locks ran against real images: the baked rag-service under
`--network none` reaches healthy in 3.5s and serves the health body; the wait-failure path raises
naming the service with its logs (16.8s); the degraded path was verified with the model cache hidden
(503, `unhealthy` at start-period plus retries) and against the pre-bake image as the negative
control. #1062 proves credential delivery end to end with one docker-marked test in which the runner
container reads back the exact `$`-bearing payload — no API keys, nothing paid. #1023 additionally
observed three of its six differential cells over the wire against the deployed db-service.

**Corpus and differential validation.** `tolokaforge validate` over `examples/native/**`: 29 valid, 0
invalid (#1084); the wider corpus walk reports 29 valid / 2 invalid, the two being the pinned
external-adapter pair. `tolokaforge reconcile --dry-run` over the real tree exits 0 with both declared
rows named. #767's gate walk covers 48 previously-unreached authored packs with two positive controls.
#709's refutation was re-derived independently three times — plan, implementer, two reviewers —
converging on 23 of 34 and thirteen trials at exactly `1.0`. #781's gate suite went 202 → 236 checks.
Every PR reported per-lock falsifier evidence: a patch that reds exactly the rows that own the patched
rule, quoted in the stage reports.

**Paid work.** Deliberately almost none. #793's corpus generation is the only paid stage in the
branch: **$0.28958 total** across two model arms against a `$1.50`-per-arm ceiling. #697's only
network traffic was an 88MB model download at image build time. #774, #688 and #691 are explicitly
`$0` and keyless.

**Deliberately not done, stated rather than implied.** #1123 proves the gate over a host-supplied
message view rather than container-written records — the records-joined variant is unreachable until
#1114 lands, and the module docstring says what is and is not proven. #1069's paid `401 → 200` proof
was deferred to #793's re-run rather than bought twice. #1092 did not re-run the deploy lane wholesale
(its change is a constant lookup, guarded canonically). #742's first deliverable was found already
satisfied at a cheaper tier and this branch records that verification rather than rebuilding it, and
#958 closes a bug-labelled issue prose-only because its behavioural deliverable had already shipped
inside #693 with both falsifiers live. The eight environmental unit failures are pre-existing, reddest
on a clean base worktree too, and are not claimed as green.

## What's next

This is an iteration PR, not milestone completion: milestone 28 stays open with seven issues on it —
#1032, #1019, #1018, #735, #725, #690, and the umbrella #684. The remainder are the tail's own tail: two guard
issues about enforcement prose and integration rows that nothing ties to the manifest, one about how
the replay and migration consumers read an unmakeable-comparison verdict, one documentation
cross-reference, one closed-vocabulary codification, and the protobuf codegen guard.


azorej:
**CodeQL triage** — the check reports "5 new alerts (5 high)", with its own caveat that the diff was too large for precise attribution. Walked all 12 alerts the PR query returns:

- `tests/unit/test_orchestrator_typesense_startup.py:207` (clear-text logging) — deliberate: the test logs a fake key through the production logger precisely to assert the redaction set scrubs it; the comment above the line says so. The value is a fixture constant, not a credential.
- `tests/unit/test_logging_formatter.py:393`, `test_system_prompt.py:124`, `test_diff.py:141-142` — same fixture pattern or URL-substring heuristics in tests; none of these files are touched by this branch (0 branch commits).
- `tolokaforge/env/json_db_service/app.py:343-402` (sql-injection) and `mock_web_service/app.py` (XSS) — task-environment mock services: identifiers come from the task pack's own `initial_state`, values are parameterized, and the service runs inside the trial's sandboxed container. The pattern predates this branch; the branch's one commit to that file only touches the keyless-upsert refusal.

Recommend dismissing the test-fixture alerts as "used in tests" and the env-service alerts as "won't fix (sandboxed mock)" — or say the word and a follow-up issue can track hardening the mock services properly.

balazs-toloka (commented):
## Review: 15 findings, 1 blocker

Reviewed the ~9k lines of production diff under `tolokaforge/` (105 files) across 13 independent lenses, then adversarially verified every candidate that could change the merge decision. Five sizeable candidates were **refuted** during verification and are listed at the bottom so nobody re-litigates them.

Line numbers are at head `6fd6e20`. Where I say "measured", I ran it.

---

## Blocker

### 1. `runner/service.py:456` refuses a `$.filesystem` assertion the runner grades correctly

`_unreachable_state_checks_refusal` classifies every `$.filesystem[...]`-rooted `path:` as unreachable, and its message asserts the runner "composes db and tables from the trial's database and nothing else, so this assertion can never match there".

Both halves of that sentence are false about the same file. `service.py:1683` composes `"filesystem": self._read_agent_visible_filesystem()`, and `runner/grading.py:358` resolves against that whole dict. Measured at head, with `AGENT_WORK_DIR` pointed at a directory holding the asserted file:

```
filesystem keys: ['/env/fs/agent-visible/buggy_math.py']
RUNNER SCORE : (1.0, "State: PASS: $.filesystem['/env/fs/agent-visible/buggy_math.py']")
NEG CONTROL  : (0.0, "State: FAIL: ... expected contains 'NOPE', got 'return amount * (1 + tax_rate)\n'")
```

A working negative control, so this is not an accidental pass. The refusal fires identically with and without a seeded database, and `path_glob:` is the only escape:

```
A) filesystem-only pack, no tables:  <refusal>
B) SAME assertion, database seeded:  <refusal>
C) path_glob variant:                None
```

**What makes this a blocker rather than a bug report.** The change rests on a measurement in CHANGELOG.md:135: "a `$.filesystem`-rooted `path:` on a task that seeds a database scores **1.0 on the core engine and 0.0 on the runner**". Case B above is that exact shape and the runner scores 1.0, so I cannot reproduce the 0.0. And the second half of the same justification, "on a task that seeds none, `GradeTrial` returned `Grading error: TrialNotFoundError`", is **fixed by this very PR**: the new `fetch_db=any(addresses_the_database(check) for check in jsonpath_checks)` argument to `_assemble_jsonpath_state` means a filesystem-only pack no longer touches the DB client at all. So the diff repairs the reason the shape failed, and then adds a refusal that makes the repair unreachable.

Blast radius is the whole trial, not one component: `service.py:1791` returns `success=False` before any component runs, `shared_stack_runtime.py:585-590` leaves `grade=None`, and `trial_grader.py:226` raises `GradingFailedError`. In-tree packs were migrated by this PR, so the exposure is packs maintained elsewhere.

`docs/GRADING.md:1270-1274` still carries the merge-base promise that the now-refused shape grades, unmodified by the diff.

Either the measurement needs re-stating or `filesystem` needs exempting on the runner. I don't think the branch should land until it is clear which.

---

## Confirmed, would hold for

### 2. `runner/service.py:1580` loses the tool-call record when a backstop rebuild overruns the RPC slack

`ExecuteTool` submits with `timeout=timeout_seconds + _TOOL_TIMEOUT_SLACK_S` (5.0, `service.py:195`), and the new `except asyncio.TimeoutError` handler awaits `_reset_backstopped_tool` inside that remaining 5s. `trial_context.record(...)` sits after the try/except, not in a `finally`. `_run_async` calls `future.cancel()` then re-raises, and its own docstring says cancel "cannot interrupt a coroutine already blocked inside a `run_in_executor` call" which is exactly where `_rebuild_session` blocks.

For a `PersistentShellToolWrapper` with `tool_config.service` (`rebuild_clears_backstopped_state` defaults True), the rebuild worst case is `proc.wait(5)` + `open()` 15s + `_seed_cwd` 15s, the same ~35s the `BACKSTOP_GRACE_S` docstring names. That grace is inside `timeout_seconds`, not added to the submit deadline, so the rebuild window is 5s flat.

Reproduced with `stop()` sleeping 8s at `timeout_seconds=0.5`:

```
elapsed = 5.51s
status  = 2 (ERROR)   # EXECUTION_STATUS_TIMEOUT is 3
error   = 'Internal error: TimeoutError'
record entries = []
```

Scoping correction worth stating: the host-side `TrialToolCallRecorder` (`core/loop.py:436-444`) records every agent call unconditionally, so `trajectory.tool_log` keeps it and core-side grading is unaffected. The loss is confined to the runner-side record that runner-side `GradeTrial` reads, which is the docker path. There, `_check_disallowed_tool` reports a forbidden call that actually ran as never called and passes (`records_present` is whole-timeline, and `_executed_by_agent` filters `status is not None`).

The baseline had a plain string assignment on that line, so this PR introduces the only cancellation point on the timeout path. No test exercises a rebuild that overruns the slack: every backstop test in `tests/canonical/test_lifecycle_tool_survives_its_backstop.py` uses a local session whose rebuild is milliseconds, and `test_rebuilding_a_session_does_not_stall_the_runners_event_loop` awaits `_reset_backstopped_tool` directly from the test's own loop, bypassing `_run_async` and its deadline.

Also, at RPC return the tool is caught mid-rebuild with `unusable_reason` unset, so a call arriving in that window raises `ToolExecutionError("bash session not started")` with no refusal naming why.

---

## The `tools.user.enabled` cluster

Four findings, one root cause: this PR makes `tools.user.enabled` a real registered surface for the first time (the deleted line read `user_tools: list[ToolSchema] = []  # always empty`), and three gates still reason over the **union** inventory while their evaluators read the **agent** only. `_check_required_action_names`, added in this same PR, does consult `inventory.declared_by(executor)`, so the pattern was known.

No existing pack is affected, because nothing uses the surface yet. But the feature would ship with three broken gates.

### 3. `adapters/native.py:81` lowers a user tool source-less, and can silently substitute a builtin

`_actor_tool_schemas` reads `mcp_server` from the actor's own block with no fallback, and `task_config.py:450-455` permits a user block without one (it refuses only two *different* names). Its own message directs authors here: "Point both blocks at one server, or declare the user's tools as builtins."

Reproduced with `tools.agent: {mcp_server: mcp_server.py, enabled: [get_order]}` plus `tools.user: {enabled: [calculator]}`:

```
tolokaforge validate  ->  Summary: 2 valid, 0 invalid
user wire tool: {"name":"calculator", "source":null,
                 "parameters":{...,"properties":{"expression":{...}}}}
inventory:      {"calculator": {"properties":{"expression":...}}, ...}
reconstruct:    BuiltinGenericToolWrapper -> tolokaforge.tools.builtin.calculator.CalculatorTool
```

The pack's own `fixtures/tools.json` declares `calculator` as `{order_id, op}`; the builtin's `{expression}` won on both the wire and the argument-matching inventory. So the simulator executes the wrong implementation and `executor: user` argument rules grade against the wrong schema. 8 of the 14 registry names (`bash`, `calculator`, `browser`, `http_request`, `db_query`, `db_update`, `read_file`, `write_file`) can shadow an MCP tool this way.

For a name with no builtin the failure is loud instead: `ToolConfigurationError("Tool has no source configuration, cannot reconstruct")` at `RegisterTrial`, for every trial, on a pack `validate` passed.

Suggested fix: `user_block.get("mcp_server") or agent_block.get("mcp_server")`, or refuse a `tools.user.enabled` naming a tool the block cannot resolve.

### 4. `core/grading/config_validation.py:1807` accepts a golden action naming a user-only tool

`_check_golden_action_names` refuses only when the name is absent from `inventory.declared`, which is `agent_declared | user_declared` (`_task_loader.py:779`). The runner resolves against the agent registry alone (`service.py:2667-2671`, `candidates=trial_context.agent_tools.keys()`). So `golden_actions: [{name: <a user tool>}]` passes `validate`, the trial runs to completion, `_execute_hash_grading` step 0 raises `UnresolvableGoldenAction`, and `GradeTrial` answers `success=false`. That is the exact cost the docstring at 1792-1795 says the gate exists to avoid, and its claim to be "stricter than either replay substrate" is now false in this direction.

`inventory.declared_by(ToolExecutorIdentity.AGENT)` is available for every native pack via `actor_split_known=True` (`_task_loader.py:797`).

### 5. `core/grading/config_validation.py:1143` checks the union while the evaluator was narrowed to agent calls in this same PR

`transcript.py:244-248` (renamed `_executed` to `_executed_by_agent` here) now filters `call.executor is ToolExecutorIdentity.AGENT`, but `_check_tool_expectation_names` still tests `inventory.declared`. So `required_tools: [<a user tool>]` passes the gate and fails on **every** trial, indistinguishable from the misspelling the gate exists to catch; and `disallowed_tools: [<a user tool>]` passes on every trial even when the simulator calls it each time, so the prohibition asserts nothing.

The docs contradict themselves three paragraphs apart: `docs/GRADING.md:1646` says "Both lists read the agent's own calls", `:1665-1667` states the gate's contract as "a name **no actor of the task can call** is an authoring error".

### 6. `core/runner.py:758` bootstrap guard is dead for the case its own comment names

The new `if not first_user_result.text.strip(): raise RuntimeError(...)` cannot fire for a tool-call-only opening, because `llm/client.py:2468` already substituted `"Let me check that."` inside the closure `reply_guard.enforce` invokes. Measured end to end with a fake provider returning `text="", tool_calls=[lookup_order]`:

```
first_user_result.text = 'Let me check that.'
guard fires: False
opening: 'Let me check that.\n\nlookup_order() result: order A-1: shipped'
```

So the engine's own filler becomes the task statement the agent is graded against. This is newly reachable: the filler's comment says it is "Unreachable while the conductor wires no `user_tool_executor`" and carries `TODO(#1089)` calling a universal filler hazardous, and `conductor.py:618-622,697` wires one where the merge base hardcoded `None`. Two comments are now stale as a result, neither file touched by the diff: `core/actors/reply_guard.py:10-11` still asserts "the conductor always passes `None`", and the `client.py` comment above.

The guard does still fire for an empty reply with no tool calls, which is the only shape `tests/unit/test_runner_logic.py:864-880` covers, and both bootstrap tests drive a `MagicMock()` simulator so the real filler is never in the loop.

---

## Confirmed, lower severity

### 7. `core/conductor.py:590` drops the trial out of every metric instead of accounting for it

The typed search-plane refusal crosses as free text and is raised as a bare `RuntimeError`, which is not a `ProvisionError`. `Conductor.run` has no `try` around `_setup_trial`; `ProvisioningTrialExecutor.execute` catches only `ProvisionError` and only around `provision`/`await_ready`, while the `conductor.run` call at `trial_executor.py:176` sits in `try/finally` with no `except`. It lands at `orchestrator.py:2351`, which does queue and state bookkeeping only and appends nothing to `self.results`. Since `_generate_reports` computes every metric from `self.results`, the trial produces no row at all: absent from `total_trials`, `measured_trials` and every rate, after `retryable=True` burns the full retry budget on a deterministic refusal.

By the codebase's own tie-break (`failure_attribution.py:67-78`) that is the worse direction: it "inflates every benchmark number with nothing in the output to show it". A typed `ProvisionError` would let `trial_executor.py:141` synthesise a `PROVISION_ERROR` trajectory that shows up in `infrastructure_aborts`.

### 8. `runner/service.py:3385` gates the "no plane serves this corpus" refusal on an address that cannot resolve

`no_plane_serves_the_corpus = task_declares_kb and address_resolved and resolved_plane is None`. But `resolved_plane is None` implies both `search_config.plane` and `search_config.host` are None, so `address_resolved` could only come from the stack env, which is injected only when the run has a TypeSense plane. For a knowledge-base task in a run with no TypeSense, nothing registers and nothing refuses; every `search_policy` call fails on paid turns and grades as agent behaviour. The refusal's own text says the task "carries no connection details to derive it from", a state under which the guard excludes it.

Related and also silent: a fully migrated task declaring `plane: typesense` with no host, in a run without the stack env, registers live with no client. `_typesense_stack_kwargs` derives the env from run config only, and nothing cross-checks a task's declared plane against the run having one. `docs/ADAPTER_INTERFACE.md:139-141` tells adapters the migrated steady state is to "emit no connection details at all", so every migrated task depends entirely on the run YAML, and the doc's claim that the runner "refuses at `RegisterTrial`" does not hold in a run with no plane.

### 9. `core/grading/replay.py:894` lets one redacted bundle abort a whole rejudge batch

`read_replay_inputs` now raises `RedactedBundleError`, documented at `core/output/artifacts.py:136` as "Deliberately **not** a `ValueError`", so it matches neither arm of `except (MissingReplayInputError, ValidationError)`. Reproduced on a source holding one clean and one redacted bundle: the exception leaves `run_replay_batch` with zero outcomes returned. On a real 200-bundle rejudge that means the judge spend for every bundle sorting earlier is real and `_print_rejudge_summary`, `emit_replay_report` and the new `ReplayBatchCensus` never run.

Contradicts the function's own docstring ("a named per-trial failure and the batch continues (no silent skips)") and the two sibling commands reading the same bundles: `trace_replay._load_trajectory` converts it per bundle to `TraceReplayFailure.REDACTED_BUNDLE`, and `corpus_curation.py:350` catches it per bundle. `ReplayOutcomeStatus.FAILED` already has a slot for it.

Not reachable today, since nothing constructs `SensitiveKeyRedaction` and the CHANGELOG defers the switch to #1171, so this can land with the redaction switch rather than blocking here.

### 10. `core/grading/config_validation.py:1579` gives the gate the false-reject mode it documents itself as not having

`_check_id_fields_against_seeded_tables` calls `id_fields_findings` before reading `relaxed_validation`, and `table_key` (`runner/id_resolution.py:66`) raises `IdFieldResolutionError` for an invalid key component. Verified by calling `_check_id_fields_against_seeded_tables` directly with `state_checks: {relaxed_validation: true, id_fields: {orders: ["ok", ""]}}`: the exception propagates out of `inspect_grading_authoring` past `report.fatal(fail_on)`, so the finding never enters the errors/advisories channels, the `fail_on` severity is bypassed, and the `relaxed_validation` branch three lines below at 1581-1587 is unreachable. `orchestrator.py:1732` turns it into an unconditional per-task refusal.

### 11. `runner/models.py:730` and `:820` drop the #717 guards while two failure classes still diverge

Both `_require_a_success_status_beside_a_result_predicate` and `_require_a_success_status_beside_a_result_extraction` are gone, and `status` gets no enum-membership validation anywhere, so `status: {equals: timeout}` loads clean. But the convergence traded for them only unified the tool-authored text:

- `ToolExecutionStatus.TIMEOUT` is never assigned anywhere in `tolokaforge/` except `service.py:1563`; `registry.py:371-374` states why ("This executor runs a tool to completion: it applies no band of its own"). So `status: timeout` is unproducible in-process.
- In-process invalid arguments are `"Invalid arguments: {e}"` / `"Tool '<name>' takes no arguments but received: ..."` from jsonschema (`registry.py:445,459`); the runner's only site is `"Invalid arguments JSON: {e}"` from a `JSONDecodeError` (`service.py:1320`), and `grep -rn jsonschema tolokaforge/runner/` returns nothing.

Separately, `result` differs by view: `_append_result` uses `record.output` (no prefix) while `_append_message_result` uses the message `core/loop.py:460` wrote as `f"Error: {resolve_tool_output(...)}"`, and nothing strips it. So `result: {regex: "^insufficient funds"}` passes on a bundle carrying `tool_log.yaml` and fails on the same bundle re-graded without it.

The CHANGELOG asserts "One failure now has one text on both" enumerating four classes; neither `timeout` nor invalid arguments appears, and `tests/canonical/test_trace_timeline_substrate_parity.py` has zero hits for either.

### 12. `core/runner.py:378` prices a typed provider fault as a scored agent failure

The new `TerminationReason.TRIAL_LOST if isinstance(e, TrialNotRegisteredError) else TerminationReason.ERROR` sits in a catch-all guarding `_seed_first_user_message`, which makes a real LLM call (`reply_guard.py:414-420` re-raises with a bare `raise`, `client.py:2029-2035` passes `LLMApiTimeoutError` through untouched). `ERROR` maps to `HARNESS_ERROR`, stays in `measured`, and `trial_grader.py:181-193` returns `Grade(binary_pass=False, score=0.0)`. `self.agent_client.classify_loop_error` is in scope 22 lines earlier and would have returned `API_TIMEOUT` or `RATE_LIMIT`, both in `EXCLUDED_TYPED_REASONS`. In probe mode `init_attempts` is 1, so one typed 429 on the opening generation is scored while the identical 429 one turn later is excluded.

The mispricing itself is pre-existing (the merge base was unconditional `ERROR`), but this is the PR whose theme is that rule, and its own test names the duplication: "the simulator's opening call runs before the loop, under a different `except`, so a fix applied to one route leaves the other visibly open" (`tests/canonical/test_lost_trial_attribution.py:14-16`).

### 13. `runner/db_proxy.py:554` narrows `except Exception` to `except DBServiceError`

`get_by_id`'s docstring at 442-444 promises "Every kind of miss (no literal, a query error, an unverified hit, no usable key to query at all) falls to the scan, which is why the query can only ever save a round trip, never change an answer." But `db_client.query` only wraps `httpx.ConnectError`; `httpx.ReadTimeout`, `httpx.RemoteProtocolError` and a `QueryResponse.model_validate` failure now escape `get_by_id` and fail the tool call instead of falling through to `get_all`, a different endpoint that may be healthy. Every branch below already means "fall to the scan".

### 14. `secrets/log_filter.py:59` misses the form the credential actually travels in

`_current_redaction_values` unions the raw and compose-escaped forms, but `container_secrets_env()` (`manager.py:433-436`) returns `json.dumps(payload)` and the transform writes `compose_escaped(...)` of that whole JSON string. For a value `ab\cd`, `json.dumps` writes `ab\\cd` and `compose_escaped` changes nothing, so neither scrub entry is a substring of what is on disk. A log line quoting the materialised file back prints it verbatim, which is the leak the `compose_escaped` union was added to close (`manager.py:418-419`). Non-ASCII values miss the same way via `\uXXXX`. Narrow trigger, since most keys are alphanumeric, but the fix is small: add `json.dumps(value)[1:-1]` and its compose escape.

### 15. `dx/cli/main.py:953` leaves the FAILED branch unescaped

The diff adds `escape()` to the adjacent no-grade branch with the comment explaining exactly why, and leaves `console.print(f"[red]failed[/red] {rel} — {outcome.reason}")` where `reason` is `str(exc)` for a `MissingReplayInputError` or a pydantic `ValidationError`. Reproduced against this tree's own `rich`:

```
'failed b1 - expected [/orders] segment'
  -> MarkupError: closing tag '[/orders]' at position 21 doesn't match any open tag
'failed b2 - Tool 'x' not found in [agent tools]'
  -> prints "failed b2 - Tool 'x' not found in"   (fragment silently deleted)
```

The second wording is what `testing/certify/_capability.py:235` is being edited away from in this same PR, so it is live in recorded bundles. Same omission at `dx/trace_replay_render.py:87`. `rel` is also interpolated unescaped in all four branches, though `_fail_on_ungradeable_trials` at 479-481 escapes trial ids with the comment explaining why.

---

## Refuted during verification

Listed so they don't get re-opened. Each of these looked like a finding and is not.

1. **Phrase-rule evidence set (`core/grading/transcript.py`).** Core does narrow from USER+ASSISTANT+TOOL_RESULT case-sensitive to assistant-only case-insensitive, but the runner was **already** assistant-only and case-insensitive at the merge base (`runner/grading.py:324,435-436,456,465`), so no docker-graded run's verdict moves. Both changes are named at CHANGELOG.md:176 and :177, and `tests/canonical/test_transcript_substrate_parity.py` pins the exact rows including `disallow_regex_only_in_a_tool_result` expected 1.0. Passes. `_TRANSCRIPT_PHRASE_REASON`'s deletion records the resolution structurally, via the `BOTH_SIGNAL_PARITY` to `BOTH_SCORE_PARITY` upgrade and both evaluator fields collapsing to one.

   One residual: `examples/native/rag_search/.../kb_lookup_01/grading.yaml` (`must_contain: ["HX49-QORVEN-7731"]`) and `examples/native/mock_web_booking/.../booking_01/grading.yaml` (`must_contain: ["BKSEA12345"]`) declare tokens that arrive in a tool result, so their core-side verdict does move and neither is in the CHANGELOG's move list. The direction is a tightening toward each pack's intent, so it removes a false pass rather than creating a false fail, but the list is incomplete.

2. **`extra="ignore"` to `extra="forbid"` on `RequiredAction` / `CommunicateInfo`.** Real, and fully documented, by this PR: CHANGELOG.md:185 names the flip, :183 names the wire rename with the both-directions lock, and `docs/GRADING.md:1157` carries the version-lock row while :1189 confirms "nothing in a task pack migrates". No coercer is needed because the authored key was already `name:` at the merge base. Independent scan of 275 pack files: 42 `required_actions` elements, 16 `communicate_info`, zero stray keys.

3. **`custom_checks.fail_on_error` unread on the core substrate.** The asymmetry is real (core hard-codes `0.0`, the runner uses `-1.0` which `shared_stack_runtime.py:196-201` treats as not-evaluated) but **byte-identical at the merge base**. The deleted `result_to_score` was dead code, called only from two unit tests and an ADR. Worth a follow-up issue and a `reason` on the `key_manifest` entry, not a change here. The parity suite drives `custom_checks` only through suites that run successfully, so nothing covers it.

4. **Compose credential-exposure refusal ordering.** The refusal does run before the empty-payload early return, and `.`/`./` do match while `_check_safe_bind_mounts` accepts them. All of that is deliberate, asserted by name in `test_a_pack_declaring_the_variable_is_refused_without_any_secrets` and `test_a_bind_mount_that_would_read_the_payload_back_is_refused`, and stated in `docs/SECURITY.md:25`. Zero in-tree compose docs affected. (One factual note for the docs: `SharedStackRuntimeBackend._materialise` is per-run, not per-trial.)

5. **Trace-timeline two-assigner occurrence keying.** Both mis-pairing mechanisms reproduce in isolation, and neither is reachable: there is one trial-scoped `EpisodeUniqueCallIds` (`core/runner.py:159`) shared by the agent loop (`:353`) and both user-side paths (`:962`), applied before the call is executed, recorded or written onto the message; and within a turn `sequence` cannot invert, since it is a monotone append counter and execution is strictly sequential in declaration order (no `gather`/`create_task` anywhere in `core/loop.py` or `core/runner.py`). Pre-PR bundles cannot carry a user turn with `tool_calls` at all.

Also checked and found sound: the regenerated `runner_pb2.py` matches `runner.proto` field-for-field across all 24 messages (`precomputed_expected_hash` gone, `reserved 3` present); the version-lock census genuinely walks the models rather than a hand-written list, and it caught this PR's own new `search.plane` key; every call site repo-wide binds against the new signatures of the changed callables; and the secrets contract holds, with the only `os.environ.get` calls in the diff being non-credential.

---

## Missing regression locks

Four invariants the branch relies on and does not pin. Two of them would have caught findings above.

1. A backstop rebuild that **overruns** the RPC slack (finding 2). Every existing test rebuilds in milliseconds or bypasses `_run_async`.
2. A `fail_on_error: false` erroring suite graded on the core substrate (refuted item 3).
3. The trial-scoped `EpisodeUniqueCallIds` is shared across both actors and applied before write/record (refuted item 5).
4. Within a turn, `sequence` order equals declaration order (refuted item 5).

---

## Cleanup, not blocking

Happy to file these separately if useful:

- `config_validation.py:137` has five layer dataclasses each hand-writing the same `known` flag, `unresolvable()` classmethod and "carries facts while known=False" refusal, about 90 lines where only two invariants are genuinely per-layer. A sixth layer is a fresh paste.
- `config_validation.py:1932` re-decides "closed schema is an error, open is an advisory" at four call sites, three of them textually identical. That rule sets an author's exit code.
- `service.py:409` duplicates `config_validation`'s refusal templates and its "name the offending assertion" derivation, and the wording has already drifted: `_ADDRESS_THE_DATABASE` is byte-identical across the two files, but the file-glob remedy reads "the pairing both substrates read" in one and "which both substrates read the same way" in the other. `jsonpath_addressing.py`, added here and imported by both, is not in `RUNNER_SUBSET_EXCLUDED_FILES`, so it can hold the shared copy.
- `rubric_migration.py:2012` calls `_resolve_pack` once per declared sidecar, and each call `rglob`s and YAML-parses every `task.yaml` under every pack root. Quadratic in declared migrations: a 2000-file pack root with 50 sidecars is ~100k parses before any evidence is weighed. One index pass over `roots` fixes it.
- `combine.py:382-391` hashes the whole `initial_state.json_db` with `state_digest` on every trial, though the digest is a per-task constant and `adapters/base.py:570` builds a fresh engine per trial. Measured ~35ms per trial on a 1.4 MiB seed. `:566` then reads the same file a second time in the same `grade_trajectory`.
- `trace_checks.py:295-315` lost `_resolve`'s short-circuit, so every predicate is evaluated on every candidate and `_ComparisonRecord`s are built even for matchers with no binding operator, where the record pass is provably discarded. Measured 1.59x on a non-matching event.
- `replay_layout.py:35` declares a second constant for `"trajectory.yaml"` instead of importing `TRAJECTORY_FILENAME`, in the module that defines bundle identity; `:45` omits `RECONCILE_DIRNAME` from `RESERVED_DIRNAMES` though `reconcile` is the third command adopting this discovery here.
- `corpus_curation.py:92` hardcodes `"tools_schemas.yaml"` beside two imported constants although `output_writer.py:59` defines `TOOLS_SCHEMAS_FILENAME`. Copies are `if exists()` only, so a rename silently drops the file from every curated corpus.
- `dx/cli/main.py:930` is a fourth private copy of "bundle path relative to source", and two of the four disagree on the fallback. `replay_layout.py` was created for exactly this question.
- `docs/ADAPTER_INTERFACE.md` documents neither `own_budget_s` nor `rebuild_clears_backstopped_state`, both new public override points on `ToolWrapper`, and its "Current limitations of the lifecycle contract" list is now incomplete: `service.py:1519-1520` reopens a lifecycle tool mid-trial. Its two sibling extension points each have a dedicated section.
- `orchestrator.py:473`: `grading_completeness` has no reader anywhere in the repo, so the fail-loud intent is currently inert.
- `core/output_writer.py:99` (and `output/artifacts.py:367`, `logging.py:378`) default `redaction: RedactionPolicy = NoRedaction()`, evaluated once at definition time. Harmless while both policies are stateless frozen dataclasses, but the annotation is a `Protocol` and the docstring two lines below warns against exactly this sharing.

---

Method note: 13 finder lenses over the production diff, 10 adversarial verifiers, and a final gap sweep. Everything above marked "measured" or "reproduced" was executed against `6fd6e20`.

CiroGamboa:
Follow-up to @bberkes-toloka's review — I've applied twelve of the fifteen findings on top of your branch as PR #1227 (base: `feat/trace-checks-tail`). Merge that first and only #6, #7, #11 will need your read on this PR.

**Applied — twelve findings**

| # | File | What changed |
|---|---|---|
| 1 (blocker) | `jsonpath_addressing.py` + `service.py` + `config_validation.py` | `$.filesystem[…]` paths grade instead of refuse; the runner already reads them. |
| 2 | `runner/service.py` | `ExecuteTool` outer `try/finally` so the record survives RPC cancellation of the slack coroutine. |
| 3 | `adapters/native.py` | USER actor falls back to the agent block's `mcp_server` when the user block has none. |
| 4, 5 | `config_validation.py` | Golden-action / tool-expectation gates use `declared_by(AGENT)` only when `actor_split_known`; fall back to `declared` when the split is unknown. |
| 8 | `runner/service.py` | `no_plane_serves_the_corpus` drops `address_resolved`; the `mode: disabled` shape now refuses loudly. Locked by `test_a_kb_task_in_a_no_plane_run_refuses`. |
| 9 | `core/grading/replay.py` | `RedactedBundleError` joins the except tuple. |
| 10 | `config_validation.py` | `_check_id_fields_against_seeded_tables` catches `IdFieldResolutionError` from `id_fields_findings`. |
| 12 | `core/runner.py` | Init branch routes non-`TrialNotRegisteredError` through `classify_loop_error` so opening-generation faults price like mid-loop ones. |
| 13 | `runner/db_proxy.py` | Reverts to `except Exception` + `# noqa: BLE001` — the block is defensive around a broader class than `DBServiceError`. |
| 14 | `secrets/log_filter.py` | Scrub set unions three surfaces (raw / compose-escaped / JSON-encoded). |
| 15 | `dx/cli/main.py`, `dx/trace_replay_render.py` | Escape rich markup + bundle path in FAILED branches. |

**Deferred — needs your read**

- **#6** — Bootstrap guard for `tools.user.enabled`; needs your call on how the union should be evaluated once the new user-tool wire is fully consulted.
- **#7** — Conductor error accounting across the new entry-points; needs your view of the intended attribution.
- **#11** — Dropped `#717` guards on the trace-checks path; the intended replacement needs your read on which guards were load-bearing.

Full sweep is green locally on the fixup branch: 3417 passed, 1 skipped across `tests/unit` + `tests/canonical`; ruff clean.

CiroGamboa:
Follow-up: PR #1227 now covers **every** finding — @azorej is out and this branch is the whole path to `main`, so the three deferred earlier are landed here too.

**#6 — bootstrap guard.** `GenerationResult.filler_substituted` flag; `_bootstrap_via_simulator` refuses when the LLM client substituted `"Let me check that."` for a tool-call-only reply. Stale comments at `client.py:2464` and `reply_guard.py:11` corrected. Locked by `test_a_filler_substituted_bootstrap_refuses`.

**#7 — conductor dropping trials from metrics.** `conductor.py:590` raises typed `ProvisionError(stage="register_trial")` (new literal on `ProvisionStage`); `trial_executor` catches it around `conductor.run` and routes through `_synthesize_provision_failure_result` — same path as `provision`/`await_ready` failures. Refusal is non-retryable per the existing classifier. Locked by `test_register_trial_refusal_synthesises_failed_result_and_tears_down`.

**#11 — dropped #717 guards + `Error:` prefix drift.**
- `TraceMatcher.status` validates literal values against `ToolExecutionStatus` for closed-vocabulary operators (`equals`, `not_equals`, `in_`, `not_in`) at load time.
- `_append_message_result` strips the `"Error: "` prefix `core/loop.py` writes onto a failed call's message body, so `result:` matchers see one text on both substrates (#977).

Locked by four new tests across `test_trace_checks_matchers` and `test_trace_timeline`.

**Left for follow-up.** The reviewer's sibling flag on #11 — `TIMEOUT` and `INVALID_ARGUMENTS` classes still write different text in-process vs runner-service — needs a wider substrate-parity pass than this PR. Called out explicitly in PR #1227's body.

Full `tests/unit` + `tests/canonical` sweep green on the fixup branch; ruff clean.

PR #1214, opened 2026-08-15
fix(runtime): reap stale named containers before docker compose up

CiroGamboa:
## Summary

- Fix reliability bug: a prior run's containers can survive a killed session or crashed docker daemon and refuse the next \`docker compose up --wait\` with an "already in use" conflict.
- Introduces \`reap_stale_named_containers(compose_file)\` — reads \`container_name\` values from a per-trial compose file and force-removes any that exist. Called right before \`compose.start()\` in \`PerTrialRuntimeBackend.provision\`.
- Silent no-op when nothing is stale; unreadable/absent compose files are also silent no-ops.

## Failure mode this fixes

Repro (matrix run from this Friday's autonomous session):

1. Engine-loop run of \`fix-billing-holds\` creates container \`tbench_fix-billing-holds_0_main\`.
2. Session interrupts before compose teardown — container survives.
3. Next CLI-harness run of the same task provisions a compose file with the SAME fixed \`container_name\`.
4. \`docker compose up --wait\` fails immediately with \`Conflict\`; trial errors with \`provision_error\` and no LLM call is made.

Prior symptom: 4 out of 8 matrix cells (all \`fix-billing-holds\` under any CLI harness after the first engine-loop cell had run) errored at provision time with 0 latency and \$0 spend. Airline-segmentation cells were unaffected because no prior run had claimed their name.

## Design choices

- **Reap by name, from the compose file itself.** The container names live in the compose file the runtime is about to invoke. Reading them there means the reaper can never target a container this run does not intend to own — the trial's own trial_index is the discriminator.
- **Force-remove is safe here.** Terminal-bench trials are single-writer per (task_id, trial_index): a running container by the expected name is by construction a stale one from a prior run. The reserved trial_index prevents in-flight collisions across concurrent workers.
- **Silent no-op on absence.** Docker's \`No such container: X\` is the healthy path, not an error condition. Only successful reaps are logged (INFO), so a clean run stays quiet.

## Verification

- New behaviour-locking unit tests in \`tests/unit/test_compose_materialisation.py::TestReapStaleNamedContainers\`:
  - reaps every named container declared, in sorted order
  - missing container is a no-op
  - services without \`container_name\` are skipped
  - unreadable/absent compose file is a silent no-op
- \`uv run pytest tests/unit/test_compose_materialisation.py\` → 74 passed
- \`ruff check\` + \`black --check\` clean

## Test plan

- [x] Unit tests green
- [ ] Full CI green on this PR
- [ ] Manually re-runs the matrix cell that hit the stale-name conflict and observes it succeeds

### DIFF
--- a/tests/unit/test_compose_materialisation.py
+++ b/tests/unit/test_compose_materialisation.py
@@ -35,6 +35,7 @@ from tolokaforge.core.compose_materialisation import (
     first_published_port,
     make_project_temp_dir,
     mount_docker_socket_into_runner,
+    reap_stale_named_containers,
     render_squid_config,
     resolve_env_endpoints,
     resolve_host_port,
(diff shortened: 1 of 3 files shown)


CiroGamboa:
Already merged upstream (commit 1803bbf6 is on main). Closing as stale.

PR #1215, opened 2026-08-16
fix(tbench): pin claude-code to 2.1.233 (close matrix Δ vs harbor)

CiroGamboa:
## Summary

- Bump \`data/harnesses.yaml\` \`claude-code\` pin from \`2.1.231\` → \`2.1.233\` to match harbor 0.21's own claude-code install.
- The version drift was the root cause of a real Δ observed on opus-4.7 during n=3 head-to-head runs this morning: TF averaged 0.955 on \`fix-billing-holds\` while harbor averaged 0.800 (Δ -0.155, ~2.6σ). Bumping the pin closed the gap to +0.022 (billing) and 0.000 (segmentation).

## Why the pin, not just \`latest\`

Same reason the docstring already gives: reproducibility. Every trial-image tag encodes \`local-claude-code-<version>\`, so a floating \`latest\` would silently change what the matrix is measuring. Pinning to whatever harbor's pinned to keeps the two pipelines running the same agent code without giving up the tag stability.

## Verification

- Opus 4.7 × claude-code × 2 tasks × n=3 × both pipelines (12 cells) run with the mismatched pin: TF billing 0.955 (σ 0.031), harbor 0.800 (σ 0.098), TF seg 1.000 (σ 0), harbor 0.911 (σ 0.063).
- Same 12 cells re-run with matched 2.1.233: TF billing 0.878 (σ 0.079), harbor 0.900 (σ 0.094), TF seg 0.956 (σ 0.063), harbor 0.956 (σ 0.063). Δ collapses to +0.022 and 0.000, inside both pipelines' shared noise floor.
- Four earlier matched rows (fable-5, sonnet-5, gpt-5.5, gpt-5.6-sol) already agreed within 0.011 across pipelines, so the fix isn't retroactively wrong for them — this pin change is what makes the whole priority-table comparison consistent going forward.

## Test plan

- [x] Local Opus 4.7 n=3 matrix confirms Δ collapse
- [ ] CI green on this PR

### DIFF
--- a/external_adapters/tolokaforge-adapter-terminal-bench/src/tolokaforge_adapter_terminal_bench/data/harnesses.yaml
+++ b/external_adapters/tolokaforge-adapter-terminal-bench/src/tolokaforge_adapter_terminal_bench/data/harnesses.yaml
@@ -9,7 +9,7 @@ harnesses:
     # Pinned to the version the out-of-tree host we compared against resolved
     # when the matrix comparison was baselined. The pin closes the CLI-code
     # delta with that host while keeping reproducibility.
-    version: "2.1.231"
+    version: "2.1.233"
     argv_prefix: ["claude"]
     # `--verbose --output-format=stream-json` matches the reference vendor-CLI
     # invocation. Aligning the flag block aligns the CLI's internal reasoning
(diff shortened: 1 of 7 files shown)


CiroGamboa:

PR #1217, opened 2026-08-16
fix(tbench-tasks): fix-billing-holds — supervisord auto-restart for uvicorn

CiroGamboa:
## Summary

- Adds \`supervisor\` to the fix-billing-holds task image and supervises uvicorn with \`autorestart=true\`.
- Ships \`start-billing-service\` helper that calls \`supervisorctl restart uvicorn\` and waits for the health window to close.
- Rewrites task step 3 to point at the helper (still keeps a raw \`uvicorn ... &\` fallback for models that ignore helpers).

## Motivation

The task grades pytest against \`localhost:8000\`. If uvicorn is dead at grade time, ~8 tests error at fixture setup (\`HTTPConnection: connection refused\`) and the trial scores a stochastic 0.4-0.7 that has nothing to do with whether the model's fee-calc + migration fixes were correct.

Matrix v16 traced this exactly: GPT-5.6 Sol × Harbor × billing t0 scored 0.6 because the model relaunched uvicorn in the foreground on turn 35 (dropped the \`&\`). Uvicorn died with its parent shell at session end; the run-tests.sh wait-loop timed out and pytest fired against a dead service. Same failure mode on Sonnet 5 × billing on some runs. **Both pipelines hit this at the same rate** — it's task fragility, not pipeline behavior.

Task grades correctness, not shell discipline. Belongs in infrastructure.

## Verification

- \`docker build\` the environment locally, then \`docker exec\` in and \`kill -9\` uvicorn — supervisord respawns it within 3s.
- Rerun GPT-5.6 Sol × billing at n=3 (both pipelines) after this lands — expect the variance to tighten and the row-level mean to rise.

## Test plan

- [x] Dockerfile builds cleanly (verify locally before merge)
- [ ] Full CI green on this PR
- [ ] Live rerun of GPT-5.6 Sol × billing shows tighter variance and no \`connection refused\` errors in grade.yaml


CiroGamboa:

PR #1244, opened 2026-08-20
fix(tbench-adapter): make pass_threshold task-authored, default to 1.0

CiroGamboa:
## Summary

The terminal-bench adapter hardcoded `pass_threshold=0.5` in both grading config sites (`get_grading_config` and `to_task_description`). Under `grading_method="test_execution"` — where the score IS the fraction of a reference test suite that passed — a 0.5 threshold silently graded a half-broken trial as PASS.

**Surfaced by Arena's `task_design_oracle` five-dimension verdict** on GitHub Actions run 32337928139 in `toloka-partners/tolokaforge-tasks`: 6/10 Opus 4.7 trials on TBench balanced-10 passed only via the 0.5 threshold; micro pass@1 reported 1.00 vs 0.40 under all-tests-pass. Analyzer's recommendation: *"Make the threshold task-authored (or 1.0 for test_execution grading) and regrade."*

## Fix

Follows both parts of the recommendation. Task-authoring path matches how `[verifier].timeout_sec` is already surfaced from `task.toml`:

- `TerminalBenchTask` gains a `pass_threshold` field, default `1.0`.
- `task_parser` reads `verifier.get("pass_threshold", 1.0)` with float coercion (a plain `1` in `task.toml` parses as int in `tomllib` and would fail downstream Pydantic validation).
- Both adapter sites (`get_grading_config`, `to_task_description`) read `meta.pass_threshold` instead of hardcoding.

Task authors who want partial credit declare it explicitly:

```toml
[verifier]
pass_threshold = 0.75
```

Nothing declares that today across the shipped example tasks or the `echo-hello` fixture, so the default of `1.0` kicks in and every task now grades under all-tests-pass semantics — matching how the `frozen-mcp-core` and `tlk-mcp-core` adapters grade their reference test runs.

## Test plan

- [x] 3 new parser tests: default, task-authored override, int→float coercion
- [x] 4 new adapter tests: both `get_grading_config` and `to_task_description`, each covering default and override
- [x] 2 canonical snapshots regenerated (`grading_config.json`, `task_description.json` for `tbench_echo_hello`) — the diff is `pass_threshold: 0.5 → 1.0`
- [x] Full terminal-bench test suite: 163 passed, 1 warning
- [ ] Downstream: `toloka-partners/tolokaforge-tasks` eval branches that repin will start reporting all-tests-pass numbers on TBench — expected behaviour, matches other adapters

### DIFF
--- a/external_adapters/tolokaforge-adapter-terminal-bench/src/tolokaforge_adapter_terminal_bench/adapter.py
+++ b/external_adapters/tolokaforge-adapter-terminal-bench/src/tolokaforge_adapter_terminal_bench/adapter.py
@@ -366,11 +366,24 @@ class TerminalBenchAdapter(BaseAdapter):
     # -- grading config -------------------------------------------------------
 
     def get_grading_config(self, task_id: str) -> GradingConfig:
+        # Terminal-bench grades by running the reference test suite in the env
+        # container (``grading_method="test_execution"`` in the runner-side
+        # config below), so ``pass_threshold`` is the fraction of the suite
+        # that must pass for the trial to grade PASS. Task-authored via
+        # ``[verifier].pass_threshold`` in task.toml, defaulting to 1.0 —
+        # the natural semantics of a test suite is binary "all tests pass".
+        # A silent low default inflates pass rates on partial credit (surfaced
+        # by Arena's task_design_oracle verdict on GH Actions run 32337928139
+        # in toloka-partners/tolokaforge-tasks: 6/10 Opus-4.7 trials passed
+        # only via the previous hardcoded 0.5 threshold; micro pass@1 reported
+        # 1.00 vs 0.40 under all-tests-pass).
+        self._ensure_discovered()
+        meta = self._tasks[task_id]
         return GradingConfig(
             combine=GradingCombineConfig(
                 method="weighted",
                 weights={"custom_checks": 1.0},
-                pass_threshold=0.5,
+                pass_threshold=meta.pass_threshold,
             ),
         )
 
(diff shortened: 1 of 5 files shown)


CiroGamboa:
Closing this PR. The adapter fix is in a codepath we shouldn't be running for TBench-in-Arena at all.

The Slack thread that set up TBench-in-Arena (`C0B42KY5GM6/1785527098.558109`, 2026-07-31 → 2026-08-19) agreed on **Terminus-2 as the universal agent**. Terminus-2 is Harbor's scaffold and grades via Harbor's own logic — same as Diego's baseline, no threshold semantic to argue about.

Our Arena pilot used `agent_harness: engine-loop` (tolokaforge's own turn loop) via this adapter because the standard `arena-eval` skill dispatches `tolokaforge run` and tolokaforge has no Terminus-2 harness. That means the adapter's grading — including this `pass_threshold` — ran on trials it shouldn't have graded. The fidelity finding from Arena's `task_design_oracle` verdict is real, but the right fix is to stop running the adapter's grader for TBench, not to change its threshold.

The pivot: Balázs's proposed Path B (see Slack). `harbor-runner --agent terminus-2` drives the trial (Harbor grades, matching Diego), a results-only adapter reshapes the bundle for `benchmark-results-collector`, no touch on this adapter's grading path.

Filing a follow-up so the task-authored `pass_threshold` idea isn't lost: even under Path B, hardcoded config in an external adapter is a smell, and the field + parser plumbing added here is the right shape. If someone wants to revive it later, this diff is the reference.

Issue #1234, opened 2026-08-18
fix(testing): harness-registry replay snapshot is self-referential and lands stale


CiroGamboa:
## Context

`tests/canonical/test_harness_registry_replay.py` replays every HEAD-reachable
commit that touched a path in
`tests/canonical/_harness_registry_replay/git_walk.py:HARNESS_SURFACE_PATHS`,
classifies each as Bucket A (data-only) or Bucket B (code), and pins the result
to `tests/canonical/snapshots/harness_registry_replay/metric.json`.

Because the metric is derived from git history reachable from HEAD, **any commit
that touches a harness-surface path invalidates the snapshot** — including the
commit that regenerates it. `--update-canon` runs before the commit exists, so
the regenerated snapshot cannot contain its own commit. The branch lands green
locally and red for the next person.

This has already happened. On `main` today:

    $ uv run pytest tests/canonical/test_harness_registry_replay.py -q
    E   AssertionError: Mismatch with golden .../harness_registry_replay/metric.json
    E     {'bucket_b_count': 8} != {'bucket_b_count': 7}
    [harness-registry replay]  Bucket A: 1  |  Bucket B: 8  |  total: 9

Commit 33f9916c ("docs: strip internal-repo references from public artifacts")
regenerated `metric.json` in the same commit that touched
`.../data/harnesses.yaml`, so the snapshot omits that very commit.

There is a convergent sequence — `metric.json` is itself not on
`HARNESS_SURFACE_PATHS`, so a commit touching *only* the snapshot converges —
but nothing in the repo says so, and the natural workflow (edit, regenerate,
commit) is the non-convergent one.

## What to do

Pick one:

1. **Make the replay exclude HEAD itself**, or exclude commits whose only
   harness-surface contribution is the snapshot regeneration. Removes the footgun
   at the cost of the metric lagging one commit.
2. **Assert a bound rather than an exact snapshot.** The test's stated purpose is
   measuring the Bucket-A fraction over time; an exact-match snapshot of the
   whole commit list is a stricter lock than the metric needs. A monotonic
   assertion (Bucket-A fraction never regresses) survives any new commit.
3. **Document and enforce the convergent sequence** — regeneration lands in its
   own commit touching only `metric.json` — with a CI check that fails when one
   commit both touches a harness-surface path and edits `metric.json`.

Option 2 is the most faithful to what the test is for; option 1 is the smallest
change.

## Verification

[...]

PR #1245, opened 2026-08-21
feat(grading): trace-checks v2 operators — dates, negative text, nullness + full-path validation

### LINKED ISSUE #765: feat(grading): validate nested trace-check argument paths past the first segment
## Context

#679 validates a trace-check matcher's `args` paths against the referenced tool's JSON schema, but
only the **first** dotted segment. `args: { json.q: { len_gt: 0 } }` on `http_request` checks that
`json` is a declared property and stops.

That limit is forced for `http_request` specifically: its `json` property is declared
`{"type": "object", "description": "Optional JSON body for POST/PUT"}` with no nested `properties`,
so the schema cannot answer whether `q` is a valid key — and it never will, because the body shape
is the task's, not the tool's.

It is **not** forced in general. A task-pack tool whose parameter is a nested Pydantic model emits
a schema with `properties` under the object, and a second-segment typo there is as checkable as a
first-segment one. #679 stops at the first segment uniformly rather than descending as far as each
schema allows, so those typos still reach grade time as a silently-unmatchable matcher.

Filed so the limit has an owner: #679's plan lists it in `docs/GRADING.md` § Trace Checks under
"Declared limits, and what owns each", where every row names the issue that closes it.

## Deliverables

- [ ] Argument-path validation descends a dotted path as far as the schema declares nested
      `properties`, and stops at the first segment whose schema declares none
- [ ] The severity rule from #679 applies per level: a segment missing from a level whose schema is
      `additionalProperties: false` is an error; missing from an open level is advisory
- [ ] The docs state how deep validation reaches, replacing #679's flat "first segment only"

## Acceptance Criteria

- A task-pack tool with a nested model parameter rejects a second-segment typo at
  `tolokaforge validate`
- `args: { json.q: … }` on `http_request` still produces no finding — the segment below an
  object with no declared `properties` stays unchecked, not rejected
- The example corpus still validates clean

## Out of Scope

- Inferring a body shape the schema does not declare (`http_request.json` and friends stay
  unchecked below the first segment — there is nothing to check against)
- Argument correlation across matchers (#681)

## Dependencies

- Requires #679 (the argument-path validator and its severity rule)

kostiantynvoiku-web:
## Summary

Delivers three of the four features from Valentin's schema feature request ([#tolokaforge-development thread](https://toloka-ai.slack.com/archives/C0AELU19H0S/p1786720252901289), spec: `trace_migration/TRACE_CHECKS_V2_SPEC.md` @ `73eb4c227` on tolokaforge-tasks `dev/ots`), plus the #765 loader walk. After ~1,500 migrated testcases, these were the operators blocking further rubric→trace-check migration; together they unblock roughly 37% of the engine-askable blocked rubric weight. Part of epic #684.

The fourth ask (`binds:` multi-binder) is deliberately deferred — it is the one architectural change to the evaluator and waits for its owner to return.

## The operators

- **`date_gt` / `date_gte` / `date_lt` / `date_lte`** — chronological comparison of ISO-8601 dates and datetimes under one normalization policy (`predicates.date_comparison_key`): a date-only string is midnight UTC, a naive datetime reads as UTC, a `Z`/offset converts to UTC. The accepted grammar is shape-gated rather than left to `fromisoformat`, so one config parses identically on every interpreter. A malformed **authored bound** is a load error (the predicate would hold for no trial); an unparseable **trial value** is false, never an error.
- **`not_contains` / `not_regex`** — negate the match, never the evidence: they sit behind the same None-guard as every operator, so an absent or `null` field satisfies neither and a call whose outcome was never recorded cannot pass "the result does not mention X". `not_regex` joins the compile-time regex walk.
- **`is_null` / `omitted`** — subdivide the two conditions `exists: false` reads as one, via a found/missing sentinel in `_argument_at`. `exists` keeps its exact prior meaning, every existing operator keeps byte-identical behavior on both states, a missing intermediate key on a dotted path reads as omitted, and a nullness probe on record-only fields (`status`/`executor`/`result`) is a load error — a missing value there is missing *evidence*, held undecidable.

## Loader full-path validation (#765)

The authoring gate now validates an `args` path as deep as the tool schema keeps declaring `properties`, and stops conservatively where it stops declaring them (an untyped blob, an array, a permissive object) — everything below is unchecked rather than wrong, so packs that lawfully grade on undeclared keys (e.g. `json.q`) keep loading. A typo'd tail under a declared level is now a load error. Closes #765 (partially — the declared-limits row in `docs/GRADING.md` records what remains).

## Compatibility

- Closed schema preserved: the eight operators are additive `ValuePredicate` fields defaulting `None`, so every shipped `grading.yaml` evaluates identically and declaredness survives the gRPC round trip.
- Packs that **use** the new operators need a runner image built from this engine — rebuild the runner image before rolling the engine.
- No `ENGINE_PROTOCOL_VERSION` bump expected — please confirm.

## Tests

- The spec's normative three-state matrix (omitted / explicit null / valued) parametrized per operator; date normalization (naive-as-UTC, date-only-as-midnight, mixed-shape compare, numeric ops still refuse date strings); negation edges (co-occurrence with the positive form, absent/null satisfies nothing, empty string is a value); missing-intermediate-key-reads-as-omitted.
- Load errors: malformed date bounds, nullness probes on record-only fields.
- Authoring gate: deep-path typo refused, untyped level tolerated with the pinned skip reason; the three shipped nested-argument packs keep their documented skip.
- Canon: the wire round-trip config now spans all eight operators.
- Three new substrate-parity packs (`tests/data/tasks/trace_checks_operator_{dates,negation,nullness}`), each driven end to end through the core engine and the runner's real gRPC `GradeTrial`, asserting the identical 1.0/0.0 split on both substrates.


kostiantynvoiku-web:
**CI status analysis** (test-smoke: 13 failed / 9,077 passed)

All 13 failures are pre-existing on `main` — its scheduled CI run ([32455367488](https://github.com/Toloka/tolokaforge/actions/runs/32455367488), 2026-08-21) fails on the same 13 plus one more, and #1242 merged over the same set:

- `test_harness_registry_replay` / `test_models_wheel_replay` — replay baselines invalidated by squash-merge history rewrite; the failure message itself says to regenerate via `update_canonical_snapshots` (tracked in #1234).
- `test_tolokaforge_models_bootstrap` — models-wheel contract pins `1.0.0` while the released wheel is `1.2.0` (drift from the models 1.2.0 release).
- `test_config_validator::test_and_one_the_overlay_does_not_declare_still_is` — env-dependent (missing OPENROUTER key changes the validator output).
- 9 × `test_run_display` keyboard-dispatch tests — display code this PR does not touch.

One main failure this PR **fixes**: `test_the_rejection_table_names_every_validator_that_raises` — main’s rejection table misses `_reject_a_status_literal_no_execution_produces`; the table updates in this PR cover it.

No failure is introduced by this PR; every grading / trace-checks / corpus / parity test passes.

tolokaforge/runner/models.py
```diff
@@ -750,6 +809,29 @@ class TraceMatcher(BaseModel):
...
 
+    @model_validator(mode="after")
+    def _reject_a_nullness_probe_on_recorded_evidence(self) -> TraceMatcher:
+        """``is_null`` / ``omitted`` subdivide a JSON value's absence — argument
+        material. On ``status``, ``executor`` and ``result`` a ``None`` is missing
+        *evidence*, which the evaluator holds undecidable rather than reads, so a
+        nullness probe there would answer a question the record cannot ask.
+        """
+        offending: dict[str, list[str]] = {}
+        for field_name in ("status", "executor", "result"):
```
CiroGamboa, line 820: **Nullness-probe rejection misses `text` and `tool` matchers**

`_reject_a_nullness_probe_on_recorded_evidence` (this validator) iterates `("status", "executor", "result")` but not `text` (on `assistant_message`) or `tool` (on `tool_call`). `is_null` / `omitted` on those two fields are semantically dead — the matcher selects nothing on every event, and the default `on_missing` reports it as the agent's failure. That's the exact hazard the sibling validator was built to prevent.

**Failure scenario:** an author writes `text: {is_null: true}` on an `assistant_message` matcher (or `tool: {omitted: true}` on a `tool_call` matcher).
- `event.text` is always a real string → `is_null` returns False on every event.
- `event.tool_name` is always set → `is_null` returns False on every event.
- `omitted` is even worse: the `_MISSING` sentinel is only ever produced by `_argument_at` for `args` paths, so `omitted` on any non-`args` field is False on every event.

Result: the task rubric ships a silent always-false predicate; the failure surfaces as an agent failure, not as a config-time diagnostic.

**Fix:** extend this validator (or add a peer) to refuse `is_null`/`omitted` on `text` and `tool` too — the same construction the `status`/`executor`/`result` case already uses.

tolokaforge/core/grading/trace_checks.py
```diff
@@ -505,16 +545,24 @@ def _matches_regex(value: Any, pattern: str) -> bool:
...
+    "date_gt": _dateish(operator.gt),
+    "date_gte": _dateish(operator.ge),
+    "date_lt": _dateish(operator.lt),
+    "date_lte": _dateish(operator.le),
     "in_": lambda value, allowed: value in allowed,
     "not_in": lambda value, rejected: value not in rejected,
     "len_gt": _by_length(operator.gt),
     "len_gte": _by_length(operator.ge),
     "exists": lambda value, expected: (value is not None) is expected,
+    "is_null": lambda value, expected: (value is None) is expected,
```
CiroGamboa, line 564: **`is_null` / `omitted` `_OPERATORS` entries are unreachable dead code that would silently give wrong answers if the early-return branches ever go**

`_operator_holds` (lines 482–485 above) handles `is_null` and `omitted` before dispatch, then coerces `value is _MISSING` → `value = None` (lines 486–487). By the time control reaches the dispatch table these two `_OPERATORS` lambdas can never fire.

**Failure scenario:** a maintainer refactors the early returns away ("dispatch is already in the table"). `_OPERATORS['omitted']` lambda now executes on the coerced value:
```python
"omitted": lambda value, expected: (value is _MISSING) is expected
```
The coercion has already destroyed the sentinel, so `omitted: true` returns False on every input. Every `omitted:` predicate silently fails — both lambdas type-check and pass their unit checks in isolation, so nothing catches it in review.

**Fix (pick one):**
- Delete the two lambdas from `_OPERATORS` — they document a contract they can't fulfil.
- Or delete the early-return branches from `_operator_holds` and route dispatch through the table using the RAW (pre-coercion) value, so the lambdas' `value is _MISSING` check actually works.

tolokaforge/core/grading/trace_checks.py
```diff
@@ -462,6 +479,12 @@ def _operator_holds(name: str, value: Any, expected: Any, bindings: Mapping[str,
     written out. The name is in the environment by construction — a reference to a
     name the constraint does not bind is a load error.
     """
+    if name == "is_null":
+        return (value is None) is expected
+    if name == "omitted":
+        return (value is _MISSING) is expected
+    if value is _MISSING:
```
CiroGamboa, line 486: **`is_null` / `omitted` / `_MISSING → None` ordering is load-bearing and one-line-fragile**

The three gates in `_operator_holds` sit in this exact order:

1. `is_null` returns on `value is None` (line 483 above).
2. `omitted` returns on `value is _MISSING` (line 485).
3. `value is _MISSING` → coerce to `None` (line 486, the diff line here).

Ordering is load-bearing but unmarked: no comment names the invariant, no test probes the ordering. Same failure class as the `_OPERATORS` lambda concern — reached via a different path.

**Failure scenario:** a refactor moves gate (3) above gate (2). `omitted: true` never fires — the sentinel is coerced to `None` before the check runs — and `omitted` silently conflates with `is_null`.

**Fix:** either
- one inline comment naming the ordering constraint ("early-return `is_null`/`omitted` must run BEFORE `_MISSING` is coerced to `None`; `_OPERATORS` gets the coerced value only"),
- or lift the sentinel-aware dispatch into a small helper whose signature makes the pre/post-coercion distinction explicit at the type level.

tolokaforge/core/grading/predicates.py
```diff
@@ -109,6 +111,47 @@ def ever_satisfiable(operator: str, held: str | None, bound: str | None) -> bool
...
+    shape = _ISO_DATE_SHAPE.match(value)
+    if shape is None:
+        return None
+    text = value[:-1] + "+00:00" if value.endswith("Z") else value
+    fraction = shape.group("fraction")
+    if fraction is not None and len(fraction) not in (3, 6):
+        # Python 3.10's ``fromisoformat`` reads exactly 3 or 6 fractional digits;
+        # later versions read any count. Padding to 6 keeps one parse per config
+        # on every interpreter this package supports.
+        text = text.replace(f".{fraction}", f".{fraction.ljust(6, '0')}", 1)
```
CiroGamboa, line 145: **Fractional-second padding rides on a single-dot invariant this line doesn't own**

`text.replace(f".{fraction}", f".{fraction.ljust(6, '0')}", 1)` is safe only because `_ISO_DATE_SHAPE` guarantees exactly one `.` in `text` — the fractional-second dot. The moment the regex accepts a shape that puts a `.` elsewhere, `str.replace` with `count=1` silently targets whichever `.<fraction>` match comes first.

**Failure scenario:** the shape is later widened (locale prefix, timezone-in-seconds decimals, a leading duration marker with a decimal, or any future ISO 8601 relaxation). The first `.<fraction>` match may now land anywhere in the string, and the replacement rewrites the wrong span — the padding drifts off its intended anchor, silently normalizing the wrong digits.

**Fix:** slice-based build off the regex's own positions:

```python
m = _ISO_DATE_SHAPE.match(text)
if m and (fraction := m.group("fraction")):
    padded = fraction.ljust(6, "0")
    text = text[:m.start("fraction")] + padded + text[m.end("fraction"):]
```

That way the padding cannot slide off its regex-anchored position, regardless of what the shape accepts later.

PR #1277, opened 2026-08-25
integrate: z-ai/glm-5.3

kostiantynvoiku-web:
# Integrate `z-ai/glm-5.3` (OpenRouter) via auto-resolve

**Candidate:** provider `openrouter` · name `z-ai/glm-5.3` · model_id `openrouter__z-ai_glm-5.3`
**Engine ref:** `4e67c4a97ef36441e627c12b3a153e033d5ddceb`
**Observed:** 2026-08-25 — 31 capability probes × 15 reps, 6 shape variants × 15 reps, 199 wire
trials / 688 tool calls. Fix loop CONVERGED in 1 iteration.

## The policy

One new model-specific preset, `z_ai_glm_5_3` in `model_presets.yaml`, declared before the shared
`openrouter_dict_stringify_recovery` preset whose `z-ai/glm-5*` glob already claimed this route (so
the baseline was measured on those shared axes). Because overlay entries shadow rather than merge,
the entry is a **superset**: `passthrough` / `json_coerce` / `dict_map_hints` are re-declared
verbatim, and exactly one axis changes —

```
reasoning_codec: openai → openai_summary_replay
```

`OpenAISummaryReplayReasoningCodec` already ships (`tolokaforge_models/policies/deepseek.py`); **no
new adapter class**, and no engine-side change under `tolokaforge/core/llm/`. It inherits the OpenAI
`extract` verbatim — `THINKING_EMITS_BLOCKS` keeps its 15/15 summary surface — and adds only the
unsigned OpenRouter `reasoning_details` rebuild.

The globs are vendor-anchored (`z-ai/glm-5.3`, `*/z-ai/glm-5.3` — exact shapes, no trailing `*`) so no sibling
route drifts: `glm-5`, `glm-5-turbo`, `glm-5.1`, `glm-5.2` and `glm-5v-turbo` were all re-resolved
post-edit and still land on the shared preset with `reasoning_codec: openai`.

**Effect:** `test_unsigned_thinking_replay` 0/15 → 5/5. The other 25 tool-call / shape probes were
already 15/15, all 6 shape variants 15/15, and 0 tool-arg rejections over 688 tool calls — all
unchanged.

- **`PROMPT_CACHING`** — 0 `cache_creation_input_tokens` on an 8.2 k-token first call, 15/15
  identical. No Anthropic-style ephemeral cache exists on this z-ai route, so no `cache_policy` can
  manufacture one. Same posture as the glm-5.1 / glm-5.2 siblings. `IMPLICIT_PROMPT_CACHING` passes
  15/15 and stays required.
- **`THINKING_REPLAY_ROUNDTRIP`** — "no signed blocks on turn 1", 15/15. The route emits no
  per-block signature and no policy may synthesise one; `openai_summary_replay` deliberately does not
  flip this.

Cert: 21 required / 2 known_unsupported. `UNSIGNED_THINKING_REPLAY` is `required` here (the siblings
declare it unsupported) — it is a payload-level contract, verified by reprobe.

Pricing verified live: `"z-ai/glm-5.3": {"input": 1.4, "output": 4.4, "cache_read": 0.26}` USD/1M.

## Files

- `tolokaforge_models/src/tolokaforge_models/data/model_presets.yaml` — new `z_ai_glm_5_3` entry
- `tolokaforge_models/src/tolokaforge_models/certificates/registry.py` — new `MC(...)` in `_ALL`
- `tolokaforge_models/src/tolokaforge_models/data/pricing.json` — `z-ai/glm-5.3` pricing
- `tests/unit/llm/test_z_ai_glm_5_3_preset.py` — routing invariants (new)
- `tolokaforge_models/tests/unit/fixtures/openrouter_z_ai_glm_5_3_reasoning_response.json` — live z-ai wire envelope (new)
- `tolokaforge_models/tests/unit/test_reasoning_codec_openai_summary_replay.py` — z-ai fixture round-trip + route→codec pin
- `tests/canonical/snapshots/certify_suite_collection/collection.json` — +37 nodeids for the new cert

---

Integrated automatically via **auto-resolve** (observe → fix → reprobe → finalize); the policy was
written by the automation, the review follow-ups below by hand. The branch was opened as a
disposable test branch; it is now intended to merge (#1278, the effort-rule half, is stacked on it).

## Review follow-ups (CiroGamboa, 2026-08-25)

1. **Canonical certify-suite snapshot** regenerated with `--update-canon`: +37 nodeids for
   `openrouter__z-ai_glm-5.3`, nothing else moved. `test-smoke` on this branch had exactly one
   failure beyond `main`'s 14 pre-existing ones; this was it.
2. **Routing pinned** — `tests/unit/llm/test_z_ai_glm_5_3_preset.py`: bare and gateway-prefixed
   `z-ai/glm-5.3` resolve to `z_ai_glm_5_3`; unmeasured variants (`-turbo`, `-fast`, `5.30`,
   `5.3.1`, `:exacto`) do not; 5.1 / 5.2 / 5 / 5-turbo / 5v-turbo stay on
   `openrouter_dict_stringify_recovery` with the plain `openai` codec; the overlay equals the
   shared stack on every axis but the codec.
3. **Globs narrowed** to `z-ai/glm-5.3` and `*/z-ai/glm-5.3` — no trailing `*`, so a future point
   release falls through to the shared preset instead of inheriting an unmeasured codec.
4. **z-ai wire fixture** captured live 2026-08-25 (`openrouter_z_ai_glm_5_3_reasoning_response.json`,
   287 reasoning tokens) and pinned: `extract` yields one non-empty `summary_text` block equal to
   the flat summary, and `encode_for_replay` rebuilds exactly the route's own unsigned
   `reasoning.text` envelope (type + text; no signature, no format).
5. **"SUPERSET" wording** replaced in `registry.py` (and the YAML) with the load-bearing version:
   the overlay is a FULL COPY plus a codec swap; `_match_preset` has no fallback merge, so the
   re-declared fields cannot be pruned as duplicates.
6. **Sibling pin** — 5.1 / 5.2 locked to the shared preset (item 2).
7. **`_meta.updated_at` bump dropped** — the field belongs to `pricing-updater`; `pricing.json`
   now differs from `main` by the `z-ai/glm-5.3` entry only.


kostiantynvoiku-web:
Heads-up for the reviewer: **#1278 is stacked on this branch** and edits the same `z_ai_glm_5_3` preset entry this run created — it adds a `param_value_rules` effort rule (`low`/`medium` → drop to the provider default) plus a new `REASONING_EFFORT_HONOURED` capability probe that reproduces why: under an evaluation-shaped context Z.AI degrades `medium` to 0 reasoning tokens (the Arena v1 eval ran 704 trials with 0 reasoning tokens), which the short-prompt probes here cannot see (glm-5.3 thinks at `medium` on a one-liner). Merge this PR first; #1278 retargets to `main` automatically. Note this branch's CI did not run (pushes by `github-actions[bot]` do not trigger workflows); #1278's `test-smoke` shows the same 12 pre-existing failures as `main`'s push run 32837851793.

tolokaforge_models/src/tolokaforge_models/certificates/registry.py
```diff
@@ -2122,6 +2122,98 @@
...
+    # ("no signed blocks on turn 1", 15/15 → THINKING_REPLAY_ROUNDTRIP; the
+    # replay codec synthesises no signature), and no Anthropic-style ephemeral
+    # cache is wired on this z-ai route (0 cache_creation_input_tokens on an
+    # 8.2 k-token first call, 15/15 → PROMPT_CACHING; ``anthropic_ephemeral``
+    # would only inject cache_control markers the route ignores). Same posture as
+    # the glm-5.1 / glm-5.2 siblings. IMPLICIT_PROMPT_CACHING passed the observe
+    # baseline 15/15 and is ``required``.
+    # -----------------------------------------------------------------
+    MC(
+        model_id="openrouter__z-ai_glm-5.3",
```
CiroGamboa, line 2172: **Canonical certify-suite collection snapshot is not updated for this cert**

Merging this PR fails `tests/canonical/test_certify_suite_collection.py::test_certify_suite_collection_snapshot` immediately: the new `openrouter__z-ai_glm-5.3` cert emits ~37 new parametrised nodeids (same shape as the 111 existing `deepseek-v4-flash-0731` rows), and `tests/canonical/snapshots/certify_suite_collection/collection.json` still has zero refs to `z-ai_glm-5.3`.

**Failure scenario:** `test_certify_suite_collection_snapshot` runs `pytest --collect-only` over `tolokaforge.testing.certify.suite` and pins the result byte-for-byte against `collection.json`. The new cert's parametrised suite adds rows for every declared capability × repeat; the pinned snapshot doesn't have them; CI fails at the snapshot diff.

**Fix:** regenerate `collection.json` with `--update-canon` (or whatever the repo's canon-refresh flag is) and commit the delta as part of this PR.

Note: the PR body itself flags the branch as "disposable — must NOT be merged", so this is downstream of that decision. If disposable, still worth teaching the auto-integration flow to run `--update-canon` before opening the PR so future runs land in a mergeable state.

tolokaforge_models/src/tolokaforge_models/data/model_presets.yaml
```diff
@@ -305,6 +305,60 @@ presets:
...
+  # It synthesises no signature, so signed ``THINKING_REPLAY_ROUNDTRIP`` ("no
+  # signed blocks on turn 1", 15/15) stays a declared ceiling — the honest half
+  # of the pair this codec deliberately does NOT flip. ``PROMPT_CACHING`` (0
+  # cache_creation_input_tokens on an 8.2 k-token first call, 15/15) is the other
+  # ceiling: no Anthropic-style ephemeral cache exists on this z-ai route, so no
+  # ``cache_policy`` can manufacture one (``anthropic_ephemeral`` would only
+  # inject cache_control markers the route ignores) — same posture as the
+  # glm-5.1 / glm-5.2 siblings. ``IMPLICIT_PROMPT_CACHING`` passes 15/15 and
+  # stays required. See registry.py.
+  z_ai_glm_5_3:
```
CiroGamboa, line 350: **No test pins first-match-wins routing of `z-ai/glm-5.3` to this preset**

`z_ai_glm_5_3` is declared BEFORE `openrouter_dict_stringify_recovery` (whose `z-ai/glm-5*` glob would otherwise claim this route). That order is load-bearing: it's the whole reason the model gets `openai_summary_replay` instead of `openai`. But there's no unit test that pins the resolution — any future reorder that moves the shared preset above `z_ai_glm_5_3`, or broadens its glob, silently reverts the codec and reintroduces the 0/15 `UNSIGNED_THINKING_REPLAY` failure.

**Failure scenario:** a maintainer sorts `presets:` alphabetically for readability, or a later refactor moves the shared preset up to consolidate the OpenRouter overlays. `_match_preset` returns first match → `z-ai/glm-5.3` gets `openai` codec → thinking-replay reprobe silently regresses. Only live CI with `OPENROUTER_API_KEY` catches it; offline suites stay green.

**Fix:** add a sibling of `tolokaforge_models/tests/unit/test_reasoning_codec_openai_summary_replay.py::test_the_0731_route_resolves_to_this_codec` that asserts `_match_preset("z-ai/glm-5.3")` and `_match_preset("openrouter/z-ai/glm-5.3")` both return the `z_ai_glm_5_3` preset. That test is trivial and locks the ordering invariant.

tolokaforge_models/src/tolokaforge_models/data/model_presets.yaml
```diff
@@ -305,6 +305,60 @@ presets:
...
+    match:
+      # Exact vendor-anchored globs only. A bare "*glm-5.3*" would also claim
+      # foreign routes whose name merely contains it; the gateway-prefixed forms
+      # (openrouter/…, litellm_proxy/…) reach the same model through a transport
+      # prefix, so the vendor segment is anchored rather than matched anywhere
+      # in the string.
+      - "z-ai/glm-5.3"
+      - "z-ai/glm-5.3*"
```
CiroGamboa: **`z-ai/glm-5.3*` glob claims unmeasured future releases**

`fnmatch`'s `*` matches ANY chars including `.` and `-`, so `z-ai/glm-5.3*` and `*/z-ai/glm-5.3*` will silently claim a future `z-ai/glm-5.30`, `z-ai/glm-5.3.1`, `z-ai/glm-5.3-fast`, `z-ai/glm-5.3-beta`, etc. — each of which inherits the `openai_summary_replay` codec + `json_coerce` recovery without ever being measured.

**Failure scenario:** OpenRouter ships `z-ai/glm-5.3-fast` (variant with a different reasoning envelope) or `z-ai/glm-5.3.1` (bugfix release). Both match this glob; both inherit the codec the 5.3 baseline was measured under. If the fast variant emits nested `reasoning_details` where the codec expects flat `reasoning`/`reasoning_content`, `OpenAISummaryReplayReasoningCodec.extract` returns `None`, `encode_for_replay` emits `{}`, and every production request silently loses thinking replay.

Sibling `deepseek_v4_flash_0731_resolve` avoids this by anchoring on a date tag (`-0731*`) that resists numeric extension. This route doesn't have that anchor.

**Fix:** narrow to the exact three shapes you actually measured — drop the trailing `*`. If a new point release needs the same overlay, add it explicitly in a PR whose title carries the reprobe evidence. That mirrors the discipline the rest of the certificate registry follows.

tolokaforge_models/src/tolokaforge_models/certificates/registry.py
```diff
@@ -2122,6 +2122,98 @@
...
+                C.DISCRIMINATED_UNION_TOOL_CALL,
+                C.DECIMAL_FIELD_TOOL_CALL,
+                C.RECURSIVE_REF_TOOL_CALL,
+                C.HETEROGENEOUS_ARRAY_TOOL_CALL,
+                C.ALLOF_MERGE_TOOL_CALL,
+                C.THINKING_EMITS_BLOCKS,
+                # Fixed by the ``openai_summary_replay`` codec (re-emits the
+                # turn-1 summary as an unsigned OpenRouter reasoning.text detail
+                # on replay); observe was 0/15, reprobe 5/5.
+                C.UNSIGNED_THINKING_REPLAY,
```
CiroGamboa, line 2194: **No z-ai wire-envelope fixture ships — the codec choice is un-regression-testable offline**

This cert flips `UNSIGNED_THINKING_REPLAY` to required for the first time in the z-ai family, but only `openrouter_deepseek_v4_reasoning_response.json` exists in `tolokaforge_models/tests/unit/fixtures/`. There's no z-ai analogue, so `OpenAISummaryReplayReasoningCodec.extract` / `encode_for_replay` cannot be regression-tested against a captured z-ai wire envelope.

**Failure scenario:** OpenRouter's z-ai route later drops the flat `reasoning` / `reasoning_content` fields and emits only nested `reasoning_details` (or a differently-keyed envelope). `extract` returns `None`, `encode_for_replay` emits `{}`, and the "5/5 reprobe" claim silently regresses — but every offline test still passes because none of them exercises the z-ai wire shape. Only live CI with `OPENROUTER_API_KEY` catches the drift, and only until the model is deprecated.

**Fix:** capture one representative z-ai OpenRouter response into `tolokaforge_models/tests/unit/fixtures/openrouter_z_ai_glm_5_3_reasoning_response.json` and add a sibling of the deepseek round-trip test that pins:
1. `extract(envelope)` returns a non-empty `ThinkingBlock`
2. `encode_for_replay(...)` round-trips to the wire shape the route expects

tolokaforge_models/src/tolokaforge_models/certificates/registry.py
```diff
@@ -2122,6 +2122,98 @@
             }
         ),
     ),
+    # -----------------------------------------------------------------
+    # GLM-5.3 (Zhipu / Z.AI via OpenRouter) — landed via auto-resolve (PR
+    # #1277). Routes through the model-specific ``z_ai_glm_5_3`` preset (see
+    # model_presets.yaml), declared BEFORE the shared
+    # ``openrouter_dict_stringify_recovery`` preset whose ``z-ai/glm-5*`` glob
+    # already claimed this route. That overlay is a SUPERSET of the shared
```
CiroGamboa: **"SUPERSET" wording invites a future editor to silently delete dict-stringify recovery**

Preset resolution is match-first-wins — `_match_preset` picks up the first matching entry with **no fallback merge** to the shared preset. Calling the overlay a "SUPERSET" of the shared preset reads as "the shared preset's fields are inherited", which is exactly the opposite of what resolution does. A future cleanup pass that treats the `passthrough` / `json_coerce` / `dict_map_hints` lines on `z_ai_glm_5_3` as "duplicated with the shared preset" will remove them, and z-ai/glm-5.3 quietly loses discriminated-union dict-stringify recovery under production load.

The YAML preset already warns explicitly (`model_presets.yaml:316`: *"dropping them here would silently regress the dict-stringify recovery the whole baseline was measured under"*), but this comment on the registry side undoes that warning by calling the same fields inherited.

**Fix:** replace the "SUPERSET" wording here with the load-bearing version, e.g.:

> The overlay is a FULL COPY of `openrouter_dict_stringify_recovery` PLUS a `reasoning_codec` swap. The copy is load-bearing — `_match_preset` returns the first match with no fallback merge — so the `passthrough` / `json_coerce` / `dict_map_hints` fields cannot be pruned as duplicates. See `model_presets.yaml:316`.

tolokaforge_models/src/tolokaforge_models/data/model_presets.yaml
CiroGamboa, line 350: **No test pins `z-ai/glm-5.1` / `z-ai/glm-5.2` to the shared preset**

The z-ai 5.1 and 5.2 siblings currently route through `openrouter_dict_stringify_recovery` — that's the invariant the paragraph above (`# same posture as the glm-5.1 / glm-5.2 siblings`) leans on. But no test locks them there. A future broadening of `z_ai_glm_5_3.match` — say, to `z-ai/glm-5*` when 5.4 lands and wants the same overlay — silently poaches 5.1 and 5.2 onto the summary-replay codec whose 5.1/5.2 wire shapes were never measured.

**Failure scenario:** onboarding of `z-ai/glm-5.4` in a later PR relaxes `z_ai_glm_5_3.match` to a `glm-5*` glob for convenience. First-match-wins then routes 5.1 and 5.2 to `openai_summary_replay` too. Every offline suite stays green; live traffic on 5.1/5.2 quietly loses thinking replay because their reasoning envelopes are not what this codec expects.

**Fix:** add a sibling of `test_reasoning_codec_openai_summary_replay.py::test_siblings_keep_the_shared_preset` that asserts `_match_preset("z-ai/glm-5.1")` and `_match_preset("z-ai/glm-5.2")` both resolve to `openrouter_dict_stringify_recovery`. Analogous to how deepseek pins `deepseek-v4-flash` / `deepseek-v4-pro` there.

tolokaforge_models/src/tolokaforge_models/data/pricing.json
```diff
@@ -2,7 +2,7 @@
   "_meta": {
     "description": "Model pricing in USD per 1M tokens. Source: OpenRouter API.",
     "source_url": "https://openrouter.ai/api/v1/models",
-    "updated_at": "2026-08-14T10:17:21.788923+00:00",
+    "updated_at": "2026-08-25T11:40:29.130510+00:00",
```
CiroGamboa: **Manual `_meta.updated_at` bump gets overwritten by the pricing-updater**

`tools/pricing-updater/src/pricing_updater/fetcher.py:228` writes `updated_at: datetime.now(tz=timezone.utc).isoformat()` unconditionally on every `pricing-updater update` run. The manual `2026-08-25T11:40:29+00:00` stamp is transient — it survives only until the next auto-refresh, so any consumer keying off this timestamp gets a false "freshness" signal during the intervening window.

Not a runtime bug: keeping the old timestamp (or omitting the manual bump) would have been equally correct. But it teaches the auto-integration flow to write a field it doesn't own.

**Fix:** drop the manual `_meta.updated_at` bump from the integration commit — let the updater own the field. If you want to signal "pricing was touched in this PR", add a `_meta.notes` entry or a per-model `added_at` field that lives outside the updater's write set.

tolokaforge_models/src/tolokaforge_models/certificates/registry.py
kostiantynvoiku-web: Done in c22fb48: regenerated `tests/canonical/snapshots/certify_suite_collection/collection.json` with `--update-canon` (+37 nodeids for `openrouter__z-ai_glm-5.3`, nothing else moved; re-verified without the flag). This was the branch's one `test-smoke` failure beyond `main`'s 14 pre-existing ones. Also dropped the "disposable, must NOT be merged" line from the body; the branch is now meant to merge (#1278 is stacked on it). Agreed on teaching the auto-integration flow to run `--update-canon` before opening the PR; noted for the automation.

tolokaforge_models/src/tolokaforge_models/data/model_presets.yaml
kostiantynvoiku-web: Done in c22fb48: `tests/unit/llm/test_z_ai_glm_5_3_preset.py::test_measured_shapes_route_to_the_overlay` asserts `resolve_effective_preset(...)== "z_ai_glm_5_3"` for `z-ai/glm-5.3`, `openrouter/z-ai/glm-5.3` and `litellm_proxy/z-ai/glm-5.3`, and `test_overlay_is_the_shared_stack_plus_the_codec_swap` pins the codec end (`OpenAISummaryReplayReasoningCodec`) plus adapter parity with 5.2 on every other axis. A reorder or glob broadening now fails offline.

tolokaforge_models/src/tolokaforge_models/data/model_presets.yaml
kostiantynvoiku-web: Done in c22fb48: globs are now exactly `z-ai/glm-5.3` and `*/z-ai/glm-5.3` (bare route + gateway-prefixed form), no trailing `*`. `test_unmeasured_point_releases_are_not_claimed` pins that `-turbo`, `-fast`, `5.30`, `5.3.1`, `:exacto` (and `openrouter/…-fast`) fall through to `openrouter_dict_stringify_recovery`. The YAML comment now says why (fnmatch's `*` matches `.` and `-`) and that a new point release gets its own entry with reprobe evidence.

tolokaforge_models/src/tolokaforge_models/certificates/registry.py
kostiantynvoiku-web: Done in c22fb48: captured a live z-ai/glm-5.3 OpenRouter response through litellm (2026-08-25, 287 reasoning tokens) into `tolokaforge_models/tests/unit/fixtures/openrouter_z_ai_glm_5_3_reasoning_response.json`, same message shape as the deepseek fixture (flat `reasoning_content`, `thinking_blocks` null, route-emitted `reasoning_details` = one unsigned `reasoning.text` entry). `TestZaiGlm53Fixture` pins (1) `extract` → one non-empty `summary_text` block equal to the summary, identical to the parent codec's read; (2) `encode_for_replay` rebuilds exactly the route's own envelope shape (type + text; no signature, no format). Plus `test_the_z_ai_glm_5_3_route_resolves_to_this_codec` in `TestPresetWiring`.

tolokaforge_models/src/tolokaforge_models/certificates/registry.py
kostiantynvoiku-web: Done in c22fb48: replaced the SUPERSET wording in `registry.py` with your version (FULL COPY plus a `reasoning_codec` swap; `_match_preset` returns the first match with no fallback merge, so the three fields cannot be pruned as duplicates), and aligned the YAML comment the same way so both sides say the load-bearing thing. The parity test in `test_z_ai_glm_5_3_preset.py` would also catch an accidental prune.

tolokaforge_models/src/tolokaforge_models/data/model_presets.yaml
kostiantynvoiku-web: Done in c22fb48: `test_the_rest_of_the_family_stays_on_the_shared_preset` asserts `z-ai/glm-5.1`, `z-ai/glm-5.2`, `z-ai/glm-5`, `z-ai/glm-5-turbo` and `z-ai/glm-5v-turbo` resolve to `openrouter_dict_stringify_recovery` and get the plain `OpenAIReasoningCodec`. A future `glm-5*` broadening of the overlay fails this offline.

tolokaforge_models/src/tolokaforge_models/data/pricing.json
kostiantynvoiku-web: Done in c22fb48: reverted `_meta.updated_at` to `main`'s `2026-08-14T10:17:21.788923+00:00`; `pricing.json` now differs from `main` by the `z-ai/glm-5.3` entry only. Agreed the updater owns that field; noted for the auto-integration flow.

kostiantynvoiku-web:
CI note for the merger: after c22fb48 the `test-smoke` failure set on this PR is identical to `main`'s own latest run ([32862748378](https://github.com/Toloka/tolokaforge/actions/runs/32862748378) @ ce84556): the same 26 tests (`test_run_display` ×9, `test_example_pack_grading_corpus` ×11, `test_grading_wire_lock`, `test_adapter_convert_entry_point`, the two canonical replays, models bootstrap, config_validator). None of them touch presets or certificates; the 13 example-corpus / entry-point ones arrived with #1276 / #1279 on main (`pull_request` CI runs on the merge with main). The one failure this branch used to add on top of main — `test_certify_suite_collection_snapshot` — is green now. Before c22fb48 the PR's set was main's (then 14) + that one.

CiroGamboa (approved):

Issue #1183, opened 2026-08-14
feat(grader): Context hygiene — grader plug-in receives serialisable configuration only


CiroGamboa:
## Context

The grader plug-in factory currently receives a live runtime-backend object as part of its construction context. That object is a channel to a specific runner instance chosen by the orchestrator, so a grader running on a different machine cannot use it. Replacing the live object with serialisable configuration (endpoint URL, auth token, timeout, run-scoped logger) is the smallest change that unblocks the transport-split and queue-backed variants. No user-visible behaviour change on the existing runner-RPC grader.

Part of #1181. First code-carrying PR of the milestone; safe to land while other work is in flight.

## Deliverables

- [ ] Grader plug-in construction receives only serialisable configuration.
- [ ] The existing runner-RPC grader continues to reach its target endpoint using configuration alone.
- [ ] A canonical test asserts that no live runtime-backend object is reachable from the grader construction context.
- [ ] Baseline `Grade` JSON captured against the pinned smoke config and committed to a stable location, so subsequent PRs can assert byte-identity.

## Acceptance Criteria

- All existing canonical + unit tests pass unchanged.
- Real-task smoke on the pinned smoke config produces byte-identical `Grade` JSON vs. baseline.
- A new construction-context test fails if a runtime-backend object leaks back into the context.

## Design Decisions

- **Serialise the endpoint + auth as data, do not hand over a client.** Because a serialisable context is the only shape a remote grader can accept — and the local grader loses nothing by adopting it.
- **Ship the baseline capture in this PR.** Because every subsequent PR in the milestone gates on byte-identity; the baseline needs to exist before the second PR merges.
- **No deprecation cycle needed.** Because no external code constructs the grader context directly — the seam is internal.

## Out of Scope

- Splitting the grade RPC off the runner service — separate ticket.
- Any change to the caller-facing `TrialGrader.grade` signature.

CiroGamboa:
Merged via PR #1197 into feat/grader-detachment.

PR #1328, opened 2026-08-27
fix(grader): route RunnerRPCTrialGrader through runtime_backend for per-trial runtimes

CiroGamboa:
## Summary

PR #1183 replaced the grader context's live `runtime_backend` with a
serialisable `runner_address` — a good idea for the P2/P3 out-of-process
grader path ADR-0038 pins. Side-effect: `PerTrialRuntimeBackend` (the
runtime every terminal-bench engine-loop run uses) has no `runner_address`
attribute because each trial owns its own runner endpoint.
`orchestrator._build_conductor` read `getattr(runtime_backend,
"runner_address", None)` and passed `None`; the factory built a
`GrpcRunnerClient(runner_address="")`; every `grade_trial` call failed with
`_InactiveRpcError`. Result: engine-loop trials completed cleanly but
produced no `grade.yaml` and `binary_pass=None` on every trial.

**Reproduced live** on a downstream `eval/tbench-balanced-10-engine-loop`
sample sweep: 110/110 completed kimi_k3 trials of a T-Bench balanced-10 run
carried `grading_error: "Grading failed for trial ...: gRPC error:
<_InactiveRpcError..."`. That is $50 of LLM spend with zero gradable
results — blocking the engine-loop vs Harbor scaffold comparison entirely.

## Fix — restore in-process routing without re-coupling the wire

- `TrialGraderContext` gains one optional field, `runtime_backend:
  RuntimeBackend | None = None`. Every out-of-process factory
  (`grader_rpc`, `queue`, `judge_backed`) MUST ignore it — ADR-0038's
  serialisable-configuration invariant stays intact on the wire because the
  field is `None` by default and never crosses an address boundary.
- `RunnerRPCTrialGrader.__init__` accepts an optional `runtime_backend`.
  When set, `grade` delegates to `runtime_backend.grade_trial(...)`, which
  `PerTrialRuntimeBackend._client_for(trial_id)` routes to the correct
  per-trial runner client. When unset, the pre-existing address-built
  `GrpcRunnerClient` fallback runs — the P2/P3 shape is bit-identical to
  today.
- `runner_rpc_trial_grader_factory` threads the new field through.
- `orchestrator._build_conductor` passes `runtime_backend=runtime_backend`
  alongside the still-populated `runner_address`.

No adapter changes — the terminal-bench adapter already emits the correct
`grading_method="test_execution"`.

## Test plan

- [x] Stage 1 (context hygiene, canonical): 11 tests — extends
  `test_trial_grader_context_hygiene.py` with (a) new
  `test_runtime_backend_is_optional_and_defaults_to_none`, (b) new
  `TestOutOfProcessFactoriesIgnoreRuntimeBackend` with both `ast.walk`
  (static) and sentinel-backend (runtime) checks on all three
  out-of-process factories.
- [x] Stage 2 (dispatch preference, unit): 3 tests —
  `TestRuntimeBackendDispatchPreference` locks the
  `runtime_backend`-wins-when-set contract and the `runner_client`
  fallback for the P2/P3 shape.
- [x] Stage 3 (end-to-end regression, canonical): new
  `test_trial_grader_per_trial_backend_routing.py` — wires
  `PerTrialRuntimeBackend` + `runner_rpc` + a `USER_STOP`-terminated
  trajectory and asserts a `Grade` returns. On `main` before this fix,
  that test hangs on the phantom-address health-check loop and raises
  `ConnectionError`.
- [x] Full grader-wide test sweep (`test_grader_rpc_trial_grader.py`,
  `test_grader_queue_wire_v2.py`, `test_judge_backed_trial_grader.py`,
  `test_grader_composite_dispatch.py`, `test_grader_end_to_end.py`,
  `test_grader_service_contract.py`, `test_grader_wire_v2.py`): 122
  passed, no P2/P3 regressions.
- [x] Orchestrator + grader surface: `pytest -k "trial_grader or
  _build_conductor or grader"` — 216 passed.
- [x] `ruff check` + `black --check` + `import-linter`: all clean.
- [ ] Live smoke: cut `v0.21.3`, bump the pin on the downstream engine-loop
  task-set branch, re-dispatch the same kimi_k3 sample. Success signal =
  `grade.yaml` files present in every completed trial's bundle, and the
  aggregate reports a real pass rate (not 0/N).

## ADR-0038 compliance

The escape hatch is `None` by default. The out-of-process factories are
verified — statically and at runtime — to never read it. The wire never
carries a live backend. A one-paragraph ADR-0038 addendum under
"Consequences → In-process routing" would be a doc-only follow-up if the
reviewer wants a formal record.

### DIFF
--- a/docs/adr/0038-grader-detachment.md
+++ b/docs/adr/0038-grader-detachment.md
@@ -64,7 +64,7 @@ Adopt the extraction target #1, the throughput property #2 (both variants ship),
 
 Concretely:
 
-1. **Grader plug-in receives serialisable configuration only.** The construction context becomes endpoint URL + auth token + timeout + run-scoped logger. No live runtime-backend object crosses the seam.
+1. **Grader plug-in receives configuration only — nothing live crosses the seam.** The construction context is endpoint URL + auth token + timeout + run-scoped logger, plus one optional in-process routing shim (`runtime_backend`) populated only for backends whose runner endpoints are reachable per-trial (`PerTrialRuntimeBackend` — each trial owns its own endpoint, so no static address applies). The shim MUST be `None` whenever the grader will cross an address boundary; every out-of-process factory (`grader_rpc`, `queue`, `judge_backed`) MUST ignore the field. The hygiene test at `tests/canonical/test_trial_grader_context_hygiene.py` pins the ignore contract statically (AST) and at runtime (sentinel-backend) for every bundled out-of-process factory.
 2. **New RPC contract for the grader, owned by the grader.** The runner's grade RPC ships one more release with a deprecation notice, then goes away.
 3. **Standalone grader service binary.** Embeds the pure evaluator plus the runner-specific grader glue. No tool execution, no sandbox, no env-state.
 4. **New grader plug-in registered under the existing plugin group,** reaching the new service.
--- a/tests/canonical/test_trial_grader_context_hygiene.py
+++ b/tests/canonical/test_trial_grader_context_hygiene.py
@@ -132,7 +132,17 @@ class TestOutOfProcessFactoriesIgnoreRuntimeBackend:
     every out-of-process factory. A leak of the field into ``grader_rpc``,
     ``queue``, or ``judge_backed`` is a Blocker — a live backend at
     deployment time it cannot marshal. Both static (``ast.walk``) and runtime
-    (sentinel-backend) checks are enforced here."""
+    (sentinel-backend) checks are enforced here.
+
+    Static AST checks cover all three factories via
+    ``test_factory_source_does_not_reference_runtime_backend``. Runtime
+    sentinel-backend checks cover ``grader_rpc`` and ``judge_backed`` — the
+    queue factory spawns worker threads and constructs real
+    ``GrpcGraderClient`` instances mid-body, so a runtime sentinel would
+    either need broad monkey-patching (opaque) or skip the second half of
+    the factory body (weak); the AST check is strictly stronger for that
+    factory anyway.
+    """
 
     _OUT_OF_PROCESS_FACTORIES = (
         grader_rpc_trial_grader_factory,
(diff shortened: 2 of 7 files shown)


CiroGamboa:
## Live smoke — the fix works end-to-end

Ran a single-trial local smoke on `fix-billing-holds` (the same task pack the engine-loop sweep hit): `moonshotai/kimi-k3` via openrouter, `agent_harness: engine-loop`, `runtime: per_trial`, `docker.image_source: build`, `max_turns: 12`.

**Result: real grade written to disk.**

| Artefact | Before fix (main) | After fix (this PR) |
|---|---|---|
| `grade.yaml` on disk | missing | **present** |
| `binary_pass` | `None` | `False` |
| `score` | `None` | **`0.433333`** (real T-Bench pytest reward, 13/30 tests passed) |
| `grading_error` on trajectory | `"gRPC error: <_InactiveRpcError..."` | `null` |
| Cost | wasted ($50 for 110/110 ungraded) | $0.16 for one graded trial |

The full grade path fired: `RunnerRPCTrialGrader.grade` → `PerTrialRuntimeBackend.grade_trial` → per-trial `GrpcRunnerClient` → gRPC `GradeTrial` → `_grade_via_test_execution` → pytest inside the trial container → reward. 30 T-Bench cases actually ran; agent solved 13. Trajectory bundle at `trials/fix-billing-holds/0/` carries the full 9-file layout including the previously-missing `grade.yaml`.

Log excerpt:
```
09:38:25.504 | INFO | Graded trial fix-billing-holds:0: pass=False, score=0.433333
09:38:25.505 | INFO | binary_pass=False score=0.433333 task_id=fix-billing-holds trial_index=0 | Grading via Runner RPC
09:38:25.505 | INFO | binary_pass=False score=0.433333 task_id=fix-billing-holds trial_index=0 | Trial graded
09:38:25.575 | INFO | output_dir=results/.../trials/fix-billing-holds/0 | Trial output saved
```

PR #1407, opened 2026-09-01
feat(llm): namespace-matched gateway wildcards, one provider-pin rule, route provenance

balazs-toloka:
## Motivation (live findings on a production LiteLLM deployment)

Two measured defects when routing evals through an OpenAI-compatible gateway whose OpenRouter coverage is the `openrouter/*` passthrough:

1. **Every OpenRouter model fell back to direct calls.** `resolve_gateway_route` accepts exact catalog entries only; a live probe proved all 19 "missing" models actually work through the gateway as `openrouter/<slug>`. The per-model fallback is loud but silently splits one run across two wire paths (agent direct, sim/judge via gateway).
2. **Provider pinning silently dropped on the resolved path.** `_build_kwargs` popped `extra_body.provider` whenever a route resolved, even when the route forwards to OpenRouter itself, which honours the pin. Meanwhile the unreadable-catalog path forwarded the pin - two unit tests enshrined the contradiction (`test_llm_proxy.py` vs `test_gateway_routing_applied.py`).

## What changes

- **One provider-pin rule on both gateway paths:** `extra_body.provider` survives exactly when the wire name's first segment is the model's own provider namespace; any other namespace drops it with a warning. Docs and both tests now state the same rule.
- **Opt-in namespace-matched wildcards** (`LLM_PROXY_TRUST_NAMESPACE_WILDCARDS`, default `false`, strict-bool, included in the orphan check): `<ns>/*` routes `provider: <ns>` models by their untranslated name. This implements exactly the extension point the existing wildcard ban in `docs/LLM_LAYER.md` reserved - a foreign-namespace wildcard never routes (the measured `anthropic/*`-to-Bedrock serving-path break stays impossible), exact entries always win, and the `GatewayRouteError` ambiguity semantics are untouched.
- **Per-call route provenance:** `gateway_route` + `gateway_route_kind` (`exact`|`wildcard`) on `ProviderRawCall`, so run artifacts record which serving path produced a number (additive optional fields; `ModelConfig` and grading hashes untouched - `ModelConfig` feeds no hash/fingerprint).
- **Live verification, not declaration:** new `test_gateway_live.py` case sends a provider-pinned call through the gateway and checks the actually-serving upstream via OpenRouter's `/generation` lookup, because litellm proxies may drop `extra_body` fields version-dependently and the request shape alone proves nothing. Opt-in via `LLM_PROXY_INT_TEST_PINNED_MODEL` / `_PINNED_PROVIDER`, auto-skips otherwise.
- **Companion automation fix:** `gateway_catalog.lookup()` derived the wildcard namespace only from the bare slug, so an `openrouter/*`-only gateway reported every model ABSENT and the Slack poller downgraded explicit "via litellm" requests. Namespace now derives per candidate, in the exact-check's order, and `route` is the matched candidate.

## Deployment-neutrality

No deployment specifics enter the engine: the new variable is a generic statement about any gateway ("do you trust namespace passthroughs"), default-off preserves today's behavior byte-for-byte for every existing deployment, and all values keep resolving through `SecretManager`.

## Verification

- `tests/unit/llm/` green (the 3 `test_rate_limit_probe_retry.py` failures pre-exist on `main@41537dc0`, unrelated); resolver, proxy-parse, client-routing and automation-lookup suites extended.
- Canonical contracts green: `test_llm_gateway_envelope_contract.py`, `test_cost_extraction_canon.py`.
- ruff check + format clean on touched files.

Internal ref: TECHDEL-598.

### DIFF
--- a/tolokaforge/core/llm/client.py
+++ b/tolokaforge/core/llm/client.py
@@ -1867,12 +1867,16 @@ class LLMClient:
                 extra_body = kwargs.get("extra_body")
                 if isinstance(extra_body, dict) and "provider" in extra_body:
                     extra_body.pop("provider")
-                    self.logger.warning(
-                        "Dropping the OpenRouter provider pin: the gateway route "
-                        "targets another namespace, whose upstream rejects it",
-                        route=wire_model,
-                        provider=self.provider,
-                    )
+                    if not getattr(self, "_pin_drop_warned", False):
+                        # Once per client, not per call: a full eval makes
+                        # thousands of calls and the fact does not change.
+                        self._pin_drop_warned = True
+                        self.logger.warning(
+                            "Dropping the OpenRouter provider pin: the gateway route "
+                            "targets another namespace, whose upstream rejects it",
+                            route=wire_model,
+                            provider=self.provider,
+                        )
                     if not extra_body:
                         kwargs.pop("extra_body")
             kwargs["api_base"] = self._proxy.base_url
@@ -2204,7 +2208,9 @@ class LLMClient:
             latency_s=latency_s,
             cost_usd=cost_usd,
             cost_source=cost_source,
-            gateway_route=self._gateway_route,
+            # Plain str on purpose: the artifact carries data, not the resolver's
+            # str subclass (asdict deepcopies every call record).
+            gateway_route=str(self._gateway_route) if self._gateway_route is not None else None,
             gateway_route_kind=self._gateway_route_kind,
         )
 
--- a/tolokaforge/core/llm/client.py
+++ b/tolokaforge/core/llm/client.py
@@ -614,6 +614,7 @@ class LLMClient:
         # to the direct provider rather than post a name it cannot route.
         # docs/LLM_LAYER.md § Speaking to the gateway.
         self._gateway_route: str | None = None
+        self._gateway_route_kind: str | None = None
         if self._proxy is not None:
             catalog = fetch_gateway_catalog(self._proxy)
             if catalog is None:
@@ -628,8 +629,18 @@ class LLMClient:
                 # A readable catalog is never empty here: the fetch maps an empty
                 # answer to None, so resolver-None below can only mean "omitted".
                 self._gateway_route = resolve_gateway_route(
-                    self.model_name, catalog, self._proxy.preferred_route
+                    self.model_name,
+                    catalog,
+                    self._proxy.preferred_route,
+                    trust_namespace_wildcards=self._proxy.trust_namespace_wildcards,
                 )
+                self._gateway_route_kind = getattr(self._gateway_route, "kind", None)
+                if self._gateway_route_kind == "wildcard":
+                    self.logger.info(
+                        "Gateway route resolved via the model namespace wildcard",
+                        base_url=self._proxy.base_url,
+                        model=self.model_name,
+                    )
                 if self._gateway_route is None:
                     self.logger.warning(
                         "Gateway does not serve this model; calling the provider directly",
@@ -1843,11 +1854,25 @@ class LLMClient:
             if route is not None:
                 kwargs["model"] = route
                 kwargs["custom_llm_provider"] = "openai"
-                # Same class as the usage extension the dialect switch removes: an
-                # OpenRouter-only body field a non-OpenRouter upstream rejects.
+            # ONE rule for the provider pin on BOTH gateway paths (resolved route
+            # and unreadable catalog): ``extra_body.provider`` survives the hop
+            # exactly when the wire name's first segment IS the model's own
+            # provider namespace - the upstream behind such a route is the same
+            # family the pin was written for, and honours it. A route into any
+            # other namespace is another upstream, which rejects the field or,
+            # worse, silently ignores it. docs/LLM_LAYER.md § Speaking to the
+            # gateway.
+            wire_model = str(kwargs.get("model", self.model_name))
+            if wire_model.split("/", 1)[0] != self.provider.split("/", 1)[0]:
                 extra_body = kwargs.get("extra_body")
-                if isinstance(extra_body, dict):
-                    extra_body.pop("provider", None)
+                if isinstance(extra_body, dict) and "provider" in extra_body:
+                    extra_body.pop("provider")
+                    self.logger.warning(
+                        "Dropping the OpenRouter provider pin: the gateway route "
+                        "targets another namespace, whose upstream rejects it",
+                        route=wire_model,
+                        provider=self.provider,
+                    )
                     if not extra_body:
                         kwargs.pop("extra_body")
             kwargs["api_base"] = self._proxy.base_url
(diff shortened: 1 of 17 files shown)


CiroGamboa (commented):
Three findings from a review pass — one plausible correctness gap between the engine's resolver and the automation catalog lookup, two annotation/init cleanups. Details inline.

tolokaforge/core/llm/gateway_route.py
```diff
@@ -135,17 +171,24 @@ def resolve_gateway_route(
 
     hits = [c for c in _candidates(model_string) if c in catalog]
     if not hits:
+        if trust_namespace_wildcards and "/" in model_string:
+            namespace = model_string.split("/", 1)[0]
```
CiroGamboa, line 180: **Wildcard asymmetry between engine and automation resolver (plausible)**

`resolve_gateway_route` only checks the *full* `model_string`'s own namespace against `<ns>/*` in the catalog. The automation lookup in `tools/automation/src/automation/gateway_catalog.py:146-149` now iterates every candidate from `_candidates(...)` — both the prefixed and the bare form — and matches each candidate's namespace against wildcards.

For an `openrouter/*`-only catalog the two agree, so the headline scenario is fine. But for a *bare-vendor* wildcard like `x-ai/*` or `anthropic/*` with `LLM_PROXY_TRUST_NAMESPACE_WILDCARDS=true`:

- Slack `/integrate-model x-ai/grok-4.5`: automation matches candidate `x-ai/grok-4.5`'s namespace `x-ai` against `x-ai/*` → poller reports `STATUS_WILDCARD`, reachable via litellm.
- The workflow then runs with `provider="openrouter"` (the integrate-model.yml comment states it's always openrouter), so `model_string="openrouter/x-ai/grok-4.5"`. Here the engine only checks namespace `openrouter` against `openrouter/*` — not present → returns `None` → the client silently falls back to the direct provider.

That undermines the new slack-integrate.yml comment ("the poller reads the catalog with the same routing policy a run uses, so its wildcard-coverage verdicts match what the run will honour").

Suggested fix: mirror the automation's candidate iteration here — walk each candidate from `_candidates(model_string)` and check `<candidate-namespace>/*` against the catalog before returning `None`.

tolokaforge/core/llm/client.py
```diff
@@ -614,6 +614,7 @@ def __init__(
         # to the direct provider rather than post a name it cannot route.
         # docs/LLM_LAYER.md § Speaking to the gateway.
         self._gateway_route: str | None = None
```
CiroGamboa: **Type annotation understates the runtime type**

`self._gateway_route: str | None = None` — but the assignment at line 631 is `resolve_gateway_route(...)`, which now returns `ResolvedGatewayRoute | None`. That's why the very next new line has to defensively coalesce: `getattr(self._gateway_route, "kind", None)`.

Not a bug today (ResolvedGatewayRoute is a str subclass and every reader either does an identity check or a `str(...)` demotion), but the annotation hides the `kind` axis from any future reader — including type checkers. Widening to `ResolvedGatewayRoute | None` (importing from `gateway_route`) would remove the need for the getattr guard on line 637.

tolokaforge/core/llm/client.py
```diff
@@ -1843,11 +1854,29 @@ def _build_kwargs(
...
+            # gateway.
+            wire_model = str(kwargs.get("model", self.model_name))
+            if wire_model.split("/", 1)[0] != self.provider.split("/", 1)[0]:
                 extra_body = kwargs.get("extra_body")
-                if isinstance(extra_body, dict):
-                    extra_body.pop("provider", None)
+                if isinstance(extra_body, dict) and "provider" in extra_body:
+                    extra_body.pop("provider")
+                    if not getattr(self, "_pin_drop_warned", False):
+                        # Once per client, not per call: a full eval makes
```
CiroGamboa, line 1881: **`_pin_drop_warned` read via `getattr` but never initialized in `__init__`**

The sibling fields added in this PR (`_gateway_route`, `_gateway_route_kind`) are declared at the top of `__init__` around 616-617, but `_pin_drop_warned` only ever comes into existence on the first pin-drop path — every earlier read has to go through `getattr(self, "_pin_drop_warned", False)`.

Not a runtime bug, but it puts the flag outside the client's declared attribute set. A future move to `__slots__`, a mypy strict pass, or a subclass overriding `__init__` would surface this. Cleanup: initialize `self._pin_drop_warned: bool = False` alongside `self._gateway_route_kind` and drop the `getattr` default.

tolokaforge/core/llm/gateway_route.py
bberkes-toloka: Real asymmetry, thanks - fixed in `5442efb7`, but in the opposite direction to the suggestion, and I want to be explicit about why.

The engine deriving the namespace from the FULL model string is the deliberate rule, not an oversight. In your scenario the run addresses `openrouter/x-ai/grok-4.5`; an `x-ai/*` passthrough forwards to xAI directly, so honouring it would swap the upstream from OpenRouter to xAI - a serving-path change, which is the exact break the namespace-match design refuses (docs/LLM_LAYER.md, and why an exact entry is trusted for either shape while a wildcard is not). Walking the candidates in the engine would reintroduce that.

So the loose side is the automation lookup, and the fix deletes it rather than mirroring it: `lookup()` now calls `resolve_gateway_route()` with the model string the run builds. The two agree by construction, which also picks up the two gaps the bot review flagged on the same code - bare `*` is no longer accepted, and the wildcard match is gated on `LLM_PROXY_TRUST_NAMESPACE_WILDCARDS` via a new `configured_policy()` (env-only, same property as the catalog read). One behaviour change worth knowing: an ambiguous catalog (both names, no `LLM_PROXY_PREFERRED_ROUTE`) is now reported *unknown* instead of silently picking the prefixed name, because a run raises `GatewayRouteError` there - the plan keeps the OpenRouter default instead of queueing a run that dies at client construction. `docs/AUTO_INTEGRATION.md` and the availability table are updated.

tolokaforge/core/llm/client.py
bberkes-toloka: Fixed in `5442efb7`: `_gateway_route` is now annotated `ResolvedGatewayRoute | None` and the defensive `getattr` is gone (plain `.kind` behind a None check). `kind` itself is now `RouteKind = Literal["exact", "wildcard"]`, exported from `gateway_route`, and the same alias types `LLMClient._gateway_route_kind`, `ProviderRawCall.gateway_route_kind` and `UsageExtractor.extract(...)` - so the two-value contract is stated once rather than in four places (also the bot review nit #3).

tolokaforge/core/llm/client.py
bberkes-toloka: Fixed in `5442efb7`: `self._pin_drop_warned: bool = False` is declared in `__init__` next to the gateway fields, and the drop site reads it directly.

CiroGamboa (approved):

PR #1449, opened 2026-09-02
integrate: anthropic/claude-fable-5.1

balazs-toloka:
# integrate: anthropic/claude-fable-5.1

Adds `openrouter` / `anthropic/claude-fable-5.1` (`openrouter__anthropic_claude-fable-5.1`) as a
certified model. Integrated via **auto-resolve** — the observe → fix → reprobe loop converged on
iteration 1, and this branch is the mechanically-landed result of the proven overlay.

**Engine ref:** `51a092e`.

## Policy

One model-scoped preset, `anthropic_claude_fable_5_1`, declared before the generic `anthropic`
preset so first-match-wins routing picks it up for `anthropic/claude-fable-5.1*` /
`*claude-fable-5.1*` only. No shared glob was broadened; no new adapter class and no engine-side
change (data-only).

It restates the generic `anthropic` axes verbatim (content / reasoning / cache = anthropic,
`supports_seed: false`) plus one fix: `param_value_rules.reasoning_effort.medium → override with
high`. The observe baseline's only failure was `thinking_emits_blocks` 5/15 (all 10 failures
`reasoning is None`), and it is the only probe sending `effort_hint="medium"` — the same codec and
transport passed 15/15 at `"high"` and 15/15 with signed blocks in budget mode. "medium"
under-allocates the adaptive thinking budget, so the model often declines to think at all; that is
effort-routing under-allocation, not a capability ceiling. Reprobe: **5/5** on the fix-target.

## Ceilings

None. `known_unsupported` is empty and all 23 probed capabilities are `required`, including
`THINKING_EMITS_BLOCKS`, `UNSIGNED_THINKING_REPLAY`, `IMPLICIT_PROMPT_CACHING` and
`DISCRIMINATED_UNION_TOOL_CALL`.

## Changes

- `model_presets.yaml` — new `anthropic_claude_fable_5_1` preset entry.
- `certificates/registry.py` — new `MC(...)` cert (23 required, 0 known_unsupported).
- `data/pricing.json` — `anthropic/claude-fable-5.1` pricing (10.0 / 50.0 / 0.25 USD per 1M in / out / cache_read).
- `tests/canonical/snapshots/certify_suite_collection/collection.json` — regenerated for the new cert's probe nodeids.

> ⚠️ **Disposable test branch — do not merge.** This branch exists to exercise the auto-resolve
> integration path; it must not land on `main`.

### DIFF
--- a/tolokaforge_models/src/tolokaforge_models/certificates/registry.py
+++ b/tolokaforge_models/src/tolokaforge_models/certificates/registry.py
@@ -520,6 +520,63 @@ _ALL: list[MC] = [
             }
         ),
     ),
+    # Claude Fable 5.1 — Anthropic adaptive thinker, landed via auto-resolve
+    # (Slack-requested integration, PR #1449). Routes through the
+    # model-specific ``anthropic_claude_fable_5_1`` preset (see
+    # model_presets.yaml): the generic ``anthropic`` axes (content/reasoning/
+    # cache = anthropic, supports_seed false) PLUS
+    # ``param_value_rules.reasoning_effort.medium -> override with high``.
+    #
+    # The observe (default) baseline surfaced ONE preset-fixable failure —
+    # ``thinking_emits_blocks`` 5/15, every failure the same assertion
+    # ("StructuredReasoning should be surfaced"). That probe is the only one
+    # sending effort_hint="medium"; the two siblings on the SAME anthropic
+    # reasoning codec and OpenRouter transport passed 15/15 — unsigned replay at
+    # effort="high" and signed replay at budget_tokens=4000 — so the model does
+    # emit structured signed thinking and "medium" merely under-allocates the
+    # adaptive budget. Overriding medium -> high took the reprobe to 5/5 on the
+    # fix-target, so THINKING_EMITS_BLOCKS is ``required`` here.
+    #
+    # Every other capability passed the observe baseline 15/15 under the
+    # anthropic axes and none went known_unsupported (decision.json ceilings
+    # were empty) — so, like the opus-5 sibling and UNLIKE
+    # opus-4.6/4.7/4.8/fable-5, UNSIGNED_THINKING_REPLAY,
+    # IMPLICIT_PROMPT_CACHING and DISCRIMINATED_UNION_TOOL_CALL are all
+    # required: this route recorded no such ceiling.
+    MC(
+        model_id="openrouter__anthropic_claude-fable-5.1",
+        provider="openrouter",
+        name="anthropic/claude-fable-5.1",
+        env_key="OPENROUTER_API_KEY",
+        required=frozenset(
+            {
+                C.ALLOF_MERGE_TOOL_CALL,
+                C.BASIC_COMPLETION,
+                C.COST_USD_POPULATED,
+                C.DECIMAL_FIELD_TOOL_CALL,
+                C.DICT_MAP_TOOL_CALL,
+                C.DISCRIMINATED_UNION_TOOL_CALL,
+                C.ENUM_SLASH_TOLERANCE,
+                C.HETEROGENEOUS_ARRAY_TOOL_CALL,
+                C.IMPLICIT_PROMPT_CACHING,
+                C.LEXICAL_TOOL_INVENTION,
+                C.MULTI_TURN_ERROR_RECOVERY,
+                C.MULTI_TURN_TOOL_USE,
+                C.PROGRESS_AFTER_SUCCESS,
+                C.PROMPT_CACHING,
+                C.RE2_PATTERN_TOLERANCE,
+                C.RECURSIVE_REF_TOOL_CALL,
+                C.REQUIRED_FIELDS_COMPLETE,
+                C.SIMPLE_TOOL_CALL,
+                C.THINKING_EMITS_BLOCKS,
+                C.THINKING_REPLAY_ROUNDTRIP,
+                C.TOOL_NAME_DISCIPLINE,
+                C.UNSIGNED_THINKING_REPLAY,
+                C.USAGE_METRICS_POPULATED,
+            }
+        ),
+        known_unsupported=frozenset(),
+    ),
     # Claude Opus 5 — Anthropic adaptive thinker, landed via auto-resolve
     # (Slack-requested integration, PR #614). Routes through the model-specific
     # ``anthropic_claude_opus_5`` preset (see model_presets.yaml): the generic
--- a/tolokaforge_models/src/tolokaforge_models/data/pricing.json
+++ b/tolokaforge_models/src/tolokaforge_models/data/pricing.json
@@ -2,7 +2,7 @@
   "_meta": {
     "description": "Model pricing in USD per 1M tokens. Source: OpenRouter API.",
     "source_url": "https://openrouter.ai/api/v1/models",
-    "updated_at": "2026-08-14T10:17:21.788923+00:00",
+    "updated_at": "2026-09-02T09:51:31.333093+00:00",
     "notes": "Run `uv run pricing-updater update` to refresh from OpenRouter API."
   },
   "models": {
@@ -143,6 +143,12 @@
       "input": 10.0,
       "output": 50.0,
       "cache_read": 1.0,
+      "cache_write": 12.5
+    },
+    "anthropic/claude-fable-5.1": {
+      "input": 10.0,
+      "output": 50.0,
+      "cache_read": 0.25,
       "cache_write": 12.5
     },
     "anthropic/claude-haiku-4.5": {
--- a/tolokaforge_models/src/tolokaforge_models/data/model_presets.yaml
+++ b/tolokaforge_models/src/tolokaforge_models/data/model_presets.yaml
@@ -115,6 +115,69 @@ presets:
     params:
       supports_seed: false
 
+  # Claude Fable 5.1 — landed via auto-resolve (Slack-requested integration,
+  # PR #1449). Declared BEFORE the generic ``anthropic`` preset so
+  # first-match-wins routing picks it up. It is the generic ``anthropic`` preset
+  # this model already resolved to during observe (content/reasoning/cache =
+  # anthropic, supports_seed: false — restated verbatim so the superset rule
+  # holds) PLUS one fable-5.1-observed fix, scoped to THIS model's match globs
+  # so no sibling Anthropic route is affected:
+  #
+  #   * ``param_value_rules.reasoning_effort.medium -> override with high``. The
+  #     observe (default) baseline surfaced exactly ONE failure:
+  #     ``thinking_emits_blocks`` 5/15, all 10 failures the same assertion
+  #     ("StructuredReasoning should be surfaced", i.e. ``reasoning is None``).
+  #     The two sibling probes on the SAME anthropic reasoning codec and the
+  #     SAME OpenRouter transport passed 15/15 — ``unsigned_thinking_replay``
+  #     (adaptive, effort_hint="high") and ``thinking_replay_roundtrip``
+  #     (budget_tokens=4000), and the latter asserts *signed* turn-1 blocks — so
+  #     the model demonstrably emits structured signed thinking and the codec
+  #     demonstrably decodes it. The sole discriminator is the effort LEVEL:
+  #     only the failing probe sends effort_hint="medium". At
+  #     ``extra_body.reasoning={"effort":"medium","enabled":true}`` OpenRouter
+  #     allocates a small budget and this adaptive thinker frequently declines
+  #     to think at all; at "high" it always does. Effort-routing
+  #     under-allocation, not a capability gap — hence a params fix rather than
+  #     a ceiling. ``override`` (not ``reject``, which raises; not ``drop``,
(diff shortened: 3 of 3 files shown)


tolokaforge_models/src/tolokaforge_models/certificates/registry.py
```diff
@@ -520,6 +520,63 @@
...
+            {
+                C.ALLOF_MERGE_TOOL_CALL,
+                C.BASIC_COMPLETION,
+                C.COST_USD_POPULATED,
+                C.DECIMAL_FIELD_TOOL_CALL,
+                C.DICT_MAP_TOOL_CALL,
+                C.DISCRIMINATED_UNION_TOOL_CALL,
+                C.ENUM_SLASH_TOLERANCE,
+                C.HETEROGENEOUS_ARRAY_TOOL_CALL,
+                C.IMPLICIT_PROMPT_CACHING,
```
CiroGamboa, line 561: **`IMPLICIT_PROMPT_CACHING` marked required contradicts the fable-5 sibling's ceiling on the same route**

The fable-5 cert (this file, line 511) states:
> Anthropic has no implicit auto-cache surface, so the classification holds regardless of the synthetic pass/fail... flakily passes the synthetic probe (3/5 live)

fable-5.1 rides the exact same OpenRouter transport with `cache_policy: anthropic_ephemeral`. Anthropic still has no implicit auto-cache — the probe that passes here is really measuring **explicit** cache_control markers leaking through. The observe pass isn't evidence that this changed on 5.1; it's the same synthetic flake the fable-5 comment already documented.

Expect the same 2-of-11 style CI flakes fable-5 originally saw, or a lasting misclassification of explicit-cache behaviour as implicit. Prefer moving this into `known_unsupported` and carrying the fable-5 ceiling comment forward.

tolokaforge_models/src/tolokaforge_models/certificates/registry.py
```diff
@@ -520,6 +520,63 @@
...
+                C.PROGRESS_AFTER_SUCCESS,
+                C.PROMPT_CACHING,
+                C.RE2_PATTERN_TOLERANCE,
+                C.RECURSIVE_REF_TOOL_CALL,
+                C.REQUIRED_FIELDS_COMPLETE,
+                C.SIMPLE_TOOL_CALL,
+                C.THINKING_EMITS_BLOCKS,
+                C.THINKING_REPLAY_ROUNDTRIP,
+                C.TOOL_NAME_DISCIPLINE,
+                C.UNSIGNED_THINKING_REPLAY,
```
CiroGamboa, line 574: **`UNSIGNED_THINKING_REPLAY` marked required despite Anthropic emitting only the signed shape**

fable-5's cert (lines 512–519) keeps this capability in `known_unsupported` because:
> `test_unsigned_thinking_replay` asserts the Gemini-lineage `reasoning_details` / `reasoning.text` replay shape. Anthropic's real replay contract is the SIGNED variant (`THINKING_REPLAY_ROUNDTRIP`, required above). Kept consistent with opus-4.6/4.7/4.8. Pass-but-wrong-shape.

fable-5.1 uses the identical `reasoning_codec: anthropic` on the identical OpenRouter route, so a passing probe here is still wrong-shape — the model emits signed replays, and the unsigned probe happens to accept whatever it decoded. Promoting this to `required` locks in an incorrect classification of the same behaviour every other opus-4.x / fable-5 cert deliberately opted out of.

tolokaforge_models/src/tolokaforge_models/certificates/registry.py
```diff
@@ -520,6 +520,63 @@
...
+        name="anthropic/claude-fable-5.1",
+        env_key="OPENROUTER_API_KEY",
+        required=frozenset(
+            {
+                C.ALLOF_MERGE_TOOL_CALL,
+                C.BASIC_COMPLETION,
+                C.COST_USD_POPULATED,
+                C.DECIMAL_FIELD_TOOL_CALL,
+                C.DICT_MAP_TOOL_CALL,
+                C.DISCRIMINATED_UNION_TOOL_CALL,
```
CiroGamboa, line 558: **`DISCRIMINATED_UNION_TOOL_CALL` required with no `json_coerce` recovery**

The opus-5 preset in `model_presets.yaml` sets `response_policy: json_coerce` specifically because:
> `discriminated_union_two_turns[explicit_discriminator]` 0/15, where the model stringified the item union member.

The new `anthropic_claude_fable_5_1` preset here declares no `response_policy`, so it falls back to the generic `anthropic` preset's default. The observe baseline reports 15/15 today, but the sibling shows Anthropic thinkers *do* stringify discriminated-union members — if a future probe run hits that mode, `required` with no recovery blocks the promote path. Safer to mirror the opus-5 preset's `json_coerce` policy here.

tolokaforge_models/src/tolokaforge_models/data/pricing.json
```diff
@@ -145,6 +145,12 @@
       "cache_read": 1.0,
       "cache_write": 12.5
     },
+    "anthropic/claude-fable-5.1": {
+      "input": 10.0,
+      "output": 50.0,
+      "cache_read": 0.25,
```
CiroGamboa, line 151: **`cache_read` diverges from every sibling and from the Anthropic `input/10` norm**

At identical `input=10.0` / `output=50.0`, the sibling `anthropic/claude-fable-5` (line 145 above) has `cache_read: 1.0`. Every Anthropic entry in this file follows the `cache_read = input / 10` ratio:
- claude-haiku-4.5: 1.0 / 0.1
- claude-opus-4.6: 5.0 / 0.5
- claude-opus-4.5: 5.0 / 0.5
- claude-opus-4: 15.0 / 1.5

`0.25` here is 2.5% of input rather than 10%, and 4× lower than the sibling. If the upstream OpenRouter response the updater ingested was stale or wrong, `estimate_cost()` under-reports cache-read spend by 4× on every fable-5.1 call, silently biasing cost dashboards and cost-vs-quality comparisons against every other model on the same transport. Worth cross-checking against the Anthropic price sheet before merge.

CiroGamboa (commented):
Four findings from a review pass — all plausible, all pointing at the same shape: the auto-resolve promoted every observe-pass capability to `required` without carrying over the sibling-documented ceilings on this exact anthropic-on-OpenRouter route. Plus a cache_read price that looks off vs. the sibling and the pattern of the rest of the file. Details inline.

tolokaforge_models/src/tolokaforge_models/data/pricing.json
bberkes-toloka: Checked against the primary source: the value is correct, not stale.

Anthropic's pricing page (platform.claude.com/docs/en/about-claude/pricing) lists Claude Fable 5.1 at $10 / $50 with cache hits at **$0.25 / MTok**, footnoted: "Cache hits and refreshes on Claude Fable 5.1 and Claude Mythos 5.1 are priced at 0.025x the base input price. All other models use the standard 0.1x multiplier." Fable 5 stays at $1.00, so the 4x gap to the sibling is the documented exception, and the `input/10` pattern in the rest of the file is the rule this model was explicitly carved out of. OpenRouter carries the same number (`input_cache_read: 0.00000025`).

`estimate_cost()` would be over-reporting cache-read spend by 4x if we changed it to 1.0. Keeping 0.25.

tolokaforge_models/src/tolokaforge_models/certificates/registry.py
bberkes-toloka: The observe artifact says this model does not have the quirk the policy exists for. From `findings.json`:

- `test_discriminated_union_tool_call_two_turns[...-explicit_discriminator]`: 15/15
- `test_discriminated_union_tool_call_two_turns[...-bare_union]`: 15/15
- `test_variant_discriminated_union[array_of_unions | nested_union | union_in_dict_map]`: 15/15 each

Opus-5 got `json_coerce` because it stringified the `item` member 0/15 on `explicit_discriminator`. Fable 5.1 round-trips it natively on every variant. `JsonCoerceResponse` documents itself as "recovery, not transformation" for a model-side quirk, and the auto-resolve contract for a preset is the generic axes plus the observed fix, scoped to what the evidence showed. Adding an unobserved recovery policy would be a precautionary deviation from that, and it is not free: it rewrites any string argument that starts with `{` or `[` and parses as JSON.

If the stringified mode ever appears at eval volume, it is a one-line wheel release at that point, with the trajectory as evidence. I would keep the preset as generated.

tolokaforge_models/src/tolokaforge_models/certificates/registry.py
bberkes-toloka: Agree with the mechanism, disagree with the fix, because the pipeline's own gate refuses it.

`test_implicit_prompt_caching` is 15/15 on this route (`findings.json`). `tools/automation/src/automation/cert.py::reconcile` classifies a `known_unsupported` capability whose baseline pass rate is >= 0.9 as **FALSE-PESSIMISM, a violation (exit 1)**: "the cert must follow the synthetic result; mark it `required`." `docs/ADD_NEW_MODEL.md` (section "copying a sibling cert verbatim") says the same in prose: flip every sibling ceiling to `required`, run live, and move back "only the capabilities that actually failed live". Demoting a 15/15 capability here would fail `cert_reconcile` on this PR.

The fable-5 demotion you quote was hand-made before the automation existed, and its own comment records a *flaky* 3/5 probe, which is a different situation from 15/15. The merged opus-5 cert (#614) already holds the all-required posture on this exact route, so this PR is consistent with the automation's precedent, not a departure from it.

Where you are right: `anthropic_ephemeral` attaches explicit `cache_control` markers, so on Anthropic routes this probe measures explicit caching, and the test's own docstring expects anthropic certs to declare `known_unsupported`. That is an inconsistency between the test docstring and the reconcile gate, and it applies to opus-5 today exactly as it applies here. It deserves its own PR that settles the rule and treats both certs identically. (For the record, the `IMPLICIT_PROMPT_CACHING` ratchet excludes `anthropic/*` names, so a demotion would not trip it; the only blocker is `cert_reconcile`.) I would keep it out of this PR.

tolokaforge_models/src/tolokaforge_models/certificates/registry.py
bberkes-toloka: Same situation as `IMPLICIT_PROMPT_CACHING` one thread up: 15/15 in `findings.json`, and `cert.py::reconcile` treats a `known_unsupported` capability with a >= 0.9 baseline as a FALSE-PESSIMISM violation (exit 1), so the demotion would fail the gate on this PR.

On the classification itself: the fable-5 comment calls this "pass-but-wrong-shape", and the merged opus-5 cert (#614) went the other way and kept it `required` on the same `reasoning_codec: anthropic` + OpenRouter route. So the repo already carries both postures for one behaviour; this PR follows the automation-era one. If the team wants "wrong-shape passes are ceilings" as the rule, that is a policy change that has to land on opus-5 and fable-5.1 together, plus a `cert_reconcile` exemption for it, and it belongs in its own PR rather than in one model's integration.

One detail worth having in the thread: this probe's turn 2 is mocked by design (it observes the outgoing `messages` kwarg), so on every model it proves the codec emits `reasoning_details` with the turn-1 text, not that the upstream consumes it. That is equally true for the certs that keep it `required` and the ones that do not.

bberkes-toloka:
Thanks for the pass, Ciro. I went through the four findings against the primary sources and answered each inline. Short version:

| Finding | Verdict |
|---|---|
| `cache_read` 0.25 | Correct. Anthropic prices Fable 5.1 cache hits at 0.025x input ($0.25 / MTok), a documented exception to the 0.1x rule. |
| `DISCRIMINATED_UNION` without `json_coerce` | Keep as generated. All five union probes are 15/15 natively; the policy is a recovery for a quirk this model does not show. |
| `IMPLICIT_PROMPT_CACHING` required | Keep. 15/15, and `cert_reconcile` refuses a `known_unsupported` cap with a >= 0.9 baseline (FALSE-PESSIMISM, exit 1). Real docstring-vs-gate inconsistency underneath, applies to opus-5 too, deserves its own PR. |
| `UNSIGNED_THINKING_REPLAY` required | Keep, same gate, same opus-5 precedent. |

One thing I verified on top, since a green replay probe proves nothing on its own: the 5/5 reprobe was live. `test_thinking_emits_blocks` mocks nothing (real completion, asserts `result.reasoning` is surfaced), the resolve log shows five separate pytest sessions at 17 to 26 s each, five times "1 passed", `re-probe findings: all_passed=True (capability_ran=True)`, and "Verified live against the loaded overlay: preset resolves to `anthropic_claude_fable_5_1`, adaptive/medium now emits {effort: high}". `cert_reconcile` is OK against both the reprobe and the observe baseline with 0 warnings.

Preset ordering also checked by re-running the resolver over the merged file: fable-5.1 -> `anthropic_claude_fable_5_1`, fable-5 and sonnet-5 -> `anthropic` (unchanged), opus-5 -> `anthropic_claude_opus_5`, opus-4.8 -> `anthropic_claude_4_8`. Nothing shadowed.

From my side this is mergeable as generated. The v1 board eval of this model is running independently on the v0.11.1 pin (TECHDEL-606) and is not affected by this PR either way.

Issue #23, opened 2026-06-01
Backfill foundational architecture ADRs


CiroGamboa:
After the ADR scaffolding lands (PR #11), backfill the foundational ADRs it defers: (1) adapter plugin model; (2) gRPC service decomposition; (3) single secret abstraction. MADR format under `docs/architecture/adr/`.

Internal tracking: TECHDEL-320.

CiroGamboa:
Closing: foundational ADRs backfilled — ADRs 0003 through 0018 all exist and are Accepted under `docs/architecture/adr/`. See [ROADMAP.md](https://github.com/Toloka/tolokaforge/blob/main/docs/architecture/ROADMAP.md) for the release ↔ ADR mapping.

PR #1458, opened 2026-09-02
fix(loop): empty-completion fail-loud termination + bounded API-error retry

### LINKED ISSUE #1443: grading/rubric_judge: Gemini rejects re-prompt after empty model turn; no API-error retry
## Summary

When the agent turn returns an empty completion (no `text`, no `tool_calls`), the loop appends the empty assistant message unchanged and the trial's tail message has `role=assistant, content=""`. On the next generation Google/Gemini rejects a request whose last turn is a model turn — the trial dies with an API error. There is no bounded API-error retry that could recover.

Field impact: 35 trials × 7 domains on the last v3 sample.

## Evidence

- `tolokaforge/core/loop.py:353-369` — `_run_turn` appends the empty assistant message via `_assistant_message(result)` at `:358` regardless of `result.text` / `result.tool_calls` being empty.
- `tolokaforge/core/loop.py:411-427` — `_advance_user_turn` no-ops when `tool_calls` is empty; the trailing message stays `role=assistant, content=""`.
- `tolokaforge/core/loop.py:175-237` — no API-error retry path; a generic API exception drops through `classify_loop_error` → `TerminationReason.API_ERROR` / `ERROR` termination.
- `tolokaforge/core/grading/judge.py:203-235` — the `_answer_terminating_submit_report` retry appends `role=tool` (Gemini-safe), but that retry fires only when `submit_report` was actually called; it does not cover the empty-turn case.

## Reproduction

Run a Gemini-hosted judge (or an agent) on any trial where the model produces an empty completion. The next request is submitted with the empty assistant message as the tail; Gemini's endpoint rejects the request and the trial terminates with `TerminationReason.API_ERROR`.

## Proposed fix

Two independent hardenings:

1. **On empty completion, do not append the empty assistant message.** Handle the empty case at the point of receipt — treat it as a distinguishable termination signal (or, if the loop must continue, insert a non-empty synthetic assistant message so the next request is Gemini-legal).
2. **Bounded API-error retry** for the class of provider errors that a trailing-model-turn produces (currently silent). One retry with a short backoff, then fail loud.

Either fix independently unblocks Gemini judges; both together are the fail-loud contract.

## Related

- #599 — parent umbrella for the widened fail-loud contract (this issue's sibling failure modes are captured there).

CiroGamboa:
## Summary

- Provider-side empty completions (`text == "" and not tool_calls`) terminate the trial with a new `TerminationReason.EMPTY_COMPLETION` before the empty assistant message can be appended to the trajectory tail. Gemini's endpoint rejects a request whose last turn is an empty model turn; the current behaviour was to append silently, re-invoke the provider, and die with an unclassified `API_ERROR`. Field impact: 35 trials × 7 domains on the last v3 sample.
- Loop-level bounded API-error retry — one retry with a 1.0 s backoff on `TerminationReason.API_ERROR` only. `RATE_LIMIT` / `API_TIMEOUT` / `TRIAL_LOST` / `EMPTY_COMPLETION` stay one-shot terminal.

## Design choices

- **Fail-loud termination on empty completion, not synthetic filler.** A synthetic non-empty assistant tail would fake a model turn; Gemini pattern-matches such fillers back into future turns (gotcha #23) and hides the model failure from downstream analysis.
- **`TrialStatus.FAILED` for the empty-completion path.** The harness worked; the model produced nothing. Keeps the trial in the graded MEASURED denominator — `ERROR` would attribute the trial to us as a harness bug.
- **Retry defaults `api_error_retries=1, api_error_backoff_s=1.0` hardcoded on `LoopConfig`.** Smallest non-zero values that give a transient provider hiccup one recovery window without noticeable unit-test slowdown. Not exposed via `run_config.yaml` / CLI — no compatibility surface change.
- **Retry class is `API_ERROR` only.** Every other classified reason owns a dedicated recovery path (rate-limit probe, transport-timeout retry, substrate re-registration, provider-shape fail-loud). Retrying them at the loop level would double-count and confuse the denominator.
- **Grader auto-fail branch expansion.** Every `TrialGrader` subclass (`RunnerRPCTrialGrader`, `JudgeBackedTrialGrader`, `GraderRPCTrialGrader`, `QueueTrialGrader`) gains an `EMPTY_COMPLETION` auto-fail mirroring the existing `STUCK_DETECTED` shape, so the `test_only_reasons_from_a_completed_trial_reach_grade_trial` reachability invariant continues to hold (justified drift from plan, discovered during Stage 1 implementation).

## Concepts introduced

**`TerminationReason.EMPTY_COMPLETION`** — a wire-shape signal the engine consumes directly. The rest of the engine remains behaviour-light on provider payloads; empty completions are the one exception because the next request must stay provider-legal. Failure-attribution groups the reason with `TIMEOUT / RATE_LIMIT / API_ERROR / ERROR` under `failure_class="timeout_or_resource"`, `deterministic=True`.

**`LoopConfig.api_error_retries` / `api_error_backoff_s` + `ToolCallingLoop.retry_sleep`** — a bounded loop-level retry for the narrow class of transient `API_ERROR` classifications, plus an injectable sleep seam parallel to `LLMClient._retry_sleep`. Attempt counter resets to 0 at the start of each outer-turn iteration. The `_attempt_turn` helper carries the retry state machine; the outer `run` dispatches on its return type (`TerminationDecision | tuple[...]`).

## Plan

See `/Users/cirogam/.claude/plans/toloka-tolokaforge/issue-1443-empty-completion-and-retry.md` — 2 stages, plan-critic reviewed round 1 REVISE (2🔴 + 1🟠 + 3🟡) → round 2 APPROVE WITH NOTES.

## Discovered issues

- Filed: none
- Fixed in this PR: grader auto-fail expansion for `EMPTY_COMPLETION` (Stage 1 justified drift), documented in commit body.

## Test plan

- `pytest tests/unit/test_tool_calling_loop.py tests/unit/test_failure_attribution.py tests/canonical/test_termination_reason_reachability.py` — 246 passed.
- `pytest tests/unit/test_orchestrator_logic.py tests/unit/test_runner_logic.py tests/unit/loop/ tests/unit/grading/ tests/unit/test_metrics.py` — 2160+ passed.
- Sharded reviewers: correctness APPROVE WITH NITS (1 🔵 `test_api_error_classification` sleeps 1 s on the retry — optional follow-up), hygiene APPROVE WITH NITS (1 §7a docstring fix applied at 65639559), type-fit APPROVE (1 🟡 `_attempt_turn` union return type — refactor-when-needed).

## Follow-up nits (not blocking)

- **`test_api_error_classification` 1 s sleep** (`tests/unit/test_runner_logic.py:765`). Fix: wire `retry_sleep` through `TrialRunner` and inject a no-op in `_make_runner`, or set `api_error_backoff_s=0.0` on the specific test's `LoopConfig`. Marginal unit-test runtime cost only; no behaviour change.
- **`_attempt_turn` return-type union** (`tolokaforge/core/loop.py:350-356`). The `TerminationDecision | tuple[...]` return has clean `isinstance` dispatch today; the tuple's `stop: bool` field encodes both "advance" and "stop", so the type doesn't name the semantic distinction. Refactor when either `_run_turn` or `_attempt_turn` grows a third return arm.

Closes #1443

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,13 @@ All notable changes to this project are documented in this file.
   additive on `Trajectory`; the failure-attribution table classifies it as
   `timeout_or_resource` (provider-side deterministic termination), and the
   runner-side graders auto-fail it without dispatching to `grade_trial`.
+- **loop**: transient `TerminationReason.API_ERROR` classifications are
+  retried once (bounded, hardcoded default via
+  `LoopConfig.api_error_retries=1` + `api_error_backoff_s=1.0`) before
+  terminating the trial. Rate limits, API timeouts, `TRIAL_LOST` and empty
+  completions remain one-shot. The retry budget resets at every outer
+  turn, and the `messages` list is unchanged on a failed attempt so the
+  retry attempts against the same prefix.
 
 ## v0.22.1 (2026-09-02)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 All notable changes to this project are documented in this file.
 
+## [Unreleased]
+
+### Fix
+
+- **loop**: empty completions (no text + no tool calls) now terminate with
+  `TerminationReason.EMPTY_COMPLETION` (fail-loud) instead of silently
+  appending an empty assistant message that Gemini and similar providers
+  reject on the next request as an API error. The termination reason is
+  additive on `Trajectory`; the failure-attribution table classifies it as
+  `timeout_or_resource` (provider-side deterministic termination), and the
+  runner-side graders auto-fail it without dispatching to `grade_trial`.
+
 ## v0.22.1 (2026-09-02)
 
 ## v0.22.0 (2026-09-02)
--- a/docs/LLM_LAYER.md
+++ b/docs/LLM_LAYER.md
@@ -1682,6 +1682,30 @@ as a compatibility surface — user overlay syntax and the
 `resolve_policy_names` fingerprint). Routing pinned by
 [`tests/canonical/test_message_assembly_filler_routing.py`](../tests/canonical/test_message_assembly_filler_routing.py).
 
+### Provider-side empty completion
+
+A generation that comes back with both `text == ""` and `tool_calls == []`
+is a *provider-side empty completion*: the request round-tripped and the
+provider chose to return nothing. `ToolCallingLoop._run_turn` recognises
+that shape immediately after `_generate` — before the assistant message
+would be appended — and terminates the trial with
+`TerminationReason.EMPTY_COMPLETION` and `TrialStatus.FAILED`. The empty
+assistant message is not appended, and the metrics sink still records the
+generation because the trial paid for the call.
+
+The distinction from `empty_assistant_filler` above is where the empty
+content lives. `empty_assistant_filler` handles empty **content the loop
+is about to send back to the provider on a tool-call turn** — Bedrock/Nova
+and Moonshot direct reject a request whose assistant turn has empty
+`content` alongside `tool_calls`, so those provider families opt in to a
+non-empty filler string. `EMPTY_COMPLETION` handles empty **content the
+provider produced**: appending it would send a request whose tail is a
+`role=model` turn with empty `content` and no `tool_calls` on the next
+iteration, and Gemini rejects that as an API error. The engine consumes
+this one wire-shape observation directly rather than routing it through
+`classify_loop_error` so post-run analysis can tell "the model produced
+nothing" apart from the API-error class that used to swallow it.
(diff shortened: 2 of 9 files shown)


CiroGamboa:
`test-smoke` reports one FAILURE: `tests/canonical/test_harness_registry_replay.py::test_replay_matches_baseline`. Error text explicitly identifies this as history-rewrite baseline drift ("history was rewritten (squash-merge, rebase), so the metric did not regress. Regenerate the baseline against the rewritten history. See #1234"). This branch does not touch `harness_registry`; the failure is upstream/pre-existing, tracked under #1234. Every other lane green (lint, hygiene-review, wheel-install-gate; test-full/gate SKIPPED as expected).

Issue #1463, opened 2026-09-02
grading: distinguish "cannot-score-on-this-substrate-by-design" from "should-have-scored-but-errored" in resolve_uncounted_fold


CiroGamboa:
## Context

#599 Stage 1 landed a fail-loud refusal in `resolve_uncounted_fold`
(`tolokaforge/core/grading/combine_weights.py`): any component the config
declared (in `combine.weights` with its matching config section) that arrives
at the composed verdict without a numeric score refuses the fold, marking
`FoldedGrade.refusal = True`. The runner's `GradeTrial` and the grader-service
dispatch both surface the refusal as `success = False` on the wire, so the
trial lands `UNGRADEABLE` rather than passing on a silently redistributed
weighted mean.

That rule catches the load-bearing runner-side defects the issue targeted:
a judge that errored (subtask b, landed in stage 1) and a golden replay whose
per-action execution errored (subtask a, landed in stage 2).

But it also fires for components the core substrate cannot score **by
design**, and there's currently no way to tell the two apart:

- `state_checks.db_probes` is `RUNNER_ONLY` — only the runner substrate
  executes it. On the core substrate the component reads as unscored, and
  the refusal fires.
- `llm_judge` needs a live judge call — a deterministic core aggregation
  cannot compose it. On the core substrate the component reads as unscored,
  and the refusal fires.

`lot_ops_01` and `helpdesk_01` (shipped in `examples/native/`) both declare
that shape. Before Stage 1, the core-side fold aggregated over `trace_checks`
alone. After Stage 1, both packs are `UNGRADEABLE` on the core substrate.
The fold has no way to distinguish "cannot score here by design, aggregate
the remainder" from "should have scored but errored, refuse".

## Proposal

Add a substrate-scoped "cannot-score" bit to `GradeComponentSpec`
(`tolokaforge/core/grading/grade_components.py`) or extend the fold's
signature so `resolve_uncounted_fold` can distinguish:

- **cannot score on this substrate by design** → aggregate the remainder,
  emit a note that the component was skipped
- **should have scored but errored** → refuse (current Stage 1 behaviour)

The runner substrate would still refuse on both cases (it can score both,
so an unscored one is always an error). The core substrate would fold the
remainder for RUNNER_ONLY + LLM-required components.

### Alternative shapes to consider

- **Substrate-scoped weights**: each component's weight in
  `combine.weights` becomes `{core: float, runner: float}`. Zero on core =
[...]

Issue #42, opened 2026-06-04
Decide whether the Release Gate should block PyPI publishing


CiroGamboa:
### Background

The release pipeline has two workflows that both trigger on a `v*` tag, **independently**:

- `publish-tolokaforge.yml` — builds and publishes to PyPI (trusted publishing, gated by the `release` environment) and creates the GitHub Release.
- `release-gate.yml` — re-runs the full test battery (lint, example validation, Docker smoke, unit, functional, integration, e2e) against the tagged commit.

Today they are **not coupled**: `publish` does not depend on `release-gate`. So the gate is a confidence signal / alarm, not a hard precondition. In practice a release can publish to PyPI even if the gate fails (this happened for 0.2.2 and 0.2.3, when the gate was broken).

### Question to analyze

Is it worth making the PyPI publish **conditional on the Release Gate passing**?

### Considerations

**For requiring it**
- A broken release never reaches PyPI; the artifact is validated end-to-end before it's public and irreversible (you can't truly un-publish a version).

**Against / risks**
- The gate includes Docker, integration, and e2e steps that can be slow or flaky; a flaky failure would block a legitimate release.
- Publishing is already gated by the `release` environment approval — a human already signs off.
- Coupling adds latency: the publish would wait for the full battery (potentially many minutes) on every release.

### Options to evaluate

1. **Keep them independent** (status quo) — gate is informational; rely on the `release` environment approval + green CI on `main`.
2. **Hard dependency** — `publish-pypi` `needs:` a passing gate (e.g. gate as a reusable workflow / job the publish waits on).
3. **Middle ground** — require only the fast, deterministic subset (lint + unit + functional + example validation) as a blocking pre-publish check; keep the slow Docker/integration/e2e steps as a non-blocking signal.

### Acceptance

A decision (with rationale) on coupling, and if we couple, an implementation that wires the chosen gate subset as a precondition for `publish-pypi` without making releases hostage to flaky slow tests.

_Context: surfaced while reviewing the release-automation PR (#41)._

PR #1470, opened 2026-09-02
fix(grading): widened fail-loud contract (judge-errored + golden-replay + auto-fail synthesis)

### LINKED ISSUE #599: fix(grading): golden-replay execution errors are graded as agent failures — surface them as system failures (fail loud)
## Context

When a component of the grade cannot be measured — the golden replay itself errors, the LLM judge errors, the agent's loop crashes, the trial times out or is stuck — the current engine paths either **silently redistribute the weight onto other components** or **synthesize a fabricated `state_checks: 0.0`** that reads to downstream consumers as if it were measured. In both shapes the trial reads as *the agent failed* when in fact **the component was not scored at all**. The grade / attribution / aggregates carry no structural signal that separates "agent failed" from "component could not run".

This is the fail-loud contract the judge's `JudgeStatus.ERRORED` already codified for one path. This issue widens that contract across every code path where a component-scoring surface can fail and produce a silently wrong number.

## Failure modes covered

Four concrete paths, all verified on `main` HEAD (v0.21.4+):

### (a) Golden-replay execution errors — original scope

- **Symptom:** `_execute_hash_grading` collects per-action exceptions into `golden_action_errors`, the golden hash is computed over whatever partial state the broken replay produced, `components.hash_score` lands at `0.0`, and the only trace is a string appended to `reasons` (`runner/service.py` ~1212-1214: `reasons += " | GOLDEN REPLAY ERRORS: ..."`).
- **Attribution impact:** `failure_attribution.json` labels these trials `failure_class: model_reasoning` (confidence 0.5), actively steering the reader away from the real cause.
- **Field report (2026-07-23, 6 trials, 0/6 binary pass at score 0.5, LLM judge 1.0 on every trial):** every `grade.yaml` carried identical deterministic golden errors; root cause was a tool-execution defect (a model→table registration bug in the adapter sidecar, fixed downstream). **Any** fixture regression, seed-data gap, tool refactor, or environment drift that breaks golden replay silently converts into agent failures across every trial of the task — deterministically, so it poisons pass rates and A/B comparisons until someone reads `reasons` by hand.

### (b) Judge-errored component silently redistributes weight

- **Symptom:** On `JudgeStatus.ERRORED`, `components.llm_judge_score` never assigns and stays at the `-1.0` sentinel (`runner/service.py:1900-1913`). `combine_grade_components` skips `score < 0` in its `active_components` gate (`runner/grading.py:594-696`, specifically the gate at `:649-663`), so the judge component disappears from the fold. `resolve_uncounted_fold` in `core/grading/combine_weights.py:88-124` returns `None` when other scored components carry non-zero weight, so `combine_by_method` computes `weighted_mean = Σ(score_i · w_i) / Σ(w_i)` over the *remaining* components only.
- **What this looks like on a real trial:** the judge errored (surfaced only in `reasons`), and the trial's grade is derived as if the judge's weight had been reallocated. An already-passing state_checks or transcript_rules component absorbs the missing judge share and the trial passes threshold with no judge bar at all.
- **Recent field evidence:** 6 known unearned passes across the v3 sample; the failure is invisible to any aggregate that reads `score` without also reading `reasons`.

### (c) Empty completion → agent loop crashes → HARNESS_ERROR in the denominator

- **Symptom:** The model returns an empty completion (no `text`, no `tool_calls`); `_run_turn` at `core/loop.py:353-369` appends the empty assistant message via `_assistant_message(result)` at `:358` regardless of emptiness. Downstream `_advance_user_turn` at `:411-427` calls `self.user_turn(messages)` at `:419` with no `try/except`; the user simulator crashes on the empty-model tail. Outer catch at `:315-324` routes through `classify_error(exc)` at `:232-236` → `TerminationReason.ERROR` / `TrialStatus.ERROR`.
- **Attribution outcome:** `failure_attribution.py:104-105` maps `TerminationReason.ERROR` (and `TRIAL_LOST`) to `TrialOutcomeClass.HARNESS_ERROR`. The docstring at `failure_attribution.py:37-44` says `HARNESS_ERROR` is "Counted in the rates like any other failure" — so operationally the trial lands in the denominator alongside real measured failures.
- **Field impact:** 23 trials in one v3 domain (6.58%), all bookkept as failures even though they died at turn 1-5 before the task began.

### (d) Synthesized `state_checks: 0.0` on ERROR / TIMEOUT / STUCK

- **Symptom:** `core/trial_grader.py:259-270` — on `TrialStatus.ERROR` or `TIMEOUT` — synthesizes `Grade(binary_pass=False, score=0.0, components=GradeComponents(state_checks=0.0), reasons=f"Trial failed with status: {trajectory.status.value}")`. `:273-284` — on `TerminationReason.STUCK_DETECTED` — synthesizes the same shape with `reasons="Agent got stuck (repeated actions without progress)"`. The pattern is duplicated in every runtime path: `RunnerRPCTrialGrader` at `:604-636`, `QueueTrialGrader` at `:686-732`, `JudgeBackedTrialGrader` at `:720-732`.
- **What this hides:** the synthesized components carry no marker distinguishing them from a real measured `0.0`. Downstream analytics that read `components.state_checks` see 40+ trials with what looks like a measured zero when in fact the harness auto-failed and back-filled the field.
- **Interaction with `zero_coverage` (M#38):** `orchestrator.py:479-484` fires `zero_coverage` only on `total_attempts > 0 and measured_trials == 0`. A synthesized `Grade` is not `None`, so the trial counts as measured — the fabricated `0.0` defeats the coverage gate that was built to catch this exact case.

## Deliverables — the widened fail-loud contract

- [ ] **(a) Golden replay:** when `golden_action_errors` is non-empty, the `state_checks` component is marked **not scoreable** rather than scored `0.0` — mirroring `JudgeStatus.ERRORED`. The trial must still never PASS on a broken oracle.
- [ ] **(b) Judge ERRORED:** on `JudgeStatus.ERRORED`, the grade must not silently derive a `score` from the remaining components. Either fail the trial fail-loud, or refuse to compose the score and mark the trial as ungradeable. Never a silently redistributed weighted mean.
- [ ] **(c) Empty-completion crash:** distinguish "the trial's agent loop crashed on empty completion" from ordinary `HARNESS_ERROR`. A new termination_reason (e.g. `MODEL_EMPTY_COMPLETION`) that either routes to `INFRASTRUCTURE_ABORT` (out of the denominator) or is at minimum flagged separately in the failure_attribution surface.
- [ ] **(d) Auto-fail synthesis:** synthesized components on ERROR/TIMEOUT/STUCK must carry a marker distinguishing them from a real measured zero — either a `synthesized_by_termination_reason` field on `GradeComponents`, or the components dict omits the fields that were not actually measured (whichever composes cleanly with the current `component_score` unions).
- [ ] **failure_attribution.json:** the four modes above surface as their own attribution classes — never `model_reasoning` when the model didn't get to reason.
- [ ] **Run aggregates:** count these separately from ordinary agent failures. Retries against a broken oracle, an errored judge, or a crashed simulator will replay the same failure — infra retry policy should reflect that.

## Design considerations

- Each of the four failure modes is deterministic per task (or per model, in (c)) — a loud signal at run level would catch the condition after the first trial instead of after the spend.
- The existing human-readable `reasons` strings should stay (they're what made the field diagnoses possible), but as the human-readable trace, not the machine contract. Downstream tooling (benchmark-analyzer's `golden_replay_broken` bucket string-matches `reasons`) should read a structural field instead.
- Cascading fix: (d) interacts with M#38's `zero_coverage` field — the marker on synthesized components should let `zero_coverage` fire on a run where every measured trial was actually auto-failed.

## Related

- #449 / #468 — the judge-side fail-loud precedent (ERRORED excludes the component; never a silently wrong grade). Same principle, widened here to golden replay, judge composition, agent-loop crash, and auto-fail synthesis.
- #907 / #816 / #815 — hash-contract family (parallel work on state-hash correctness).
- #1398 — silent-component-drop pattern in `composite_dispatch.audit_accounted_keys`.
- #583 — the adapter grading seam under which the field cases exist.
- #217 — the `_execute_hash_grading` live golden-replay path.

CiroGamboa:
## Summary

Closes #599 by shipping the four subtasks the issue named as one widened fail-loud contract:

- **(a) Golden-replay execution errors** — `HashGradingResult.hash_unscorable` refuses to fabricate `hash_score=0.0` when the replay errored; Stage 1's fold refusal propagates the trial to UNGRADEABLE.
- **(b) Judge-errored silent redistribution** — `resolve_uncounted_fold` refuses when any declared component produced no verdict; `FoldedGrade.refusal` flags it; runner and grader-service both surface `success=False` on the wire.
- **(c) Empty-completion crash routing** — verified fully closed by #1443. No residual work.
- **(d) Auto-fail synthesis marker** — `Grade.synthesized_by_termination_reason` names the reason the harness synthesised the grade; `GradeComponents()` all-None replaces the fabricated `state_checks: 0.0` / `llm_judge: 0.0` shapes; `failure_attribution` routes synth grades whose reason isn't otherwise enumerated to `harness_autofail`; `GradingCompleteness.zero_coverage` widens to fire on runs whose every measured trial was auto-failed.

## Design choices

- **`FoldedGrade.refusal: bool` discriminator, not substring on `reason`.** Substring-match on natural-language prose is fragile; the boolean flag is a clean wire-level discriminator both substrates read.
- **Stage 2 mechanism (b): additive `hash_unscorable` property + caller guard.** Chose over fail-loud property raise (mechanism a) to preserve the many existing `_reasons_for` test helpers that read `hash_score` on broken-replay results without needing exception handling.
- **Grader auto-fail expansion in every `TrialGrader` subclass** — mirrors the STUCK_DETECTED pattern so no synth grade falls through to `GradeTrial`, preserving `test_only_reasons_from_a_completed_trial_reach_grade_trial`.
- **`harness_autofail` scope: past the deterministic elif chain.** TIMEOUT / EMPTY_COMPLETION / ERROR keep their `timeout_or_resource` label (their existing branch catches them first); today only STUCK_DETECTED falls through. Forward-compat: any new `TerminationReason` a `TrialGrader` synthesises from without an enumerated branch lands here.
- **`FailureRecord` gains two additive fields.** `extra="forbid"` required the model edit for round-trip; additive-with-default keeps pre-Stage-3 bundle reads compatible.
- **`synthesized_trials` filters to `TrialOutcomeClass.MEASURED`** — corrects a mixed-run gap (round of reviewer correctness feedback) so `zero_coverage` preserves ADR-0041's stated intent when synth trials mix HARNESS_ERROR with MEASURED outcomes.
- **`TrialStatus` + `TerminationReason` extracted to `tolokaforge/core/models/trial_status.py`** — breaks the `grade.py` ↔ `trajectory.py` import cycle without forward references or `model_rebuild()` fragility. All existing importers preserved via re-export.

## Concepts introduced

**`FoldedGrade.refusal`** — a boolean discriminator on the fold's return value that both wire substrates (runner `GradeTrial` + grader-service dispatch) read to emit `success=False` / `GradingFailedError`. Replaces the previous "no way to structurally distinguish 'the fold declined to compose' from 'the fold composed a legitimate zero'".

**`Grade.synthesized_by_termination_reason`** — a `TerminationReason | None` marker on the Pydantic `Grade` model naming why the harness auto-failed the trial. `None` on every real evaluator grade. Downstream analytics can now separate an auto-failed trial from one that measured `state_checks: 0.0`.

**`failure_class="harness_autofail"`** — a new deterministic attribution class in `failure_attribution.py`. Fires on any synth grade whose termination reason falls past the deterministic elif chain (today only STUCK_DETECTED). Prevents the tool-log scan default `model_reasoning` from misattributing a harness-terminated trial to the model.

**Widened `GradingCompleteness.zero_coverage`** — fires when `synthesized_trials == measured_trials` on a run that had trials. Composes with M#38's `--fail-on-zero-coverage` CLI flag so a run whose every measured trial was auto-failed exits non-zero. ADR-0041 formula and `docs/CLI.md` exit-code prose both updated.

## Plan

See `/Users/cirogam/.claude/plans/toloka-tolokaforge/issue-599-fail-loud-widened.md` — 3 stages (fits persistent mode). Plan-critic reviewed round 1 REVISE → round 2 REVISE → round 3 APPROVE (2 🔴 architect-fixed, 2 🟠 architect-fixed, 3 🟡 architect-fixed + 1 🟠 rebutted-with-evidence on `docs/OUTPUT_FORMAT.md` vs `docs/ANALYTICS.md` as SSOT for `failure_attribution.json` schema).

## Commits

- \`33d73fc2\` Stage 1 — declared-but-unscored fold refusal
- \`6efd5995\` Stage 2 — golden-replay `hash_unscorable`
- \`05d6abb2\` Stage-1-followup — core-substrate corpus sweep (22 canonical tests + module docstring rewrite)
- \`b197e8a6\` Stage 3 — harness-autofail synthesis marker + widened `zero_coverage` + `FailureRecord` model edit + `TrialStatus`/`TerminationReason` module extraction
- \`58f00e0a\` review(hygiene): drop stage narration from comments/docstrings + de-duplicate TrialStatus/TerminationReason exports
- \`ff486afa\` review(correctness): synthesized_trials filters to MEASURED outcomes only

## Discovered issues

- Filed: [#1463](https://github.com/Toloka/tolokaforge/issues/1463) — substrate-aware distinction ("cannot-score-on-this-substrate-by-design" vs "should-have-scored-but-errored") in `resolve_uncounted_fold`. Deferred design work; not milestone #42.
- Fixed in this PR: 22 canonical corpus tests in `test_example_pack_grading_corpus.py` (Stage-1-followup), 1 legacy-locking test in `test_runner_pipeline.py` (Stage 2 sweep), 6 legacy-locking tests in `test_combine_weights.py` / `test_grading_ledger.py` / `test_grading_substrate_parity.py` / `test_transcript_substrate_parity.py` / `test_grader_parity_reference.py` (Stage 1 sweep).

## Test plan

- Unit lane: 7785 passed (excluding pre-existing `test_persistent_shell_local.py` macOS-only failure and pre-existing `test_ungradeable_trial_accounting.py` hang class).
- Canonical lane: 1801 passed (excluding pre-existing `test_harness_registry_replay.py::test_replay_matches_baseline` history-rewrite failure filed as #1234).
- Post-review commits: 332+336 passed on targeted files.
- Sharded reviewers: correctness APPROVE WITH NOTES (1 🟡 fixed in `ff486afa`), hygiene REQUEST CHANGES (4 Blocker 8 + 1 Major + 2 Minors fixed in `58f00e0a`), type-fit APPROVE (no findings).

Closes #599


CiroGamboa:
CI note: `test-smoke` failure is [#1234](https://github.com/Toloka/tolokaforge/issues/1234) (harness-registry replay baseline is self-referential and flips on any history rewrite). Single failing test is `tests/canonical/test_harness_registry_replay.py::test_replay_matches_baseline` with the exact history-rewrite marker in the assertion. Pre-existing; not caused by this branch. Other lanes green (lint, hygiene-review, wheel-install-gate). Proceeding with squash-merge per milestone-mode standing permissions.

Issue #1472, opened 2026-09-03
grading/state_checks: task-yaml hint for tables whose row order is/isn't hash contract (option 1 from #1444)


CiroGamboa:
Split off from #1444 (issue-family "hash verdict signature").

## Background

The runner-substrate hash `compute_stable_hash`
(`tolokaforge/core/hash.py:270-360`) serialises state via
`json.dumps(sort_keys=True)`, which sorts dict keys but preserves list
element order. So rows within a table are graded as if their creation
order were part of the contract — two tables with the same rows in
different insertion order hash differently.

For most tables that is the wrong semantics. Row order in a DB table is
typically not part of what the agent produces; it is an artefact of
insertion order under the tools the agent happened to call. Grading a
trial as a hash-mismatch because a `create_order(...)` call landed
before a `create_appointment(...)` call in one trial and after it in
the golden replay is a false-negative for tasks whose contract is
"these rows exist", not "these rows exist in this order".

For a small class of tables — audit logs, transcript-like append-only
records — row order IS part of the contract, and folding it away would
false-pass.

The parity fix in #1444 makes the runner-side diff detect same-set-
different-order and report it as a distinct verdict, so the operator
never sees "hash says mismatch, diff says match" again. That closes the
contradictory-verdict-signature class, but the false-negative it
diagnoses remains — the hash still reports mismatch on a legitimate
reordering.

## What to build

Introduce a per-table hint in `StateChecksConfig`
(`tolokaforge/core/models/task_config.py:485-518`) that names the
tables whose row order is part of the contract. All other tables become
order-blind at hash time.

Sketch (open to bikeshed):

```yaml
state_checks:
  # Per-table opt-in — row order matters for these; the default is
  # unordered, matching what most DB-shaped tables actually mean.
  ordered_tables:
    - audit_log
    - dispatched_events
```

Semantics:

- On both `compute_stable_hash` and `state_digest` (`state_checks.py`),
  a table absent from `ordered_tables` has its list sorted canonically
  before hashing. A table in the list is hashed verbatim.
- `numeric_string_fields` and `id_fields` are unchanged.

## Compatibility surface

`state_checks.numeric_string_fields` / `id_fields` are the precedent —
per-table hints in `grading.yaml`. Adding `ordered_tables` extends the
same surface; the CHANGELOG + `docs/CONFIG.md` + `docs/GRADING.md`
must all migrate together. Existing tasks with reproducible row order
[...]

Issue #907, opened 2026-08-06
state_checks.hash.golden_actions elements are untyped — a `kwrags` typo replays with no arguments and produces a wrong expected hash


azorej:
## Summary

`state_checks.hash.golden_actions` is untyped on the core authoring surface. A misspelled key
**inside one action** is silently ignored, and the replay executes that action with whatever
survived. For `kwargs` that means the tool is called with **no arguments** — so the golden world is
built wrong and the expected hash the trial is compared against is a *wrong value*, not a missing
one.

Measured at `feat/trace-checks-tail` (`19e2a379`), driving `validate_grading_yaml` with a
**resolvable** tool inventory declaring `place_order`:

| golden action element | authoring gate |
|---|---|
| `{name: place_order, kwargs: {...}}` | passes (correct) |
| `{nmae: place_order, kwargs: {...}}` | **refused** — `state_checks.hash.golden_actions[0].name: this golden action names no tool to replay…` |
| `{name: place_order, kwrags: {...}}` | **passes silently** |
| `{name: place_order, kwargs: {...}, junk: 2}` | **passes silently** |

The `name` typo is caught (by the golden-action name rule, which needs a resolvable inventory —
with `ToolInventory.unresolvable()` it is skipped into `unchecked` and *nothing* is caught). The
`kwargs` typo is caught by nothing on either surface.

This is a worse failure shape than #730: #730 turns a check off, this one leaves the check on and
compares against a state the golden replay built incorrectly.

## The two substrates hold different element shapes in this one field

This is the larger half of the ticket, and it is why the fix is not a one-line annotation.

- The **author** writes `{name, kwargs}`.
- `GoldenAction` (`tolokaforge/runner/models.py:281-291`, `extra="forbid"`) declares
  **`tool_name` / `arguments`** — *not* `name` / `kwargs`.
- `NativeAdapter.to_task_description` translates between them at `adapters/native.py:602-605`
  (`tool_name=golden_action.get("name", "")`, `arguments=golden_action.get("kwargs", {})`).
- `RunnerStateChecksConfig.golden_actions` is `list[GoldenAction]` (`:338`), and
  `_authored_hash_block()` (`:392-403`) hands that list to the shared composition rules under the
  author-facing key `golden_actions` — so the dict it calls "the author-facing `state_checks.hash`
  block" already carries non-author-shaped elements for this one key.

Nothing notices today because every shared rule reads the list for **truthiness only**
[...]

Issue #816, opened 2026-08-03
A hash verdict computed against a partially replayed golden world is admissible, and nothing decides whether it should be


azorej:
## Summary

#734 makes an **unresolvable** golden action a replay error, and makes a golden action that
**raised** name itself in `grade.reasons`. What it deliberately does not settle: whether the
hash verdict computed against that partial world should be admitted as a verdict at all.

After #734, a pack whose second golden action fails its preconditions still produces a binary
`1.0` / `0.0` hash component, folded at its declared weight, alongside a reason saying one of two
actions did not run. The verdict is now *annotated* rather than silent — but a trial that
happened to skip the same action still scores `1.0`, and one that did the right thing still
scores `0.0`.

## Measured, on `tests/data/tasks/shop_orders_02`

Golden path is `place_order` then `confirm_payment`. Point the second action at a non-existent
order so `invoke()` raises, then grade a trial that placed the order and never paid — the wrong
behaviour:

```
WRONG trial (placed, never paid) vs a golden replay whose payment raised
  -> hash score 1.0, reason "State hash matches"
```

#734 adds "1 of 2 golden actions did not run: confirm_payment (…)" to that reason. The `1.0`
stays.

## Why it was left alone

The in-code defence is a real compatibility argument, quoted at
`tolokaforge/core/grading/state_checks.py:389-392`: tau-bench continues past a golden action
whose preconditions are not met, and golden-action hash grading is the tau-style path. Changing
the *verdict* — rather than the reason — moves scores on exactly the packs most likely to be used
as reference data, which is the measurement pass #734 declines to fold in.

## What needs deciding

- [ ] Is a hash verdict over a partial world admissible, or should an incomplete replay produce
      **no** hash verdict — the answer #733 gave every other "configured but could not be
      evaluated" state source (`state_checks` is `None`, not `1.0`)?
- [ ] If it stays admissible, does an author get a per-pack say (a `hash.require_complete_replay`
      flag), or is it fixed by the adapter the pack loads under?
- [ ] The measurement: how many corpus trials are graded against a partial world today, and how
      many verdicts move under "no verdict on an incomplete replay"

## Family

Adjacent to #742 (the hash verdict's binariness is guarded by a source audit, and one producer
has no behavioural test) — both ask what a hash verdict is entitled to claim. Same family as
#734, #729, #733, #730, #731: "the grade is computed and looks fine, but it did not measure what
it claims to".

azorej:
[...]

Issue #815, opened 2026-08-03
Golden-action names resolve against two different namespaces, and the runner suffix-matches where core does not


azorej:
## Summary

A golden action's `name` is resolved to a callable by two substrates against two different
namespaces, under two different rules. Found while planning #734, which makes an
unresolvable name loud on both sides — that turns this latent divergence into "one substrate
raises, the other grades".

**Core** (`tolokaforge/core/grading/state_checks.py:378`) resolves against the pack's MCP-server
`TOOLS` map, **exact match only**:

```python
if action_name not in tools_map:
    ...
tool_class = tools_map[action_name]
```

**Runner** (`tolokaforge/runner/service.py:2229-2238`) resolves against `trial_context.agent_tools`,
**exact, then a single-underscore suffix rule**:

```python
tool = trial_context.agent_tools.get(tool_name)
if tool is None:
    # Try with domain prefix (golden actions use unprefixed names)
    for registered_name in trial_context.agent_tools:
        if registered_name.endswith(f"_{tool_name}"):
            tool = trial_context.agent_tools[registered_name]
            break
```

## Why it matters

A pack whose tools are registered domain-prefixed (`ots_19_zendesk_update_item`) and whose
golden actions are written unprefixed (`update_item`) resolves **runner-side** and does **not**
resolve core-side. Today that is silent on both: the runner replays it, core skips it and
returns a partially replayed world as the golden state (#734). Once #734 makes an unresolvable
name a replay error, the same pack becomes a grading error core-side and a normal verdict
runner-side — a substrate divergence on a shape that is plausible in the tau/OTS corpus.

The suffix rule is also ambiguous by construction: two registered names ending in `_update_item`
resolve to whichever the dict iteration reaches first.

## Measured in-repo

Four packs declare `state_checks.hash.golden_actions`. For the two with an MCP server, the
`TOOLS` keys and `tools.agent.enabled` are **identical sets**, and every golden action name is
in both, so no in-repo pack exercises the suffix path:

| pack | golden action names | resolves |
|---|---|---|
| `tests/data/tasks/shop_orders_02` | `place_order`, `confirm_payment` | exact, both substrates |
| `tests/data/projects/food_delivery_2/tasks/order_modify_with_checks` | `get_user_details`, `get_order_details`, `modify_order` | exact, both substrates |
| `tests/data/projects/food_delivery_2/tasks/order_six_items_golden` | `get_user_details`, `create_order` ×3 | exact, both substrates |
[...]

PR #1473, opened 2026-09-03
fix(grading): runner-side diff reports order-mismatch verdict (#1444)

### LINKED ISSUE #1444: grading/state_checks: state-hash preserves list order, diff is order-blind; contradictory verdict signature
## Summary

The state-hash verdict preserves list element order (rows within a table are graded as if their creation order were part of the contract), while the explanatory diff is order-blind (uses `set()` difference to compare rows). Two tables with the same rows in different orders reach an "unbelievable" verdict shape: the hash flags a mismatch, the diff prints "States match", the grade component records `hash_score: 0.0` next to a diff the operator cannot reconcile.

Field impact: 4 direct trials + 11 cascades on the last v3 sample.

## Evidence

Hash (order-sensitive on lists):
- `tolokaforge/core/grading/state_checks.py:63-67` — `to_hashable` list branch: `tuple(to_hashable(el) for el in item)` preserves order.
- `tolokaforge/core/hash.py:186-190` — `_convert_datetime_to_str` list branch preserves order.
- `tolokaforge/core/hash.py:346` — `json.dumps(..., sort_keys=True)` sorts dict keys but not list elements.

Diff (order-blind on rows):
- `tolokaforge/runner/grading.py:127-136` — `_compare_table_records` builds `record_to_tuple` per row, then `trial_set = set(trial_tuples.keys())` and uses set difference for missing/extra — order of rows within a table is discarded.
- `tolokaforge/runner/grading.py:87-91` — returns `"States match"` when set-diff is empty.

## Reproduction

Any task where the golden and the trial produce the same rows in different insertion order. `check_hash` (`state_checks.py:456-465`) or the runner-side JSON hash fails; `compute_state_diff` reports "States match".

## Proposed fix

Two options; pick one and enforce parity:

1. **Order-blind hash on tables:** sort list elements canonically before hashing where the table shape says row order is not part of the contract. Requires a per-table hint about whether row order matters (rare — but exists for e.g. audit logs).
2. **Order-sensitive diff:** teach `_compare_table_records` to detect same-set-different-order and print an order-mismatch line rather than "States match".

Whichever ships, the property to lock in a test: hash-verdict-parity — on any pair of `(trial_state, golden_state)`, the diff's verdict string must be consistent with the hash's boolean outcome.

## Related

- #907 — golden_actions hash sensitivity to argument shape.
- #816 — hash verdict admissibility under partial golden replay.
- #815 — golden-action namespace / suffix-match hash comparison mismatch.

CiroGamboa:
## Summary

- Locks the **hash-verdict-parity** invariant on the runner-side pair `compute_stable_hash` + `compute_state_diff`: when the two hashes disagree, `state_diff.identical` MUST be `False` and its `summary` MUST NOT be `"States match"`.
- Field impact: 4 direct trials + 11 cascades on the last v3 sample no longer produce a `hash_score: 0.0` verdict next to a `state_diff_json` that reads as identical.

## Design choices

- **Option 2 (order-sensitive diff)**, not Option 1 (order-blind hash). Preserves current hash semantics (row order remains part of the contract by default); no compatibility surface change on `compute_stable_hash`, which is persisted in db-service ETags, snapshot hashes, and the wire (`ResetTrialResponse.state_hash`, `GetStateResponse.stable_hash`). Option 1's per-table task-yaml hint mechanism is deferred to #1472.
- **Crisp order-mismatch trigger**: `TableDiff.order_mismatch=True` fires ONLY when the set-based `missing / extra / different` are all empty AND ordered `record_to_tuple` sequences differ. Combined shapes impossible by construction; the per-table summary vocabulary stays finite (three shapes: absent, `"<t>: rows in wrong order"`, `"<t>: N missing, M extra, K different"`).
- **`TableDiff.order_mismatch: bool = False` additive-with-default**. Wire is JSON via `Grade.state_diff_json`; consumers `json.loads` into dicts (`core/trial_grader.py`, `grader/client.py`, `grader/service.py`, `core/shared_stack_runtime.py`) — older readers deserialise the extra key transparently. Only one `TableDiff(` producer in the tree.
- **`StateDiff.identical` derivation ORs `order_mismatch`** into the existing non-emptiness check, so an order-only-mismatched table flips the property to `False` — the read-side of the invariant.
- **Hypothesis dep landed with Stage 1**, not deferred to Stage 2. `hypothesis>=6.100,<7` in `[dependency-groups] dev`; scoped to the same commit as the property test.

## Concepts introduced

**`TableDiff.order_mismatch: bool`** — a per-table wire-level discriminator on the runner-side `StateDiff` that flags the same-set-different-order equivalence class. Fires exclusively when the set-based classification produced no offence but the ordered `record_to_tuple` sequences of trial vs golden rows differ. Read by `StateDiff.identical` (the aggregate parity signal) and by the per-table summary render.

**Hash-verdict-parity invariant** — the property `compute_stable_hash(trial) != compute_stable_hash(golden) ⇒ not compute_state_diff(trial, golden).identical`. Locked at the unit tier via a parametrised equivalence-class case set + a hypothesis property. The invariant applies to the runner substrate pair only; the core substrate pair (`state_digest` + `calculate_state_diff`) is already consistent by construction.

## Plan

See `~/.claude/plans/toloka-tolokaforge/issue-1444-hash-diff-verdict-parity.md` — 2 stages (fits persistent mode). Plan-critic reviewed round 1 REVISE (1 🟠 trigger-vs-illustration + 3 🟡 grounding/no-op-stage/dep-availability) → round 2 APPROVE.

## Discovered issues

- **Filed**: [#1472](https://github.com/Toloka/tolokaforge/issues/1472) — task-yaml hint for tables whose row order is/isn't hash contract (Option 1 from #1444). Compatibility surface; deferred design work.
- **Fixed in this PR**: none. Bug is narrow, one well-defined seam. Related open issues [#907](https://github.com/Toloka/tolokaforge/issues/907), [#816](https://github.com/Toloka/tolokaforge/issues/816), [#815](https://github.com/Toloka/tolokaforge/issues/815) are hash-family siblings; cross-linked here but not solved.

## Test plan

- **Behaviour lock**: `tests/unit/grading/test_hash_verdict_parity.py` — 5 parametrised equivalence-class cases + one hypothesis property. Pinned reproduction fixtures (`same_rows_different_order`, `genuinely_different_rows`, `empty_trial_nonempty_golden`, `nonempty_trial_empty_golden`, `identical`). Property shrinks to a 2-row swap counterexample.
- **Unit lane**: `mcp__dev__run_tests marker=unit path=tests/unit/grading/` — 1964 passed post-Stage-2.
- **Canonical**: `tests/canonical/test_grader_service_contract.py` + `test_grading_substrate_dispatch.py` — 10 passed; empty `state_diff_json` in parity baselines means no snapshot regeneration needed.
- **Sharded reviewers**: correctness APPROVE, hygiene APPROVE-after-nits (2 §7a docstring current-state rewrites, applied inline in `87fb1339`), type-fit APPROVE.

## Follow-up nits (not blocking)

- **`.hypothesis` `norecursedirs` warning** — hypothesis's pytest plugin emits `UserWarning: Skipping collection of '.hypothesis' directory` because `pyproject.toml` `norecursedirs` replaces rather than extends pytest defaults. One-line fix (`norecursedirs` list gains `".hypothesis"`); optional, not in scope.

Closes #1444


CiroGamboa:
CI note: all 4 `test-smoke` failures are pre-existing on `main` at the same SHA (`a9207597`), verified via run [33745482952](https://github.com/Toloka/tolokaforge/actions/runs/33745482952) — none caused by this branch:
- `test_certify_suite_collection_snapshot` — snapshot on `main` is out of sync with the current test collection.
- `test_harness_registry_replay.py::test_replay_matches_baseline` — the known #1234 history-rewrite false positive.
- `test_models_wheel_replay.py::test_replay_matches_baseline` — same history-rewrite class as #1234.
- `test_run_trial_cli_subprocess.py::test_happy_path_matches_run_trial_library` — `No user-simulator model configured` on the bundled example.

Branch's own hash-verdict-parity tests (6 tests: 5 parametric + 1 hypothesis property) pass locally on the branch. Lint/hygiene-review/wheel-install-gate green. Proceeding with admin-merge on the pre-existing red per milestone-mode standing permissions.

PR #1532, opened 2026-09-07
feat(automation): page the Slack requester alone for model integrations

balazs-toloka:
## What

The model auto-integration Slack notifier now pages the **requester alone** when the run came in through the channel, instead of always pinging the standing `ARENA_AUTOMATION_SLACK_MENTIONS` list. This mirrors the eval orchestrator's `resolve_mentions`, which already pages the eval requester the same way.

A channel where every integration pages the same standing list is a channel people mute, and a muted ping defeats the one message that needs one. The standing list stays as the fallback for runs nobody asked for through Slack (a by-hand label add or a manual `workflow_dispatch` has a GitHub actor but no Slack identity).

## How

- **`tools/automation/src/automation/slack.py`** — add `resolve_mentions()`, `REQUESTED_BY_ENV` (`SLACK_REQUESTED_BY`, the same variable name the eval notifier uses), `looks_like_slack_user_id()`, and the `_SLACK_USER_ID_RE` guard. `cmd_reply(..., mention=True)` now pings `resolve_mentions()` rather than the raw `SLACK_MENTIONS`.
- **`.github/workflows/integrate-model.yml`** — new `requested_by` input; `SLACK_REQUESTED_BY` job env from it; a "Resolve the Slack requester" step that fills it from the branch's seed-commit `requested_by:` trailer on the label-triggered path (where inputs are absent).
- **`.github/workflows/slack-integrate.yml`** — the poller forwards `-f requested_by="$REQUESTER"` (it already holds the id for the allowlist check).
- **`docs/AUTO_INTEGRATION.md`** — a _Who gets pinged_ section.

The requester value crosses a repo boundary as a workflow input and comes back OUT as a mention, so `resolve_mentions` re-validates it against the Slack user-id shape (`^[UW][A-Z0-9]{4,}$`): a by-hand seed that named a GitHub login rather than a Slack id is dropped and the standing list takes over. Unset on both is not an error, the message still posts without a ping.

## Tests

`tools/automation/tests/test_slack.py` gains `TestResolveMentions`, `TestLooksLikeSlackUserId`, and `TestTheReplyPingsTheRequesterAlone` (requester wins, junk is dropped not rendered, `cmd_reply` pings the resolved audience). Local runs:

```
tools/automation/tests            304 passed
tests/canonical/test_integrate_model_bucket_flow.py + test_icons.py + test_gateway_catalog.py   108 passed
ruff format + ruff check          clean
```

No behavior change for a repo with no requester in scope (empty input, no trailer): it still pages `SLACK_MENTIONS` exactly as before.

Companion skill update (tasks repo): toloka-partners/tolokaforge-tasks#315.

---

### CI note

`test-smoke` fails on one pre-existing test, `tests/canonical/test_models_wheel_replay.py::test_replay_matches_baseline` — the self-referential snapshot flake tracked in #1234. It is unrelated to this change: the branch adds **zero** `integrate:` commits (the only thing that test enumerates), it passes locally on this branch and on `origin/main` alike, and it is the same failure the recent integration PRs (#1449, #1161, #1277) merged past and #1475 later regenerated. Regenerating the snapshot here would be scope creep against a moving target. Every other check (lint, hygiene-review, CodeQL, Analyze, wheel-install-gate) is green.

### Review addressed (independent review, 2026-09-07)

- **Precedence-independent requester resolution** (`fix` commit): dropped the job-level `SLACK_REQUESTED_BY` default so a job-`env:` entry can neither shadow nor be shadowed by the resolver step's `$GITHUB_ENV` write. The step now reads `inputs.requested_by` at step scope and is the single writer, so the label-triggered fallback pages the requester as intended regardless of runner env precedence. The Slack-dispatch path was already precedence-safe.
- **`looks_like_slack_user_id` now has a production caller**: `resolve_mentions` validates through it instead of an inline regex.
- Reviewer confirmed: the resolver step runs before every `--mention` call; the `^[UW][A-Z0-9]{4,}$` guard drops GitHub logins / shell metacharacters / `<!channel>` before interpolation; `tr -d '[:space:]']` prevents any multi-line `$GITHUB_ENV` injection from a commit body; the seed-commit walk cannot mis-page a stranger on the public `main`; and no behavior change when no requester is in scope.

### DIFF
--- a/.github/workflows/integrate-model.yml
+++ b/.github/workflows/integrate-model.yml
@@ -216,10 +216,18 @@ jobs:
           set -uo pipefail
           REQ="${INPUT_REQUESTED_BY:-}"
           if [ -z "$REQ" ]; then
-            # Newest-first, capped: the seed is the branch's oldest commit ahead of base and the
-            # only one carrying the trailer, so the first match is it; -n 50 bounds the log walk.
-            REQ="$(git log -n 50 --format=%B | sed -nE 's/^requested_by:[[:space:]]*//p' | head -1 | tr -d '[:space:]')"
+            # Scope to THIS branch's commits ahead of base (fetch-depth: 0 fetches origin/<base>):
+            # the seed carries the trailer and no integration commit does, so the first match is the
+            # seed's. Unscoped, `head -1` could pick a `requested_by:` line a prior squash-merge
+            # preserved in the base branch's own history and page an unrelated user. A missing
+            # origin/<base> (2>/dev/null) leaves REQ empty and the standing list takes over.
+            REQ="$(git log "origin/${GITHUB_BASE_REF:-main}..HEAD" --format=%B 2>/dev/null | sed -nE 's/^requested_by:[[:space:]]*//p' | head -1)"
           fi
+          # One line only: strip every whitespace / control char so a multi-line value (which the
+          # workflow_dispatch API accepts even where the web UI would not) cannot inject a second
+          # KEY=VALUE line into $GITHUB_ENV. Applied to BOTH sources. A valid Slack id carries none;
+          # `resolve_mentions` re-validates the shape before it is rendered into a mention.
+          REQ="$(printf '%s' "$REQ" | tr -d '[:space:][:cntrl:]')"
           printf 'SLACK_REQUESTED_BY=%s\n' "$REQ" >> "$GITHUB_ENV"
           echo "Slack requester: '${REQ:-<none>}' (empty or non-user-id -> standing mention list)"
 
--- a/.github/workflows/integrate-model.yml
+++ b/.github/workflows/integrate-model.yml
@@ -136,12 +136,11 @@ jobs:
       SLACK_BOT_TOKEN: ${{ secrets.ARENA_AUTOMATION_SLACK_BOT_TOKEN }}
       SLACK_CHANNEL: ${{ vars.ARENA_AUTOMATION_SLACK_CHANNEL }}
       SLACK_MENTIONS: ${{ vars.ARENA_AUTOMATION_SLACK_MENTIONS }}
-      # WHO a `--mention` notification pages: the requester ALONE when the run came in through
-      # Slack, else SLACK_MENTIONS (see `automation.slack.resolve_mentions`). Set from the poller's
-      # forwarded input on the Slack path; the "Resolve the Slack requester" step below fills it
-      # from the seed commit on the label-triggered path, where inputs are absent. A value that is
-      # not a Slack user id is dropped by `resolve_mentions`, so the standing list takes over.
-      SLACK_REQUESTED_BY: ${{ inputs.requested_by }}
+      # SLACK_REQUESTED_BY (WHO a `--mention` pings: the requester ALONE when the run came in
+      # through Slack, else SLACK_MENTIONS - see `automation.slack.resolve_mentions`) is NOT declared
+      # here on purpose: a job-level `env:` entry can shadow a value a step writes to $GITHUB_ENV, so
+      # the label-triggered path (no input) could never override an empty default. The "Resolve the
+      # Slack requester" step is the SINGLE writer instead, reading the input at step scope.
       # Values the ${secret:NAME} references in the LLM_PROXY_HEADERS variable resolve to.
       # Named LLM_PROXY_* because they are gateway configuration, not automation config;
       # they are NOT companions in the resolve_proxy_config sense (that check names only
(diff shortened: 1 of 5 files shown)


.github/workflows/integrate-model.yml
```diff
@@ -187,6 +200,29 @@ jobs:
...
+        # and the standing SLACK_MENTIONS list takes over.
+        env:
+          INPUT_REQUESTED_BY: ${{ inputs.requested_by }}
+        run: |
+          set -uo pipefail
+          REQ="${INPUT_REQUESTED_BY:-}"
+          if [ -z "$REQ" ]; then
+            # Newest-first, capped: the seed is the branch's oldest commit ahead of base and the
+            # only one carrying the trailer, so the first match is it; -n 50 bounds the log walk.
+            REQ="$(git log -n 50 --format=%B | sed -nE 's/^requested_by:[[:space:]]*//p' | head -1 | tr -d '[:space:]')"
```
CiroGamboa: **`git log -n 50 --format=%B` fallback walks past the branch's seed into `main`, can page a stale requester**

The fallback walk isn't scoped to commits ahead of base, so a `requested_by:` trailer preserved by a prior squash-merge onto `main` can be picked up here.

Failure scenario: `main` already carries `requested_by:` trailers from earlier squash-merged integrations — GitHub's default squash-merge preserves seed-commit bodies as bullet lists (`git log origin/main -100 | grep '^requested_by:'` today shows values like GitHub logins / emails, which the Python-side `_SLACK_USER_ID_RE` shape check discards). Once a Slack-triggered integration lands on `main` with its `requested_by: U0ABC123DEF` trailer preserved, a subsequent label-triggered run whose own seed lacks a trailer will `head -1` that stale line and page an unrelated Slack user.

Fix: scope the walk to commits ahead of base — e.g. `git log "origin/${GITHUB_BASE_REF:-main}..HEAD" --format=%B` — or read HEAD's own body if the by-hand seed is documented to be the tip.

.github/workflows/integrate-model.yml
```diff
@@ -187,6 +200,29 @@ jobs:
...
+          fi
+          printf 'SLACK_REQUESTED_BY=%s\n' "$REQ" >> "$GITHUB_ENV"
```
CiroGamboa, line 231: **`printf '...=%s\n' "$REQ" >> "$GITHUB_ENV"` writes the input unsanitized — a multi-line `requested_by` value can inject a second `KEY=VALUE` line**

On the input branch, `REQ="${INPUT_REQUESTED_BY:-}"` has no shape check and no `tr -d '[:cntrl:]'` (only the git-log fallback strips whitespace).

Failure scenario: a `workflow_dispatch` invocation whose `requested_by` input contains a newline followed by e.g. `GH_TOKEN=…` — accepted by the workflow_dispatch API even when GitHub's web UI would not enter it — lands two lines in `$GITHUB_ENV` and hijacks the second variable for the rest of the job. The Python-side `looks_like_slack_user_id` check catches the malformed value, but only *after* the env write, so it does not undo the injection.

Fix: validate the shape in shell before the write, or wrap with an `EOF`-delimited heredoc after stripping control chars.

tools/automation/src/automation/slack.py
```diff
@@ -98,6 +101,53 @@ def build_mention_suffix(raw: str | None) -> str:
...
+#: poller (slack-integrate.yml) already has it as ``requester`` for the allowlist check; it forwards
+#: it as the ``requested_by`` workflow input, and the workflow exports it here. The engine's eval
+#: notifier reads the SAME variable name, so one mental model covers both flows.
+REQUESTED_BY_ENV = "SLACK_REQUESTED_BY"
+
+#: What a Slack user id looks like (``U…``/``W…``, upper-case alphanumeric). The requester value
+#: originates in a Slack message's platform metadata and crosses a repo boundary as a workflow
+#: input before being rendered back INTO Slack as a mention - so anything that does not look like a
+#: user id is dropped rather than interpolated, and the configured list takes over.
+_SLACK_USER_ID_RE = re.compile(r"^[UW][A-Z0-9]{4,}$")
```
CiroGamboa: **`_SLACK_USER_ID_RE` accepts 5-char ids like `UAAAA`; real Slack user ids are ≥9 chars**

`^[UW][A-Z0-9]{4,}$` is looser than the trust boundary it guards.

Failure scenario: combined with the `git log` walk finding above, a stray 5-char token starting with `U`/`W` in a nearby `main` commit body's `requested_by:` line (or a hand-crafted `workflow_dispatch` input) passes shape validation and is interpolated back into `<@Uxxxx>`, either pinging a wrong workspace member if it happens to be a real id or rendering a dead mention.

Fix: raise the length floor to match Slack's real id shape (`{8,}` at minimum).

tools/automation/src/automation/slack.py
```diff
@@ -98,6 +101,53 @@ def build_mention_suffix(raw: str | None) -> str:
...
+    the requester ALONE is paged - it is their integration, and everyone else can read the thread
+    without being pulled into it. ``SLACK_MENTIONS`` stays the fallback for runs nobody asked for
+    through Slack (a by-hand label add or a manual ``workflow_dispatch`` has a GitHub actor but no
+    Slack identity).
+
+    The requester value crossed a repo boundary as a workflow input, so it is re-validated here
+    rather than trusted: anything not shaped like a Slack user id is dropped and the list takes
+    over. Unset is not an error - a run with nobody to page still posts, just without a ping.
+    """
+    requester = (os.environ.get(REQUESTED_BY_ENV) or "").strip().strip("<>").lstrip("@")
```
CiroGamboa: **Normalization prelude `.strip().strip("<>").lstrip("@")` is duplicated between `resolve_mentions` and `looks_like_slack_user_id`**

Nit / drift risk: the two call sites agree by hand on a normalization that will need to grow (e.g. accepting `<@Uxxx|display-name>`).

Fix: extract a `_normalize_user_id(value)` helper and call it from both so the accepted shapes stay in lock-step.

CiroGamboa (approved):

PR #1542, opened 2026-09-07
feat(automation): cost summary for the integration run (agents, wire probes, key deltas)

balazs-toloka:
## What

Every auto-integration run now ends with an accounting tail once the candidate is known, whatever the outcome: `automation cost-summary` writes `observation/cost/cost_summary.{json,md,txt}`, posts the markdown as a PR comment, sends the one-liner to the Slack thread (new icon role `cost_summary`, no mention) and uploads the `integration-cost-pr<N>` artifact. Three sources, kept apart because they measure different things:

| Source | Covers | What the number is |
|---|---|---|
| the `claude -p` result event of every resolve iteration and the finalize run (the agent calls now run with `--verbose --output-format stream-json \| tee observation/resolve/agent_*.jsonl`) | the agent | the CLI's own report at Anthropic list price; cache writes are not surfaced on the gateway route, so a lower bound |
| `wire_probes_*/aggregate.json` (observe) and `resolve/reprobe_*/` (reprobe) | the candidate's wire calls only | exact for what it covers; the capability + variant probes and the user simulator leave no cost artifact |
| `automation key-snapshot` (OpenRouter `GET /api/v1/key`) at run start, after observe and at the end | everything billed on `ARENA_AUTOMATION_OPENROUTER_API_KEY`, probes included | the billed figure; key-level, so a concurrent run on the same key is included |

On the `litellm` route the probes bill the gateway key, so the `.env` step records the effective route in `observation/cost/route.txt` and the summary labels the key deltas as agent-only.

## What leaves the runner

Only the summary and the **normalized** agent result events (`observation/cost/agent_*.json`: cost, turns, usage, how the run ended), written by `automation agent-digest <file> --out`. Never the raw transcripts: a `Bash(env)` or `cat .env` tool result in them would carry the run's keys, and GitHub masks the job log, not an artifact of a public repo. The key snapshots hold the key's lifetime usage figure, so they live in `$RUNNER_TEMP/cost` and only their deltas reach the summary (`cost-summary --key-dir`).

## Invariants

- Accounting can never fail the integration: every call is `|| true`, both new steps are `continue-on-error`, `key-snapshot` exits 0 on any miss, and a run with nothing spent on record writes the JSON but no md/txt, so nothing is posted.
- The key reaches `uv run automation` in the tail step under the same accepted-risk envelope as the existing post-agent `automation reprobe` calls (documented in the workflow).
- The key is read through `SecretManager([EnvProvider()])` (env-only, the same way `gateway_catalog.py` does it), never `os.environ` directly.
- On an observe -> needs-human run there is no after-observe snapshot (the gateway step is gated on a clean observe), so the summary attributes the whole key delta to observe and says so; wire aggregates with `total_cost_usd: null` (no pricing) are counted and noted rather than read as zero.

## Trade-off

The agent transcript in the job log is now one JSON event per line (stream-json) instead of the `--verbose` text rendering; `automation agent-digest` prints a one-line digest (subtype, turns, cost) per run so the log still says how a run ended.

## Verification

- `tools/automation/tests`: 306 passed (new `test_cost_summary.py`; `test_icons` checks every `--icon` role used in the workflow is declared and used).
- `ruff check` + `black --check` clean on the touched files; the workflow parses.
- Two independent review rounds; round 1 found the transcript-in-artifact exposure, the litellm-route under-report and the snapshot privacy issue, all fixed.

Not run: a live integration (needs a candidate PR); `actionlint` locally (runs in CI).

### DIFF
--- a/tools/automation/src/automation/cost_summary.py
+++ b/tools/automation/src/automation/cost_summary.py
@@ -63,7 +63,7 @@ def _usd(value: float | None, *, escape: bool = False) -> str:
     if value is None:
         return "n/a"
     # GitHub renders `$...$` on one line as math; the markdown body escapes its dollars.
-    return ("\\$" if escape else "$") + f"{value:,.4f}"
+    return ("\\$" if escape else "$") + f"{value:,.2f}"
 
 
 def _num(value: Any, default: float = 0.0) -> float:
--- a/tools/automation/src/automation/cost_summary.py
+++ b/tools/automation/src/automation/cost_summary.py
@@ -1,0 +1,568 @@
+"""Cost accounting for one auto-integration run (``automation cost-summary``).
+
+Three sources, three different truths, kept apart on purpose:
+
+* **agents** - the resolve / finalize ``claude -p`` runs report their own spend
+  (``total_cost_usd`` in the CLI's ``result`` event). That is the Anthropic list price for the
+  alias; on the gateway route cache WRITES are not surfaced, so it is a lower bound.
+* **wire probes** - the engine's own ``aggregate.json`` under every ``wire_probes_*`` run
+  (observe) and ``resolve/reprobe_*`` (reprobe). Exact for the candidate's wire calls; the
+  capability + variant probes and the user simulator leave no cost artifact at all.
+* **key usage** - snapshots of the automation key's lifetime usage (OpenRouter
+  ``GET /api/v1/key``) at run start, after observe and at the end. Their deltas are the BILLED
+  figure and the only one that also covers the probes without an artifact. Key-level, so a
+  concurrent run on the same key is included; on the ``litellm`` route the probes bill the
+  gateway key instead, and the deltas cover the agent alone. The summary says both.
+
+What may leave the runner: the summary and the NORMALIZED agent result events (cost, turns,
+usage, how the run ended - never the transcript, which can echo the agent's environment). The
+key snapshots carry the key's lifetime usage figure, so the workflow keeps them outside the
+observation dir and only their deltas reach the summary.
+
+See docs/AUTO_INTEGRATION.md § Cost summary.
+"""
+
+from __future__ import annotations
+
+import datetime as _dt
+import glob
+import json
+import re
+import sys
+import urllib.error
+import urllib.request
+from pathlib import Path
+from typing import Any
+
+from tolokaforge.secrets import EnvProvider, SecretManager
+
+KEY_USAGE_URL = "https://openrouter.ai/api/v1/key"
+
+#: Under the observation dir: the effective route marker, the normalized agent result events
+#: and the rendered summary - everything in here is safe to upload.
+COST_DIR = "cost"
+ROUTE_FILE = "route.txt"
+DEFAULT_ROUTE = "openrouter"
+GATEWAY_ROUTE = "litellm"
+KEY_SNAPSHOTS = ("key_start", "key_after_observe", "key_end")
+
+_USAGE_FIELDS = (
+    "input_tokens",
+    "output_tokens",
+    "cache_read_input_tokens",
+    "cache_creation_input_tokens",
+)
+_ITER_RE = re.compile(r"^agent_iter_(\d+)\.")
+
+
+def _log(message: str) -> None:
+    print(f"cost-summary: {message}", file=sys.stderr)
+
+
+def _usd(value: float | None, *, escape: bool = False) -> str:
+    if value is None:
+        return "n/a"
+    # GitHub renders `$...$` on one line as math; the markdown body escapes its dollars.
+    return ("\\$" if escape else "$") + f"{value:,.4f}"
+
+
+def _num(value: Any, default: float = 0.0) -> float:
+    try:
+        return float(value)
+    except (TypeError, ValueError):
+        return default
+
+
+# ---------------------------------------------------------------------------
+# Agent results (claude -p output)
+# ---------------------------------------------------------------------------
+
+
+def parse_agent_result(text: str) -> dict[str, Any] | None:
+    """The ``result`` event of a ``claude -p`` run, whatever output format produced *text*.
+
+    ``--output-format json`` prints one object; with ``--verbose`` it prints an array of every
+    message; ``stream-json`` prints one JSON object per line. All three end in a
+    ``{"type": "result", ...}`` object, which is the one carrying the totals. ``None`` when no
+    such object is found (an agent that never got to answer, or a file that is not JSON).
+    """
+    if not text or not text.strip():
+        return None
+    try:
+        parsed = json.loads(text)
+    except ValueError:
+        parsed = None
+    if isinstance(parsed, dict):
+        return parsed if parsed.get("type") == "result" else None
+    if isinstance(parsed, list):
+        results = [
+            item for item in parsed if isinstance(item, dict) and item.get("type") == "result"
+        ]
+        return results[-1] if results else None
+    # JSON lines: keep the last result event, ignore anything that does not parse.
+    last: dict[str, Any] | None = None
+    for line in text.splitlines():
+        line = line.strip()
+        if not line.startswith("{"):
+            continue
+        try:
+            item = json.loads(line)
+        except ValueError:
+            continue
+        if isinstance(item, dict) and item.get("type") == "result":
+            last = item
+    return last
+
+
+def iteration_number(path: Path) -> int | None:
+    """The resolve iteration a result file belongs to, from its name (``agent_iter_<n>.*``)."""
+    match = _ITER_RE.match(path.name)
+    return int(match.group(1)) if match else None
+
+
+def normalize_agent_result(
+    result: dict[str, Any] | None, *, file: str, iteration: int | None = None
+) -> dict[str, Any]:
+    """The subset of a result event the summary keeps - cost, turns, usage, how the run ended.
+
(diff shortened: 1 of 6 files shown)


tools/automation/src/automation/cost_summary.py
```diff
@@ -0,0 +1,568 @@
...
+    try:
+        manifest = json.loads(manifest_path.read_text()) if manifest_path.exists() else {}
+    except (OSError, ValueError):
+        manifest = {}
+    if not isinstance(manifest, dict):
+        manifest = {}
+    route = read_route(obs_dir)
+    agents = agent_costs(obs_dir)
+    wire = wire_costs(obs_dir)
+    usage = key_usage(key_dir if key_dir is not None else obs_dir / COST_DIR)
```
CiroGamboa, line 347: **Default `key_dir` is `obs_dir/cost` — inside the observation bundle uploaded as a public artifact**

The workflow currently passes `--key-dir "$RUNNER_TEMP/cost"`, so today this is safe. But if that flag is ever dropped, the key's lifetime `usage_usd` figure would land in `integration-observation-pr<N>`.

Failure scenario: someone edits the workflow and removes `--key-dir "$RUNNER_TEMP/cost"` from `cost-summary`; if snapshots are also written under `observation/cost/`, the raw `{"usage_usd": <lifetime>, ...}` snapshot rides out in the observation artifact of a public repo. The module docstring says these must never leave the runner, so the default `key_dir` is at odds with that invariant.

Fix: default `key_dir` to something explicitly outside `obs_dir` (a `TemporaryDirectory`, or make the flag required).

tools/automation/src/automation/cost_summary.py
```diff
@@ -0,0 +1,568 @@
...
+    workflow keeps these files out of every artifact.
+    """
+    if api_key is None:
+        # Env-only, like gateway_catalog: the workflow maps the secret in per step, and dotenv
+        # precedence would let a developer's local .env answer a CI run.
+        api_key = SecretManager([EnvProvider()]).get_secret("OPENROUTER_API_KEY")
+    if not api_key:
+        _log("no OPENROUTER_API_KEY in the environment; skipping the key snapshot")
+        return False
+    request = urllib.request.Request(KEY_USAGE_URL, headers={"Authorization": f"Bearer {api_key}"})
```
CiroGamboa, line 279: **`snapshot_key` uses default `urllib.request.urlopen`, which follows redirects and can forward `Authorization: Bearer <api_key>` to the redirect target**

`urllib.request.HTTPRedirectHandler` re-issues the request with the same `Request` object, so the `Authorization` header rides along on same-host redirects (all Python versions) and, on Python <3.11, on cross-host redirects too.

Failure scenario: if OpenRouter's `/api/v1/key` ever returns a 30x redirect to a different host (CDN, migration, hijacked DNS), the OpenRouter automation Bearer token is transmitted to that host.

Fix: subclass `HTTPRedirectHandler` to strip `Authorization` on redirect, or switch to `requests`/`httpx` with an explicit `allow_redirects=False`.

tools/automation/src/automation/cost_summary.py
```diff
@@ -0,0 +1,568 @@
...
+    except ValueError:
+        data = None
+    if _is_normalized(data):
+        return data
+    return normalize_agent_result(
+        parse_agent_result(text), file=path.name, iteration=iteration_number(path)
+    )
+
+
+def _agent_files(obs_dir: Path) -> tuple[list[Path], Path | None]:
```
CiroGamboa, line 167: **`_agent_files` returns iterations AND finalize from the FIRST matching directory — a mixed state silently drops finalize**

The fallback is per-directory-all-or-nothing, not per-file.

Failure scenario: if `observation/resolve/agent_iter_*.jsonl` exists (tee'd) but `observation/resolve/agent_finalize.jsonl` is missing (e.g., the finalize `tee` fails on the runner while the earlier `agent-digest --out observation/cost/agent_finalize.json` from a re-run remains), the function returns `(iterations from resolve/, finalize=None)` and the finalize cost in `cost/` is ignored — even though the docstring implies `cost/` is a fallback.

Fix: resolve iterations and finalize independently, each with its own directory search.

tools/automation/src/automation/cost_summary.py
```diff
@@ -0,0 +1,568 @@
...
+        "unpriced_runs": unpriced,
+        "total_cost_usd": round(total, 6),
+        "tokens": tokens,
+    }
+
+
+def wire_costs(obs_dir: Path) -> dict[str, Any]:
+    # Same discovery as observe._wire_findings: the observe stage writes one wire_probes_* run
+    # directly under the obs dir, a reprobe writes its own under resolve/reprobe_<i>/.
+    observe = _aggregate_costs(sorted(glob.glob(str(obs_dir / "wire_probes_*" / "aggregate.json"))))
```
CiroGamboa, line 238: **`glob.glob(str(obs_dir / ...))` isn't `glob.escape`d — bracket/`*`/`?` in `obs_dir` will be interpreted as glob metacharacters**

Not a runner issue (`obs_dir='observation'`) but silently wrong for local dev.

Failure scenario: run the CLI locally under a path with `[` (e.g. macOS `~/Some [WIP] repo/observation`) — `glob.glob('.../Some [WIP] repo/observation/wire_probes_*/aggregate.json')` returns `[]` because the bracket is interpreted as a character class. Same for the reprobe glob. cost-summary reports 0 wire cost with no warning.

Fix: `glob.escape(str(obs_dir))` before joining, or switch to `pathlib.Path.glob`.

tools/automation/src/automation/cost_summary.py
```diff
@@ -0,0 +1,568 @@
...
+GATEWAY_ROUTE = "litellm"
+KEY_SNAPSHOTS = ("key_start", "key_after_observe", "key_end")
+
+_USAGE_FIELDS = (
+    "input_tokens",
+    "output_tokens",
+    "cache_read_input_tokens",
+    "cache_creation_input_tokens",
+)
+_ITER_RE = re.compile(r"^agent_iter_(\d+)\.")
```
CiroGamboa, line 55: **`_ITER_RE = ^agent_iter_(\d+)\.` matches any file starting with `agent_iter_N.` — including `.tmp`, `.bak`, `~` leftovers**

Combined with `agent_iter_*.json*` in the glob, a stale sidecar can be counted as an iteration.

Failure scenario: a crashed `tee`, an editor swap file, or a partial re-run leaves `agent_iter_1.jsonl~` or `agent_iter_1.json.tmp` alongside the real `agent_iter_1.jsonl`. Both match; sorting by iteration number puts them adjacent; each is read and its cost added — a single iteration is double-counted (or, if the tmp parses to a stale result, silently overwrites the correct one).

Fix: anchor the regex to `\.(json|jsonl)$`.

tools/automation/src/automation/cost_summary.py
```diff
@@ -0,0 +1,568 @@
...
+            "attributed": attributed,
+            "agents": agents["total_cost_usd"],
+            "wire_probes": wire["total_cost_usd"],
+            "measured_key_delta": measured,
+        },
+        # A run with nothing on record (parse failed early, or every source is missing) has
+        # nothing worth a PR comment or a Slack line; the JSON still records that. A run that
+        # demonstrably ran agents or probes is reported even when every figure came back empty.
+        "noteworthy": (
+            attributed > 0 or bool(measured) or agents["runs"] > 0 or wire["observe"]["runs"] > 0
```
CiroGamboa, line 410: **`bool(measured)` treats a legitimate `0.0` delta as falsy in the `noteworthy` check**

A run whose `measured_key_delta == 0.0` (rare but possible: identical snapshots on a rate-limited or cached-only iteration) reports not-noteworthy even though a run did happen.

Failure scenario: if `attributed == 0.0` (all parses failed), `agents['runs'] == 0` (files gone), `wire['observe']['runs'] == 0` (no aggregate), and only the key snapshots exist with identical usage, `noteworthy=False` and no PR/Slack line is posted — hiding the fact that a run occurred (the key was reachable).

Fix: use `measured is not None` (a valid zero-delta measurement should count).

tools/automation/src/automation/cost_summary.py
```diff
@@ -0,0 +1,568 @@
...
+        manifest = json.loads(manifest_path.read_text()) if manifest_path.exists() else {}
+    except (OSError, ValueError):
+        manifest = {}
+    if not isinstance(manifest, dict):
+        manifest = {}
+    route = read_route(obs_dir)
+    agents = agent_costs(obs_dir)
+    wire = wire_costs(obs_dir)
+    usage = key_usage(key_dir if key_dir is not None else obs_dir / COST_DIR)
+    attributed = round(agents["total_cost_usd"] + wire["total_cost_usd"], 6)
```
CiroGamboa, line 348: **On the `litellm` route, `attributed = agents + wire_probes` sums costs charged to TWO different keys**

Agents billed on the automation key, wire probes billed on the gateway key — no separation in the totals block. The route caveat is in a note, but the JSON `totals_usd.attributed` carries no such distinction.

Failure scenario: a reader (or a downstream dashboard/budget) treats `totals_usd.attributed` on a litellm-route run as "what this integration cost the automation budget" and over-attributes by the wire aggregate.

Fix: split the JSON into `attributed_automation` / `attributed_gateway` on the litellm route, or gate the sum behind `route == "openrouter"`.

CiroGamboa (approved):

Issue #1546, opened 2026-09-08
[umbrella] arena-v3-engine-residuals — engine defects surviving 0.23.1


CiroGamboa:
## Problem

An audit of the arena v3 task packs — verified against engine 0.21.4 in early September 2026 and re-validated against the current 0.23.1 code — surfaced a residue of engine-side defects that survive the current version and cost real trials on the v3 board. Two of them (child A and child B below) are class-level bugs: the same shape recurs across multiple task packs and multiple downstream oracle-strictness findings, and every fix attempted at the pack level has produced consistency drift. One (child C) is a latent trace_checks default that doesn't bite the current pin but silently distorts any future board that adopts trace-checks-based grading.

## Product outcome

When this milestone lands: (1) task packs that register tools with a doubled system-prefix stop losing trials to unhelpful "Tool not found" errors; (2) task authors get an engine-level way to say "this tool-call param is permitted-extra" instead of hand-maintaining per-pack unstable-fields masks, which removes a whole class of sibling-inconsistency bugs; and (3) trace-checks authors no longer have to guess about the failure mode of an anchor whose tool errored silently. Net effect for the v3 board: hundreds of trials that were failing on engine-visible pack-shape mismatches now score based on the model's actual behavior.

## Key decisions surfaced

All three forks are now resolved. Recording here for reference:

- **TD-25 error-response shape** — resolved: Option A (hint only) as an engine-side safety net; primary root-cause fix lives on the pack tool-registry side (double-prefixing when a tool name already carries the system), filed in the internal tracker. Silent alias is out — would diverge transcript from registered tool and hide the pack bug. See child A.
- **TD-24 schema shape** — resolved: per-column `compare_columns: {mode: subset, extras_allowed_for: [...]}` on the state_checks per-table config. Mirrors trace-side `compare_args` semantics. Pack-side migration to replace hand-maintained `unstable_fields` masks filed separately in the internal tracker. See child B.
- **N9 default** — resolved: deferred out of this milestone. No arena v3 pack uses `trace_checks`, so this is not a v3 publish blocker. Long-term: flip to `WITHHOLD` if no external consumer relies on `FAIL`, else keep `FAIL` + add authoring warning. Will be re-opened as a standalone follow-up. See #1549.

### Phase 1 — Publish blockers for v3 (parallel-safe)
- [ ] #1547 — Engine gives no server-side rescue on a near-miss tool name (TD-25)
[...]

PR #1551, opened 2026-09-08
feat(runner): hint near-miss registered names on Tool not found

### LINKED ISSUE #1547: feat(runner): rescue near-miss tool names in dispatch so a wrong short name is recoverable
Part of #1546

## Context

Task packs frequently register tools with a doubled `system×tool` name suffix — for example, a `send_notification` tool inside a system also called `send_notification` ends up registered as `<pack>_send_notification_send_notification`. Models read this in the tool schema and reasonably try to call the shorter natural name, e.g. `send_notification`. The engine's tool dispatcher does an exact string lookup, misses, and returns bare `"Tool not found"` with no additional hint. On the arena v3 run this cost hundreds of trials across two task domains, with recovery rates well below half of affected trials.

## Deliverables

- [ ] The engine's tool-not-found response on a near-miss short name gives the model enough information to recover on the next turn (exact behavior depends on the option the team picks — see Open Questions).
- [ ] Exact-match dispatch behavior is unchanged.
- [ ] Ambiguous match — where two registered tools share the requested short suffix — returns a distinct disambiguation error, never silent dispatch.
- [ ] Engine unit tests cover: exact match still works; near-miss returns the chosen behavior; ambiguous match returns disambiguation.

## Acceptance Criteria

- On a task pack that registers tools with a doubled system-prefix, a model that calls the natural short name is either dispatched correctly (if silent alias is chosen) or receives an error message that names the candidate tools (if hint is chosen), and successfully calls the correct tool on its next turn.
- A synthetic pack where two registered tools share the same short suffix returns the ambiguity error; the model can then choose deterministically.
- Existing task packs that call tools by their exact registered name see no behavior change.
- One clean smoke run on the reproducer pack in the internal task repo shows the "Tool not found" count dropping to near-zero on the same model + repeats config that the audit ran.

## Design Decisions

None resolved yet — see Open Questions.

**Which response shape do we want?**

- **Option A — "Did you mean?" hint only.** The "Tool not found" error still fires, but the message enumerates up to three nearest candidates (edit distance and/or unique suffix). Model self-corrects on the next turn. Same shape as the engine's existing invalid-*table* branch on the KB search tool. Cheapest; keeps the pack registration bug visible in logs; costs the model one wasted turn per near-miss.
- **Option B — Silent unique-suffix alias resolution.** If exactly one registered tool has a name ending with the requested short name, dispatch it silently. Best possible score impact; masks the pack registration bug from anyone looking at trials; the ambiguity path must be well-defined and tested.
- **Option C — Both, alias behind a config flag.** Hint is always on; unique-suffix alias is opt-in via a per-run engine config knob. More surface, more code, but gives us both signal (packs still visibly broken in logs) and score (benchmarks can flip on alias).

**Leaning:** A — cheapest, keeps the pack signal visible, no new configuration surface.

**Open question for the team:** does silent alias resolution violate any correctness contract we care about — e.g., does the model's transcript need to reflect the *exact* registered tool name it dispatched, for downstream trace-based grading or replay?

## Out of Scope

- Renaming the doubled system-prefix in task packs. That is a task-data-team follow-up; this issue makes the engine tolerant of the shape, it does not clean up the shape at its source.
- The related travel-marketplace failure where documentation references unprefixed short names that do not exist under any spelling. That is a docs/registry mismatch on the pack side, not a dispatch defect on the engine side. This fix would still help there because the model would be told which tool names *do* exist.

CiroGamboa:
## Summary

- `_resolve_tool_or_refuse` now enriches the `Tool not found` error with up to three registered names whose suffix matches `_{tool_name}`, so a model calling the natural short name of a doubled-prefix registration (`<system>_<tool>_<tool>`) can retry against the real name on the next turn.
- Exact-match dispatch is unchanged. When nothing matches, the error stays terse ("Tool '<name>' not found"). Ambiguous suffix matches list every candidate — the engine never silently picks one.
- New helper `TrialContextRuntime.nearest_tool_names` centralises the ranking; it's suffix-match first (catches the reported stutter shape) then substring, sorted for determinism, capped at 3.

## Design choice

Team picked Option A on #1547: hint only, no silent alias. Silent alias would diverge the transcript from the registered tool (breaks replay + trace grading) and hide the pack-side bug from logs. The primary root-cause fix — the pack-generator's `tool_registry.py:151` `final_name = f"{effective_namespace}_{tool_name}"` double-prefixing when the tool name already carries the system — is filed separately on the internal tracker; this engine change is the safety net.

## Test plan

- [x] `pytest tests/unit/test_runner_pipeline.py -k "not_found or hints or ambiguous or near_miss"` — 3 new + 3 existing tests pass.
- [x] Existing `test_execute_tool_tool_not_found` and `test_unknown_tool_is_recorded_not_only_reported` still pass (message still contains "not found").
- [x] `ruff check` + `ruff format --check` clean.

Closes #1547
Part of #1546


CiroGamboa:
Addressed code-review findings in `16d713aa`:

- **F1** Suggested candidates now exclude names marked unusable via `mark_tool_unusable` (would waste a retry turn).
- **F2 / F6** Raised cap 3 → 5 and changed the return type of `nearest_tool_names` to `(candidates, total_matches)` so the error message signals `and N more registered with the same suffix` when the alphabetic tiebreak would otherwise hide equally-plausible candidates. New test `test_tool_not_found_hint_signals_hidden_matches_when_capped`.
- **F3** Warning log now includes the candidates it computed.
- **F5** Removed the dead `if name == tool_name: continue` guard (unreachable — caller only invokes after `get_tool` returned None).
- **F7** Added a comment on `nearest_tool_names` explaining why suffix-first over `difflib.get_close_matches` (the shape being rescued has a known signature; edit distance would return arbitrarily similar names for typos).

Not fixed in this PR: **F4** (sibling `ToolRegistry.execute_tool` path in `tools/registry.py` still emits the bare error). Different call surface with different callers; deserves its own scoping. Happy to file as a follow-up if you agree.

All existing tests plus three new ones pass; lint + format clean.

CiroGamboa:
Alignment check confirmed against the teammate feedback on the parent issue:

- **Primary fix**: the root cause is the pack tool registry double-prefixing when a tool name already carries the system slug. That fix has now been filed on the internal tracker and will land there — one edit in the generator fixes every affected pack in one go.
- **This PR**: safety net only. Option A (hint enumerating candidates), no silent alias — transcript stays truthful to the registered name, and the pack bug remains visible in logs so it does not silently drift back after the source fix ships. Merging this is optional but defensive; happy to hold or ship at your call.

CiroGamboa:
Superseded by #1560 (consolidated mega-PR to main).

PR #1552, opened 2026-09-08
feat(grading): compare_columns subset mode for permitted-extra tool-call params

### LINKED ISSUE #1548: feat(grading): state-hash comparator gains an opt-in permitted-extra mode for tool-call params
Part of #1546

## Context

The state-hash comparator that grades a task's final state is byte-exact on tool-call param objects. When a task pack's system prompt explicitly permits an optional param (a common shape: "you may include `param_case_number` in a notification"), a model that includes the permitted param passes the trace-based comparator — which already understands subset comparison via `compare_args` — but fails the state-hash comparator, because the resulting state row differs byte-for-byte from the golden. The only workaround today is per-pack `unstable_fields.json` masks, which task teams maintain by hand. Those masks have become inconsistent between sibling tasks in the same pack, and that inconsistency is the root cause of a whole class of downstream oracle-strictness findings in the same audit.

## Deliverables

- [ ] The state-hash comparator supports a new per-table configuration that lets task authors declare tool-call param columns as "permitted-extra" — extra keys not present in the golden do not cause a hash mismatch.
- [ ] The new mode is symmetric with the trace comparator's existing subset-compare shape, so the two paths use the same mental model.
- [ ] Default behavior is unchanged when the new field is unset — this is opt-in per pack.
- [ ] The new mode composes with existing `unstable_fields` masks (they don't fight each other or double-apply).
- [ ] Engine unit tests cover: default byte-exact behavior; declared permitted-extra allows extras; declared without permitted-extra still fails on the same input; composition with `unstable_fields` masks.

## Acceptance Criteria

- On a synthetic pack where the golden lacks a permitted-extra param and the model's completion includes it, the state check passes with the new config declared and fails without it.
- On a real reproducer pack that already uses `unstable_fields` for the same shape, declaring the new config in place of the mask produces the same pass verdict and no regressions in state-hash-only tests for that pack.
- A state-perfect completion on a permitted-extra shape crosses the pass bar in the smoke run for the reproducer pack.

## Design Decisions

- **Live in `state_checks` config, not in `unstable_fields`** — chosen because it mirrors the existing trace comparator's subset-compare shape (`compare_args`) rather than piling more work onto the hand-maintained mask mechanism that already produces consistency drift.
- **Opt-in per pack, default unchanged** — chosen because a global semantic change to state-hash comparison would silently rescore every existing pack.

**Exact schema key + granularity.** Two shapes are on the table:

- Per-column: `compare_columns: {mode: subset, extras_allowed_for: [params]}` on the per-table config.
- Per-table with a column allowlist: `permit_extra_keys: [params]`.

Either shape delivers the same functionality; the naming choice affects how obvious the config is to task authors and how easy the field is to introspect during grading. The implementer's design-doc stage should settle this, but a naming preference from the team now would save a review round.

## Out of Scope

- Retroactively converting existing per-pack `unstable_fields` masks for this shape to the new config. That is a pack-side migration and belongs on the task-data team's tracker.
- Downstream oracle-strictness findings that were caused by inconsistent per-pack masks. Fixing the engine mechanism removes the underlying cause of new inconsistency; the existing per-pack masks stay in place until the pack-side migration lands.

CiroGamboa:
## Summary

- New per-(table, column) config `state_checks.compare_columns` of shape `{mode: subset, extras_allowed_for: [k1, k2]}`. When set, the state-hash comparator drops keys named in `extras_allowed_for` from the actual state's column-dict where the golden's version of the same column does not carry them, before hashing.
- Keys the golden declares are still compared value-for-value; extras outside the allowlist still fail; symmetric `unstable_fields` masks compose cleanly.
- Shape mirrors the trace comparator's `compare_args` on `RequiredAction` (`core/grading/transcript.py`) so both grading paths speak one dialect for the subset idea.

## Design choice

Team picked per-column `{mode: subset, extras_allowed_for: [...]}` on #1548. Naming mirrors the trace side (`compare_args`) rather than piling more work onto the per-pack `unstable_fields` mask mechanism that has produced consistency drift between sibling tasks. Rule is a typed struct (`ColumnCompareRule`) rather than a bare list so future modes (case-insensitive, numeric-tolerant) have room without a schema change. Default state-hash behavior is unchanged when the field is unset.

The motivating shape: a pack's SOP permits `notifications.params.param_case_number`, the golden omits it, the model fills it in — today the state hash rejects the model's completion. Long-term this replaces per-pack `unstable_fields` masks for permitted-extra params; the pack-side migration (which also removes the drift shape behind TD-19/27/28/29 on the audit tracker) is filed separately in the internal tracker.

## Concepts introduced

- `ColumnCompareRule` in `core/hash.py` — typed rule for the per-(table, column) subset mode.
- `apply_compare_columns_extras(actual, expected, rules)` in `core/hash.py` — asymmetric filter: returns `actual` with permitted-extra keys removed from the declared column dicts where `expected` did not carry them.

## Wiring scope

Wired through `combine.py` → `state_checks.py::check_hash_against_golden_replay` (the `frozen_mcp_core` / tau-bench-style hash path, which is what the QSR / retail arena v3 packs use). The runner-side `_grade_hash` at `runner/service.py:2596-2762` is a separate call path — it uses server-side `get_stable_hash` and would need a client-side raw-state fetch to apply the same filter. Filed as a follow-up rather than in-scope for this PR because no arena v3 reproducer hits that path.

## Test plan

- [x] `pytest tests/unit/grading/test_hash.py` — 9 new `TestApplyCompareColumnsExtras` cases + 21 existing tests pass.
- [x] `pytest tests/unit/grading/test_state_checks.py tests/unit/grading/test_state_composition.py tests/unit/grading/test_state_checks_composition.py tests/unit/grading/test_hash_verdict_parity.py` — 229 tests pass, no regressions from config-schema additions.
- [x] `ruff check` + `ruff format --check` + pre-commit hooks clean.

New unit tests cover: unset = no-op; declared extra absent from golden is dropped from actual (motivating case); extras outside the allowlist still fail; keys the golden declares are compared value-for-value; composition with `unstable_fields`; non-dict column values pass through; missing table/column skipped silently; list-of-rows and single-dict table shapes both supported.

Closes #1548
Part of #1546


CiroGamboa:
Addressed code-review findings in `b7f219be` and unblocked the `test-smoke` failure at the integration-branch level (canonical baseline drift, `097a6caa`):

- **F1 / F2 (CI-blocking)** Added `GradingKey` for `state_checks.compare_columns` in `key_manifest.py` (kind `CONFIG_INPUT`, coverage `CORE_ONLY`, enforcement `FIELD_RESOLUTION_ONLY`, `tracking_issue=1558` for the runner-path follow-up). Registered as a drift exemption in `_DRIFT_EXEMPTIONS`. Follow-up filed as #1558.
- **F3** `NativeAdapter.to_task_description` now threads `compare_columns=state_checks_data.get("compare_columns", {})` into `RunnerStateChecksConfig`. Also declared `compare_columns` in `tests/data/grading_parity/all_keys/grading.yaml` so `test_adapter_translation_carries_every_runner_key` proves the wire actually carries it.
- **F4 / F5** Docstring on `RunnerStateChecksConfig.compare_columns` explicitly states the runner-side hash grading path (`_execute_hash_grading` via db-service `get_stable_hash`) does not yet apply the filter, points at #1558.
- **F6** Expanded `apply_compare_columns_extras` docstring on positional row-pairing consequence (`StateDiff` reflects the pairing that was hashed).
- **F7** Expanded docstring for the missing-golden-column case — actual is left untouched because the containing hash mismatch already covers the schema difference; a golden without the column and an actual with it is a schema difference the state hash fails on anyway.
- **F8** Test assertions now use `state_digest` (the substrate the production path actually calls via `check_hash_against_golden_replay`), not `compute_stable_hash` (runner substrate).
- **F9** `extras_allowed_for` is materialised as a `frozenset` once per rule for O(1) membership checks inside the inner comprehension.
- **F10** Dropped the redundant `if compare_columns:` guard in `state_checks.py` — `apply_compare_columns_extras` is a no-op when empty/None, so the state_checks-side branch was dead weight.

All existing tests plus 9 new hash-filter tests plus 165 in the substrate-parity suite pass; lint + format clean.

**Baseline note:** the pre-existing `test_replay_matches_baseline` failure ("baseline names commit(s) unreachable from HEAD... history was rewritten") was fixed by regenerating the canonical baseline via `--update-canon`. Landed on the integration branch (`097a6caa`), then merged into both feature branches — both children now inherit the fix.

CiroGamboa:
Alignment check confirmed against the teammate feedback on the parent issue:

- **Shape (`compare_columns: {mode: subset, extras_allowed_for: [...]}`, per-column, mirrors trace-side `compare_args`)**: matches the teammate call exactly. Shipped in this PR.
- **Long-term direction**: this replaces the hand-maintained `unstable_fields` masks that produce the sibling-task drift behind TD-19 / TD-27 / TD-28 / TD-29 on the audit tracker. That pack-side migration is now filed as its own internal-tracker item so the follow-up is on the record and not lost.
- **Runner-side gap**: separate follow-up already public at #1558.

Issue #1558, opened 2026-09-08
feat(runner): wire state_checks.compare_columns through runner-side _grade_hash path


CiroGamboa:
Follow-up from #1548. That PR added `state_checks.compare_columns` (per-(table, column) `{mode: subset, extras_allowed_for: [...]}` subset semantics for the state-hash comparator) and wired it through the core / frozen_mcp_core / tau-bench-style hash path (`combine.py` -> `state_checks.py::check_hash_against_golden_replay`).

The runner-side `_execute_hash_grading` at `runner/service.py:2596-2762` still uses server-side `db_client.get_stable_hash()` and does not apply the `compare_columns` filter. Result: a pack that declares `compare_columns` and runs through the runner-side hash grading path (native adapter, MCP_SERVER wrappers) gets the config silently ignored. Cross-substrate parity is currently broken — the field is declared on both wire configs, consumed by core only.

## Deliverables
- [ ] `_execute_hash_grading` reads `state_checks.compare_columns` and applies the filter before hash comparison.
- [ ] When `compare_columns` is declared, the runner fetches both raw states (trial + golden) client-side and hashes them locally via `core/hash.compute_stable_hash` after `apply_compare_columns_extras`, so the asymmetric filter has both sides.
- [ ] When `compare_columns` is unset, the fast path (server-side `get_stable_hash`) is preserved.
- [ ] The `SubstrateCoverage` for `state_checks.compare_columns` in `core/grading/key_manifest.py` moves from `CORE_ONLY` to `BOTH_SCORE_PARITY`; `_DRIFT_EXEMPTIONS` entry removed.
- [ ] Differential test in `tests/canonical/` proves the same declaration produces the same score on both substrates.

## Acceptance Criteria
- Cross-substrate differential passes.
- A synthetic native-adapter pack that declares `compare_columns` and includes a permitted-extra param passes state-hash grading on the runner path.
- No regression in the fast-path performance for packs that do not declare `compare_columns`.

## Out of Scope
- Any change to the shape or semantics of `ColumnCompareRule` / `apply_compare_columns_extras`. Those land in #1548.

Part of #1546.

CiroGamboa:
Runner-side wiring landed as part of #1560.

PR #1560, opened 2026-09-09
feat: arena v3 engine residuals — tool-name hint, state-hash compare_columns (both substrates), trace_checks gate advisory

### LINKED ISSUE #1547: feat(runner): rescue near-miss tool names in dispatch so a wrong short name is recoverable

### LINKED ISSUE #1549: feat(grading): decide the default for trace_checks on_missing on an unmatched anchor
Part of #1546

## Context

Version 0.21.3 added `on_missing: withhold` semantics to trace_checks so that an unmatched constraint anchor can be excluded from the numerator and denominator instead of being counted as a failure. That semantics is author opt-in, per constraint. Today, an unmatched anchor — including the shape where the tool the anchor names errored silently — still defaults to `on_missing: fail`. When the constraint is also marked `severity: gate`, this collapses the whole trace-checks component to zero. The audit demonstrated the shape live: a weight-1.0 gate discarded 97% of otherwise-passing weight because the anchor's tool had errored. No arena v3 pack currently uses trace_checks, so this does not affect the v3 board directly, but it is a landmine for any future board that adopts trace-checks-based grading.

## Deliverables

- [ ] The chosen behavior implemented in the trace_checks evaluator.
- [ ] Engine unit tests cover: matched anchor; unmatched anchor with authored `on_missing: withhold`; unmatched anchor with authored `on_missing: fail`; unmatched anchor with no author opt-in.
- [ ] Grading rubric authoring documentation reflects the chosen behavior.

## Acceptance Criteria

- A synthetic pack fixture with a `severity: gate` constraint whose anchor is a tool that errors behaves the chosen way — the component collapses to zero under option A (keep default and warn) with no authoring, or is excluded from the score under option B (flip default).
- Existing packs that use `on_missing: withhold` explicitly are unaffected.
- No arena-side smoke needed — no arena pack currently uses trace_checks.

## Design Decisions

None resolved yet — see Open Questions.

**Which default do we want?**

- **Option A — Keep default FAIL, add authoring warning.** The default stays; the grader-authoring gate emits a warning when a `severity: gate` constraint on an anchor whose tool can error is authored with no `on_missing: withhold`. No back-compat break; costs an authoring-gate change.
- **Option B — Flip default to WITHHOLD.** Cleaner semantics — a silent tool error no longer zeroes the component. Arena packs carry no trace_checks so nothing breaks in v3, but any external consumer of trace_checks that relied on the FAIL default silently changes verdicts.
- **Option C — Defer out of this milestone.** File as a standalone follow-up so the parent milestone stays tight around the two v3 publish blockers.

**Leaning:** A or C — the arena v3 board is unaffected either way, so this is a policy call, not a publish blocker.

**Open question for the team:** are there external trace_checks consumers we know of who rely on the current FAIL default? If not, B is materially cleaner and this becomes a defer-to-follow-up-or-flip-now call.

## Out of Scope

- Any change to how trace_checks composes with other grading components. This issue is scoped to the default value of one enum on one constraint field.
- Retrofitting `on_missing: withhold` into existing packs. That is a pack-side authoring change, not an engine change.

CiroGamboa:
## Summary

Consolidates the arena v3 engine residuals milestone into a single PR against `main`. Four independent engine changes, all aligned with the team feedback on the umbrella #1546:

- **TD-25 tool-name hint** (safety net) — `_resolve_tool_or_refuse` enriches `Tool not found` with up to 5 registered names whose suffix matches the requested short name, so a model calling the natural name of a doubled-prefix registration (`<system>_<tool>_<tool>`) can retry against the real name on the next turn. Skips unusable tools; signals "and N more" when truncated. Log warning carries the same candidates.
- **TD-24 state_checks.compare_columns** — new per-(table, column) config `{mode: subset, extras_allowed_for: [...]}` that lets the state-hash comparator accept model-added keys the pack's SOP declared permitted (e.g. an optional notification param). Mirrors the trace comparator's `compare_args` shape. Keys the golden declares are still compared value-for-value; extras outside the allowlist still fail; symmetric `unstable_fields` masks compose cleanly.
- **compare_columns wiring on the runner-side hash path** — `_execute_hash_grading` detects a non-empty `compare_columns` and switches from server-side `get_stable_hash` to a client-side raw-state fetch + local hash so the asymmetric filter has both sides. Fast path preserved when unset. Substrate parity restored — `state_checks.compare_columns` moves from CORE_ONLY / drift-exemption to BOTH_SCORE_PARITY.
- **N9 trace_checks authoring advisory** — grading-authoring gate emits an advisory when a `severity: gate` constraint uses the default `on_missing: fail`, naming the risk that a silent tool error on the anchor collapses the whole trace_checks component. Advisory only, no hard error — a gate whose anchor is a deterministic action is a legitimate shape. Default runtime behavior unchanged.

## Design choices, each aligned with team feedback

- **TD-25**: Option A (hint enumerating candidates). No silent alias — dispatching a name the transcript doesn't show would diverge the transcript from the registered tool and hide the pack bug in logs. The load-bearing primary fix is at the pack source (double-prefix in the pack `tool_registry`), tracked as a separate GH issue on the internal repo; this engine change is the defensive safety net.
- **TD-24**: per-column `{mode: subset, extras_allowed_for: [...]}`, mirrors trace-side `compare_args`. Long-term this replaces hand-maintained `unstable_fields` masks for permitted-extra params — the pack-side migration lives as its own internal-repo GH issue.
- **N9**: chose the FAIL-default + authoring-warning path over flipping the default to WITHHOLD — the warning keeps back-compat with any external consumer relying on the FAIL default, and surfaces the risky shape to authors at gate time so it doesn't silently zero components at runtime.

## Test plan

- [x] `pytest tests/unit/grading/ tests/unit/test_runner_pipeline.py tests/canonical/test_grading_substrate_parity.py` — 2,320 tests pass locally.
- [x] New tests: 3 for `nearest_tool_names` (hint, ambiguity, no near-miss); 2 for review-round hardening (skip unusable, signal truncation); 9 for `apply_compare_columns_extras` (no-op, drop-permitted, extra-not-allowed still fails, value-mismatch still fails, composes with unstable_fields, non-dict / missing / list / dict container shapes); 1 for the N9 authoring advisory (severity:gate + defaulted on_missing surfaces).
- [x] Substrate parity: `state_checks.compare_columns` now BOTH_SCORE_PARITY; drift-exemption entry removed; adapter translation test passes (`test_adapter_translation_carries_every_runner_key`).
- [x] Canonical baseline for `test_harness_registry_replay` regenerated on the integration branch; only remaining CI red is the pre-existing repo-wide `#1234` (main hits it too).
- [x] `ruff check` + `ruff format --check` + pre-commit hooks clean.

## Consolidates + closes

- Supersedes PR #1551 (folded in).
- Supersedes PR #1552 (already merged into this branch).
- Closes #1547, #1548, #1558.
- Reopens + closes #1549 (N9 landed as advisory).

Part of #1546.


CiroGamboa:
Addressed all 8 findings from `/code-review` (extra-high) in `4de5cc5b`:

- **F1 + F2** — Slow-path (`compare_columns`) diff step now reuses the raw states already in memory (skips two `get_stable_state` RPCs + one `restore_snapshot` round-trip on every hash mismatch). The diff shown to the author is now computed against the **filtered** trial state, matching what the hash verdict was computed against — permitted-extra keys no longer leak into the diagnostic.
- **F3** — N9 severity-gate advisory now only fires on constraints whose `require` tree contains an anchored kind. Skips `present` / `absent` / `count` (anchorless kinds where "the anchor's tool" doesn't exist) so the advisory reads truthfully. Test switched to a `before` shape so it still provokes the rule.
- **F4** — Truncation wording changed from "…and N more registered with the same suffix" to "…and N more registered name(s) whose suffix or infix matches — cap X", since rank-1 (infix) matches can be hidden too. Existing test assertion `"and 2 more"` still holds.
- **F5** — Slow-path guard tightened from `bool(compare_columns)` to `any(column_rules for column_rules in compare_columns.values())`. An inert `{"table": {}}` declaration no longer forces the extra-RPC path.
- **F6** — `key_manifest.py` runner_evaluator reference moved from the private `._execute_hash_grading` to the public `.GradeTrial` entry point, so a future rename of the internal helper doesn't stale the manifest reference.
- **F7** — Docstring on `apply_compare_columns_extras` now names the shallow-copy semantic explicitly: tables absent from `compare_columns` share references with the input; callers that hash-then-discard (the only supported use) are unaffected, callers that would mutate must not.
- **F8** — Dropped the `list(numeric_string_fields or [])` allocation in the slow path; `compute_stable_hash` accepts the raw iterable and materialises it once as a frozenset itself.

All 511 tests in the affected suites pass locally. Ruff + black + pre-commit clean.

PR #1597, opened 2026-09-15
feat(observability): a TrialObserver seam, the tolokaforge-langfuse wheel, and write-once tracing on a v4 receiver

balazs-toloka:
## What

Live tracing of trials (ADR-0047, `docs/adr/0047-live-tracing-trial-observer-otel.md`, **Proposed**, seven amendments; the section "Where the decision stands" at its top summarises the current shape) and, for a receiver whose observations are append-only, the **write-once layout** (ADR-0048, `docs/adr/0048-write-once-observations-append-only-receiver.md`).

The engine gains a **`TrialObserver` seam** and nothing receiver-specific. The Langfuse observer ships as a separate wheel, **`tolokaforge-langfuse`** (a workspace member of this repository, released on its own `langfuse-vX.Y.Z` cadence), which the engine finds through an entry-point group. A trial traced live and the same trial uploaded later from its bundle by the offline connector are one and the same trace, so a deployment whose engine traces live needs no upload pass, and a receiver-side fix reaches a deployment without moving its engine pin.

### Engine: `tolokaforge/observability/` (no third-party dependency)

- `observer.py`: the `TrialObserver` Protocol (`trial_started`, `generation`, `tool_call`, `trial_finished`, `trial_persisted`, `run_finished`), `TrialIdentity`, the null and composite observers, `LoopObserverBinding`, `safely()`. Every hook is wrapped: nothing raises into a trial, configuration errors surface at run start before any service starts.
- `ids.py`: the id contract (version 2) shared with the offline connector: `uuid5` over run tag, run id, task id, trial index and attempt for the trace; observation and score ids derived from it, so a re-send from either producer updates instead of duplicating. The contract also names the **preview kinds** (`proot`, `pgen`, `pugen`, `ptool`, `pjgen`, `pjtool`), the twins the write-once layout writes while a trial runs; no existing id moved.
- `factory.py`: `build_trial_observer` resolves the run identity (config, `TOLOKAFORGE_TRACING_RUN_ID` / `_RUN_TAG`, else the engine run id), asks every plugin registered under **`tolokaforge.trial_observers`** (`build(tracing, identity, *, engine_run_id, output_dir) -> TrialObserver | None`) and composes the answers. `exporter: otlp`, or a `*_TRACING_ENABLED` switch that is on, with no observer produced, or a plugin that cannot load, is a `TracingConfigError` at run start. Writes `run_identity.json` (with the engine version) and `tracing_receipt.json`. `PLUGIN_API_VERSION` names the contract a plugin checks against; it is **4** on this branch (3 moved the receiver's fields out of the engine config, 4 is the write-once layout), and an engine and a plugin of different versions refuse to pair.
- Wiring: `ToolCallingLoop.observer` (generation after the assistant message is recorded, tool call after the tool message), `TrialRunner(loop_observer=...)`, `InProcessConductor` (`trial_started`, `trial_finished` after grading, `trial_persisted` once nothing writes into the trial directory any more), `Orchestrator` (observer per run, receipt at the end). Loop and runner import the observer type under `TYPE_CHECKING` only; the runner subset is unchanged.
- Config `observability.tracing` (`TracingConfig`) is **receiver-neutral** (ADR-0021, plugin API 3): `exporter`, `endpoint`, run identity, `session_id`, `label`, service name, `tags`, `metadata`, the queue, flush and span-content limits, and `options`, a mapping keyed by plugin name that is passed through unchanged. Everything receiver-shaped lives under `options.langfuse`, owned by the wheel's own strict schema: `expect_project`, `attach`, `gradings`, `projection`, `attach_api_base`, `attach_timeout_s`, `attach_budget_s`, `profile`, `environment`, `model_name_normalizer`, `model_name_rules` and `server_api`. Unknown keys and old top-level receiver fields fail at run start. Receiver credentials never live in the config.
- `Trajectory.attempt_id` (additive, optional) so a bundle upload derives the same trace id.
- The `otel` extra resolves to `tolokaforge-langfuse`; the import-linter contract `trial-observer-seam` keeps the engine from importing the plugin statically.

### The plugin wheel: `tolokaforge_langfuse/`

- `plugin.py` (the entry point): enablement (`exporter: otlp`, or the one switch `LANGFUSE_TRACING_ENABLED` with `LANGFUSE_BASE_URL`, `LANGFUSE_PUBLIC_KEY` / `LANGFUSE_SECRET_KEY` read through the `SecretManager`, optional `LANGFUSE_EXTRA_HEADERS` and `LANGFUSE_PROJECT`); the receiver from the standard `OTEL_EXPORTER_OTLP_*` variables when a launcher owns it; the fail-closed project check (`expect_project` against `GET /api/public/projects`: a mismatch or a 401 refuses before any service starts, an unreachable check proceeds `unverified`); the deployment profile; the contract-version check against the engine, run only when the run is going to trace.
- `otel.py`: finished `ReadableSpan`s with the contract's ids and the trial's own clocks, the Langfuse OTel attribute conventions, a bounded and counted `SpanQueue` (drop instead of block), flush and shutdown budget, `SensitiveKeyRedaction` on tool arguments, attributes capped, `langfuse.environment` / `release` / `version` on every span (the receiver fixes a trace's environment at the first write it sees).
- `attachments.py` and `media.py`: at `trial_persisted`, every regular top-level file of the trial directory as a media object on the trace (manifest v2 in the trace metadata), after a data-safety scan against the `SecretManager`'s credential values; per-thread budget and a breaker; the ingestion route for the trial-end pass.
- `projection.py` and `gradings.py`: the trial-end pass completes the trace from the bundle through the ingestion API under the shared ids: the trace body and its metadata, the root, the generations with paired usage, the tool executions, the simulated user turns, the grading with its judge transcript and scores plus the trace-level mirror, the events, media for base64 image blocks. The live spans are the preview, the bundle projection is the truth. The offline connector's projection is the reference; a golden parity test over a synthetic bundle is committed byte-identically in both repositories. The trace metadata is a fixed 34-key schema plus the caller's keys, because Langfuse's trace page renders its metadata table only up to 100 top-level keys.
- `profile.py`: the deployment profile (TOML, schema 1): the native `environment` as a literal or a rule over one tag prefix, fixed tags and metadata, the profile version (joins the native `version` field), optionally the model-name rules file. `python -m tolokaforge_langfuse.profile <file>` validates one; `TOLOKAFORGE_TRACING_PROFILE`, `TOLOKAFORGE_TRACING_METADATA` and `LANGFUSE_ENVIRONMENT` are the launcher's variables.
- `model_names.py`: model identity as configuration (raw `provider/name`, or `toloka-model-name-normalizer` with a deployment's rules file, loaded only when selected).
- Release: `release-langfuse.yml` (`cz bump`, tag `langfuse-vX.Y.Z`) and `publish-tolokaforge-langfuse.yml` (trusted publishing), mirroring the models wheel; `docs/RELEASING.md` has the section and the contract-version rule. No `tolokaforge` requirement in the wheel, by design (the models wheel's reasoning, ADR-0030).

Documentation: `docs/OBSERVABILITY.md` (configuration, the trace, attachments, the trial-end pass, the trace metadata schema, the deployment profile, packaging, the export guarantees), `tolokaforge_langfuse/README.md`, ADR-0047 with its amendments.

## Verified

- Unit: the seam's tests (`tests/unit/observability`) and the plugin's (`tolokaforge_langfuse/tests`) are green on the branch head (268 tests together, 5 of them new for the write-once delivery and the family probe); CI's `test-smoke` runs the full tier on every push to this PR and is green on the head: **11216 passed, 24 skipped** (`unit or canonical`, with coverage), plus the tool suites (7 / 36 / 12), the external adapters (11 passed, 2 skipped) and the Docker task-pack smoke. ruff, black, mypy (only the pre-existing `yaml` stub notes), import-linter (the `trial-observer-seam` contract included), actionlint.
- The wheel builds (`uv build --package tolokaforge-langfuse`), carries the entry point and declares no `tolokaforge` requirement; `pip install 'tolokaforge[otel]'` resolves to it once it is on PyPI.
- Against a Langfuse test project: runs traced live by this branch under a deployment profile, then checked by the offline connector's `verify-parity` (download and byte-check of every attached file, dry-run projection of the same bundles, field-by-field comparison of observations, scores, tags, metadata and native fields): PASS, 22/22 files byte-equal, no difference outside the documented exclusions (the producer bookkeeping keys, the live root span's own keys, the receiver's OTLP metadata keys `attributes` / `resourceAttributes` / `scope`); a later grading attached by the connector to the never-uploaded run lands next to the live one. The trace page renders the metadata table again (41 keys).
- On a local Langfuse **4.38.0** sandbox, the write-once layout end to end: a run traced live and read back through the v2 / v3 APIs has, per trace, exactly the final observation ids the bundle projects (23 and 22), the previews beside them (14 and 13, every one marked and named), **one** root, **zero** duplicate ids, 35 scores of which 18 are `scope: primary`, the 18 tags and the manifest parsed back as an object, and `download` restores 11/11 files byte-exact; the offline connector's `verify-parity` is **PASS 2/2, 0 findings** against the same run. A second upload of an archived bundle writes no observation and reports what it skipped as already present.
- On a Langfuse **3.205.1** instance, the same commands through the v3 family path: `verify-parity` **PASS 2/2**, the archived bundle round-trips **11/11 byte-exact**, later gradings behave as before, and the live run's write-once counters are all zero, so that family's traffic is unchanged.
- Receiver facts learnt on the way and designed around: a trace's `environment` is fixed at the first write; observation and score bodies without `environment` are filed under `default`; the metadata table shows nothing above 100 keys; costs are stored with 12 fractional digits; tags merge as a set. On 4.x: a trace **is** its root observation; `langfuse.trace.input` / `.output` are accepted and stored nowhere, so the root's input and output travel as observation attributes; the receiver's own OTLP artefacts are sixteen flat dotted keys (`attributes.*`, `resourceAttributes.*`, `scope.*`) rather than the three nested objects of 3.205.1; and a categorical score comes back from 3.205.1 as a numeric placeholder beside `stringValue`, which is one rule in the reader, not two code paths.
- Independent diff reviews of every increment, findings addressed in follow-up commits (bounded flush, exception-path trace closing, headers through the `SecretManager`, per-thread attachment budget, the data-safety scan inside the guarded block, the project check refusing a 401, a switch without an observer refusing the run start, a grading pass leaving the attachment manifest to its owner on the connector side). The review of the write-once increment found **two real defects in the delivery path**, both fixed here with a test that fails against the code as it stood: the exporter's own retries could write a batch the receiver had already stored, and an ambiguous export failure made the run write a second root under the same id. The same round also gated the ingestion header and the single-post exporter on the v4 family, made the capability probe require that API's own answer, fed the attachment breaker from the only call that waits for the receiver, and ended an error root at the trial's own end instead of at run end.

### The trace vocabulary and the one profile (vocabulary amendment)

- `tolokaforge_langfuse/vocabulary.py` is the default trace vocabulary of both producers (the offline uploader imports it): the tag prefixes with their producer / caller split, the value lists the engine's output format defines (`run_kind`, `scope`), the tag order, the default environment rule (`production` for `run_kind:eval`, `development` otherwise), and the tags derived from the bundle and the model-name normalizer: `model_generation`, `model_tier`, `model_variant`, `model_size`, `model_stage`, `model_snapshot` when the rules derive the facet, `reasoning_mode`, `reasoning_effort`, `reasoning_budget` from `task.yaml` `model_config.agent.reasoning`, and `route` (the agent's configured provider). Nothing is invented for an absent fact.
- The deployment profile is schema 2 (schema 1 still loads) and is the one file both producers read: `[environment]`, `[tags] fixed / derived / values / required`, `[derive.<prefix>]`, `[metadata] keys / fixed`, `[models] rules`. The plugin refuses, at run start, a tag outside the vocabulary, a value outside a closed list, a contradiction of a fixed tag, a missing required tag under a profile and a metadata key outside the profile's list; `python -m tolokaforge_langfuse.profile <file> --tags ... --metadata ...` lets a launcher pre-check its inputs and degrade instead of failing a run.
- The live producer sets `source:trial` itself; the native `version` of a live trace names the wheel (`tolokaforge-langfuse-<version>+<rules>+<profile>`), `release` the engine. The shared modules import no engine module (a test pins it), so the offline uploader depends on the wheel next to any engine pin.
- Verified on a test project: a run traced live under a deployment profile, checked by the offline connector's parity command: PASS, every attached file byte-equal, the 18 tags identical on both sides.

### The write-once layout on a receiver whose observations are append-only (ADR-0048)

Langfuse 4.x in its default write mode takes observations over **OTLP only**, stores them
**append-only** (a re-sent id is a second row, there is no read-time dedup and no way to delete one
observation) and makes a trace **be** its root observation. The provisional root plus the
re-sending trial-end pass above cannot work there, so the observer detects the receiver's family
once per run and writes accordingly. A v3 receiver keeps its wire traffic exactly as before.

- **The family is a capability, not a version.** `GET /api/public/v2/observations` answers on a 4.x
  receiver in every write mode and 404s on a 3.x one, while the reported version says `4.x` in the
  transitional modes too. The probe is read-only, runs once at run start next to the project check,
  and takes a 2xx only when the body is that API's own answer, so an authenticating proxy's login
  page cannot switch a run's layout. `options.langfuse.server_api` (`auto` / `v3` / `v4`) overrides
  it; an unreachable receiver leaves the run on the v3 family. The answer lands in the receipt.
- **While a trial runs** the observer writes declared **previews**: a preview root whose parent is
  the final root's id, and the live bodies under the preview kinds beneath it, each named
  `preview: ...` and marked `preview: true`. A reader excludes them by the marker and by the id the
  contract derives. Until the trial ends the trace has no root row, so a running trial is reached by
  its (deterministic) trace id or by its session and joins the trace list when it ends.
- **At `trial_persisted`** the bundle's projection is converted to OTLP spans by
  `tolokaforge_langfuse/otlp_spans.py` (engine-free, byte-identical to the offline connector's copy,
  pinned by a span-attribute golden in both repositories) and written **once**, the **root last**,
  with the attachment manifest in the root's metadata as a JSON string the receiver parses back.
  `projection: full` is required on this family, since the root comes from the bundle.
- **The verdict lives in the scores**, because the trace's metadata is frozen at that single write:
  the trace-level mirror is joined by a categorical `primary_grading` score naming the grading it
  mirrors, both marked `scope: primary` in the score metadata, which is what a dashboard filters on.
  Two deliberate changes reach the v3 family as well: this pointer score, and a score carrying the
  grading's own timestamp instead of the upload's.
- **Delivery becomes at-most-once on the wire.** The stock OTLP exporter re-posts a batch up to six
  times on a connection error or a retryable status; on this receiver the batch that was written and
  lost its answer would be written again, root included, undeletably. The v4 family therefore posts
  each batch once, and because that single post is the safety argument itself, an OpenTelemetry SDK
  whose exporter cannot be asked for it fails the run at start rather than degrading to a retrying
  one. A span the queue never took still gets its **error root** at run end; a posted one the
  exporter could not confirm gets a warning and the `langfuse.roots_unconfirmed` counter instead,
  because a second root under the same id could never be removed.
- **The receipt** gains `langfuse.previews_sent`, `langfuse.final_observations_sent`,
  `langfuse.error_roots_sent` and `langfuse.roots_unconfirmed` under `extra`, and the family under
  `details`. The first three count spans **queued**; what left is `spans_exported` / `spans_dropped`.

## The design review's asks, and where each is answered

@CiroGamboa's review of 2026-09-18 gated the merge on four items; all four are in the branch, and
the three minor findings and the two design notes with them.

| Ask | Where |
|---|---|
| 1. Delete the out-of-repo `PLAN 3.x` attribution from shipped comments and docstrings (5 sites) | gone; the substantive cross-references stay |
| 2. ADR-0011 Pattern A: an `InMemoryTrialObserver` fixture and a canonical contract test | `tolokaforge/observability/observer.py` (`InMemoryTrialObserver`, call log and failure knobs), `tests/canonical/test_trial_observer_contract.py` with snapshots for the hook surface, the receipt, the run identity and the config |
| 3. ADR-0011 Pattern B: `TracingConfig` to `extra="forbid"`; `ExportReceipt` per Pattern B and the ADR-0021 transport-neutral trend | `TracingConfig` and the plugin's own `LangfuseConfig` both forbid unknown keys; `ExportReceipt` is a frozen, forbidding Pydantic model with generic counters plus an opaque `extra: dict[str, NonNegativeInt]` the composite merges without knowing the keys, pinned by a wire-shape snapshot |
| 4. AGENTS.md for the new workspace member and `docs/OBSERVABILITY.md`; the secrets static-grep guard over the wheel | both done; `tests/unit/secrets/test_no_raw_secret_access.py` scans `tolokaforge_langfuse/src/tolokaforge_langfuse` |
| Minor 8: `trajectory.attempt_id` assigned after `_produce_grade_bundle` | moved to right after `_run_agent_loop` returns, so the snapshot bundle records the traced attempt too; the duplicate on the except path is gone |
| Minor 9: functions over 100 lines in the plugin | `build_projection` 96, `_agent_observations` 91, `_events` 10, `build_grading_events` 66 |
| Minor 10: past tense in the `projection` field docstring | reworded; the field itself now lives in the plugin's schema |
| Design note: the receipt is per process | stated in the amendment "seam contracts and process-local delivery receipts" |
| Design note: the seam is an in-process tap, not a remote-ingest contract | stated in ADR-0047; the durable cross-process contract is the id contract plus the bundle projection |

Since that review the branch also gained the write-once layout and ADR-0048, so the ADR pass now
covers two decision records.

### The re-review of the write-once layout (`cc8c2a2b..cb5f1fce`)

@CiroGamboa's second round raised one blocker and two majors; all three are in `a026d868`.

| Ask | Where |
|---|---|
| Blocker: the living docs ship plugin API **3** while the code is **4**, and carry migration history | every version corrected to 4 (`docs/OBSERVABILITY.md`, `tolokaforge_langfuse/README.md`); the API 2 and API 3 migration narratives move to `tolokaforge_langfuse/CHANGELOG.md` under `BREAKING CHANGE`, where upgrade guidance belongs. ADR-0047 no longer names a number at all, so a bump touches one living doc and the code |
| Major: the at-most-once guarantee degrades silently on an OTel SDK upgrade | `make_otlp_exporter(retry=False)` raises the new `SingleAttemptUnavailable` instead of returning the stock retrying exporter, and `plugin.build` turns it into a `TracingConfigError` at run start, where the family is already resolved. Two tests lock it (the raise, and the v3 run that is deliberately unaffected because it upserts), two more lock the plugin's translation. The trade is documented in `docs/OBSERVABILITY.md` and in ADR-0048's consequences |
| Major: internal references leaked into public artefacts | the two `docs/TECHDEL-*` plan paths are gone from the ADRs' Links; all 30 `D-v4-*` / `F*` tags are gone from the wheel's source and tests, and the three `step 00` citations with them. The prose rationale each one sat on stays; the provenance is in the ADRs |
| Non-blocking: the comment on `INGESTION_VERSION_HEADER` describes the import above it | moved onto the `SERVER_V3, SERVER_V4` import it actually describes |
| Non-blocking: `otel.py` and `projection.py` hold functions over the 100-line guideline | not in this PR, as suggested: extracting per-record-kind builders is a refactor of its own |

Green after the round: 223 wheel tests (+4), the seam's own suite and the canonical tier locally (2299 passed, 22 skipped), ruff, black; CI's `test-smoke` runs the full tier on the head and passes.

## Not in this PR

- A run-level object (a Langfuse dataset run, or a run trace) for the run's aggregate files; today only the trial directories are attached.
- Deleting the legacy (v3) writer. It stays until no receiver of that family is left; the one open question about deleting it is answered (a 3.205.1 receiver also parses a JSON-string manifest on an OTLP-written root back into an object), but the deletion itself belongs in its own PR.
- Operations before the first release: the PyPI project `tolokaforge-langfuse` with its trusted publisher; `langfuse-v0.1.0` precedes (or accompanies) the engine release whose `otel` extra names it.
- Engine-owner review of ADR-0047 and ADR-0048 is the merge gate.


CiroGamboa (commented):
## Design review — TrialObserver seam (ADR-0047)

Focused on extensibility and detachment: does the seam let other observability backends plug in cleanly, and does it hold up if the engine becomes a distributed system. The import-level detachment (`.importlinter` `trial-observer-seam`), the entry-point discovery reuse, and the deterministic `uuid5` id contract (live/offline convergence, idempotent re-sends) are the strong core and are well aligned with ADR-0030/0036 and ADR-0022. Most findings below are about conforming the new seam to **ADR-0011** (our seam/data conventions) and **ADR-0021** (our existing observability seam), which ADR-0047 doesn't currently reference.

### Blockers

**1. `PLAN 3.x` attribution in shipped comments/docstrings** — delete the `(PLAN 3.x)` tokens (keep the substantive cross-refs).
- `tolokaforge/observability/ids.py:82`
- `tolokaforge_langfuse/src/tolokaforge_langfuse/gradings.py:11`
- `tolokaforge_langfuse/src/tolokaforge_langfuse/projection.py:85`
- `tolokaforge_langfuse/src/tolokaforge_langfuse/projection.py:903`
- `tolokaforge_langfuse/src/tolokaforge_langfuse/plugin.py:178`

Unlike in-repo `ADR-00xx` refs, the out-of-repo PLAN document a future reader can't resolve — exactly the rot AGENTS.md §comment-hygiene / Core Rule 8 targets.

### Major — ADR-0011 Pattern A (seam definition)

`TrialObserver` is a Pattern A seam (`@runtime_checkable` Protocol with a named second variant: other receivers). ADR-0011 makes the following mandatory for such a seam, and is explicit that a Protocol missing them is "a specific reviewable comment, not a subjective preference":

**2. No `InMemoryTrialObserver` fixture.** Every peer seam ships one (`InMemoryArtifactWriter`, `InMemoryRuntimeBackend`, `InMemoryConductor`, `InMemoryJudge`) with a `CallLog` and failure knobs for orchestrator-level injection. This PR ships `NullTrialObserver` (the `_Null*` default sink) + `CompositeTrialObserver`, and the tests use ad-hoc per-file recorders instead (`tests/unit/observability/test_ids_and_observer.py:84`, `test_factory.py:72`). Add `InMemoryTrialObserver` per ADR-0011 §Test-fixture convention.

**3. No canonical contract test.** `tests/canonical/` pins ~24 seam contracts (conductor, runtime_backend, trial_grader, judge, trial_executor, …) but not this one. Add `tests/canonical/test_trial_observer_contract.py` pinning the hook surface, argument shapes, and the `safely()` never-raises / composite fan-out semantics.

### Major — ADR-0011 Pattern B (data declaration)

**4. `TracingConfig` uses `extra="ignore"`** — `tolokaforge/core/models/run_config.py:1111`. ADR-0047 calls this "the seam's public configuration", i.e. a Pattern B shape crossing the YAML boundary, where ADR-0011 requires `extra="forbid"` ("no exceptions"). The class grew from 2 to ~27 fields; with `ignore`, a typo (`expect_projct:`, `attach_budget:`) is silently dropped and the run traces wrong or not at all. Legitimate unknown plugin keys already have the ADR-sanctioned `options` passthrough, which is precisely why unknown top-level keys should be rejected. Sibling leaf configs (`S3StorageConfig`, `LocalStorageConfig`, `GraderConfig`, `SnapshotBundleConfig`) already forbid.

**5. `ExportReceipt` — diverges from Pattern B and from the ADR-0021 observability trend** — `tolokaforge/observability/observer.py:66-132`.
- Pattern B: it's persisted to `tracing_receipt.json` (a serialization boundary) but is a frozen dataclass with a hand-rolled `to_dict()` and no `extra="forbid"` and no canonical wire-shape snapshot test. `to_dict()` also hand-mirrors all ~20 fields, so a new field silently drops from the receipt.
- ADR-0021: our existing observability seam keeps engine types transport-agnostic (`ComponentSnapshot` = id/kind/phase/detail/owner; a future k8s watcher fires the same events). `ExportReceipt` instead bakes one receiver's vocabulary into the engine type (`attachments_*`, `gradings_*`, `manifests_*`, `projections_*`, `media_*`, `expect_project`, `project_verified`). A second backend (Datadog/Jaeger/an internal collector) returns those ~20 fields hardwired to 0, and the composite sums them. Suggest: a Pydantic `forbid` model with generic counters (`spans_*`, `flushed`, `exporter`) + an opaque `extra: Mapping[str, int]` the composite merges without knowing the keys, plus a snapshot test. (`run_identity.json`, also a raw-dict persisted shape, is the same Pattern B family — minor.)

### Major — hygiene

**6. AGENTS.md not updated for the new workspace member.** `tolokaforge_langfuse` is added to `[tool.uv.workspace] members` but is absent from the "Current workspace packages" list (AGENTS.md:345-346), and `docs/OBSERVABILITY.md` is absent from the Detailed Documentation table (AGENTS.md:594). Same-PR doc rule (Core Rule 8).

**7. Secrets static-grep guard doesn't scan the new credential-reading package.** `tests/unit/secrets/test_no_raw_secret_access.py:31` `SCANNED_ROOTS = (PKG_ROOT, MODELS_PKG_ROOT)`. The current handling is correct (`plugin._secret()` is SecretManager-first), but the plugin actively reads credentials and isn't scanned, so a future raw `os.environ["LANGFUSE_SECRET_KEY"]` would escape CI. Add `tolokaforge_langfuse/src/tolokaforge_langfuse`; note the grep may need extending to catch named-constant reads (`os.environ.get(OTLP_HEADERS_SECRET)`).

### Minor

**8. `trajectory.attempt_id` assigned after `_produce_grade_bundle`** — `tolokaforge/core/conductor.py:543-546`. This does not break trace parity — `build_projection` treats `identity.trace_id` as authoritative and only logs a warning on a bundle mismatch ("the run's identity wins", `projection.py:1080-1088`). But in snapshot-grader mode the snapshot bundle serialises the trajectory with `attempt_id=None`, inconsistent with the primary `trajectory.yaml`. Move the assignment to right after `_run_agent_loop` returns; the duplicate on the except path (`:552`) can then go.

**9. Long functions in the plugin** (>100 lines, AGENTS.md Code Standards 3): `projection.build_projection`, `projection._agent_observations`, `projection._events`, `gradings.build_grading_events`. Extract per-observation-kind builders. Non-blocking follow-up.

**10. Past-tense in a source-of-truth docstring** — the `projection` field docstring (run_config.py) says "(the behaviour before the parity amendment)". Reword to describe the option's current semantics only (§docs).

### Non-blocking design notes (for the ADR-0047 engine-owner review)

- **Run-level receipt vs. address boundary.** `run_finished() -> ExportReceipt` is per-process with no cross-worker merge. Correctness is saved by id determinism, but operational counts won't aggregate across a distributed run. ADR-0038's detachment direction (independently deployable/scalable components surviving an address boundary) is the bar; worth stating the merge story, and it lines up with the "run-level object: not in this PR" note.
- **Seam is an in-process tap, not a transport.** The hooks pass live in-process objects (`GenerationResult`, `Trajectory`, …), so a `TrialObserver` can only live in the engine runtime; a remote/sidecar collector needs an in-process shim. That's fine given the durable cross-process contract is the id contract + bundle projection ("spans are the preview, the bundle is the truth") — worth making that layering explicit in ADR-0047 so the Protocol isn't mistaken for the remote-ingest contract.

Overall: strong seam, and the detachment/id design is right. The asks are to run it through the ADR-0011 Pattern A/B checklist (fixture + canonical contract test + `extra="forbid"`) and to shape the receipt per ADR-0021's transport-neutral trend.

CiroGamboa (changes requested):
Requesting changes. The detailed findings are in my review above. Merge is gated on:

1. Delete the `PLAN 3.x` attribution from shipped comments/docstrings (5 sites).
2. ADR-0011 Pattern A: add an `InMemoryTrialObserver` fixture (CallLog + failure knobs) and a `tests/canonical/test_trial_observer_contract.py`.
3. ADR-0011 Pattern B: `TracingConfig` -> `extra="forbid"`; shape `ExportReceipt` per Pattern B + the ADR-0021 transport-neutral trend (generic counters + opaque `extra`, forbid model, snapshot test).
4. Update AGENTS.md (new workspace member + docs/OBSERVABILITY.md) and extend the secrets static-grep guard to `tolokaforge_langfuse`.

The `attempt_id` ordering, long functions, and past-tense docstring are non-blocking. ADR-0047 itself (engine-owner review) still needs its own sign-off — the two design notes above are for that pass.

CiroGamboa:
### Follow-up (Major, ADR-0021) — keep the engine seam receiver-neutral

Beyond `ExportReceipt` (finding #5), Langfuse-specific vocabulary has leaked into the *neutral* engine surface. ADR-0021 is our precedent here: engine observability types stay transport-agnostic and receiver/transport specifics live entirely reporter-side ("the engine does not care about the transport"; a future k8s watcher fires the same events). The tell is a receiver's name and REST routes appearing in engine code:

- `tolokaforge/core/models/run_config.py:1118-1119` — `expect_project` docstring: "Langfuse: the name listed by `GET /api/public/projects`".
- `run_config.py:1148` — `attach` docstring: "the receiver's media API (Langfuse)".
- `run_config.py:1176` — `attach_api_base` docstring hardcodes `/api/public/otel/v1/traces`.
- `tolokaforge/observability/factory.py:117` — engine error string hardcodes "the Langfuse observer is the tolokaforge-langfuse package".

The deeper point is not the prose: `attach`, `gradings`, `projection`, `expect_project`, `attach_*` are receiver-plane concepts (media attachment, bundle projection, project verification) sitting on the neutral engine `TracingConfig`. Under ADR-0021 these belong reporter-side — configured through the ADR-0047-sanctioned `options` passthrough or the deployment profile — leaving `TracingConfig` with receiver-neutral fields only (`exporter`, `endpoint`, `run_id`, `run_tag`, `tags`, `metadata`, session/label, queue/flush knobs).

Not asking to delete the fields blindly — asking to decide the boundary: which of these are genuinely engine-neutral tracing config vs. Langfuse-receiver config that a second backend (Datadog/Jaeger/internal collector) would never set. For the error string at factory.py:117, list installed plugins generically first (it already does) and drop the hardcoded package name, or gate it behind "if the otel extra names it". This is the config-surface half of the same receiver-neutrality concern as `ExportReceipt`; worth resolving in the ADR-0047 engine-owner pass since it touches the public config contract.

CiroGamboa (changes requested):
## Re-review of the design-review fixes + ADR-0048 (`cc8c2a2b..cb5f1fce`)

Round 1 is resolved — thank you. Verified on the new head: PLAN tokens gone; `InMemoryTrialObserver` (CallLog + `fail_on` knobs) and `tests/canonical/test_trial_observer_contract.py` added; `TracingConfig` `extra="forbid"` with all receiver-plane fields (`expect_project`/`attach*`/`gradings`/`projection`/`profile`/`environment`) moved plugin-side; `ExportReceipt` reshaped to a `forbid` Pydantic model (generic counters + opaque plugin-namespaced `extra` + `details`, `merge()` surfacing a failed observer as `export_failures=1`); AGENTS.md + secrets static-grep updated (and extended to named-constant reads); `attempt_id` hoisted before `_produce_grade_bundle` and locked by a test. The detachment story is now solid: receiver-family logic is entirely plugin-side, and `otlp_spans.py` is provably engine-free (`test_engine_free_modules.py` enforces it).

The ADR-0048 write-once design is sound (capability-based family detection that fails safe to v3, root-written-last, at-most-once on the wire, error roots at run end). Three new items to resolve before merge.

### Blocker — docs ship the wrong current plugin API version + carry migration history
Code is `PLUGIN_API_VERSION = 4`, but the living docs say 3 and contradict themselves:
- `docs/OBSERVABILITY.md:398` — "currently version **3**"; `tolokaforge_langfuse/README.md:19` — "currently **3**", while `README.md:65` says "plugin API **4**".
- Migration history in living docs (belongs in CHANGELOG, Core Rule 8): `OBSERVABILITY.md:43` ("Migration (plugin API 3)…"), `:438` ("API version 2 moves…"), `README.md:63-65` ("Plugin API 3 moved…").

Fix: set every "3" → "4"; delete the migration narratives from the living docs and keep that guidance in `tolokaforge_langfuse/CHANGELOG.md` + the PR body. A reader is currently told to pair a v3 wheel against this v4 engine and hit the exact run-start mismatch these docs exist to prevent.

### Major — the v4 at-most-once guarantee degrades silently on an OTel SDK upgrade
`tolokaforge_langfuse/src/tolokaforge_langfuse/otel.py:118-190`. ADR-0048's safety argument for v4 is at-most-once on the wire, because a duplicate on an append-only receiver cannot be deleted. That is implemented by `_single_attempt_exporter_class()`, which drives the OTLP exporter's private internals (`_export`, `encode_spans`, `_shutdown`). When those internals are absent (`hasattr`/`ImportError` → `None`), `make_otlp_exporter` falls back to the **stock retrying exporter** (up to 6× re-posts) with only a `_log.warning`, and the run proceeds against the append-only receiver — silently reintroducing the undeletable-duplicate hazard the ADR exists to prevent.

Pinned deps mask this today; a future OTel bump trips it. Fix: on the v4 family, if `_single_attempt_exporter_class()` returns `None`, raise `TracingConfigError` at run start (where `make_otlp_exporter(retry=False)` is already called) rather than degrading. v3 is unaffected.

### Major — internal references leaked into public artefacts
- `docs/adr/0048-...:139` and `docs/adr/0047-...:473` cite `docs/TECHDEL-497-langfuse-integration/…` — an internal ticket ID and a plan path that does not exist in the repo.
- 22 `(D-v4-*)` / `(F*)` citation tags are spread across shipped wheel source (`otel.py`, `otlp_spans.py`, `media.py`, `gradings.py`, `plugin.py`); they resolve only against that non-repo internal plan, so they are unresolvable for any reader of the public wheel (comment hygiene, §7a).

Fix: drop the `TECHDEL-*` path and the `D-v4-*`/`F*` tags; keep the standalone prose rationale (most comments already have it) and put decision provenance in the ADR, which already summarises it. (`docs/adr/0043-...:8`'s `TECHDEL-591` is pre-existing and out of scope here.)

### Non-blocking
- Long functions: `otel.py` (1417 lines), `projection.py` (1299) — several over the 100-line guideline. Extract per-record-kind builders in a follow-up.
- `otel.py` near `INGESTION_VERSION_HEADER`: the comment's first sentence describes the `SERVER_V3/V4` import a few lines above, not the header constant it sits on — tighten or move.

### For the record
I dropped a candidate finding on re-verification: the v4 `projection != full` run-start refusal **is** behaviour-locked — `tolokaforge_langfuse/tests/unit/test_plugin.py:462` `test_a_write_once_receiver_needs_the_full_projection` exercises it against a v4-answering probe with `projection='gradings'`. No test gap there.

ADR-0047 and ADR-0048 are both still `Proposed` and pending the engine-owner decision — that remains the merge gate alongside the above.

balazs-toloka:
Thanks for the second pass. All three items are fixed in `a026d868`.

**Blocker, the docs shipped plugin API 3.** Every version corrected to **4** (`docs/OBSERVABILITY.md:397`, `tolokaforge_langfuse/README.md:19`). The migration narratives for API 2 and API 3 moved to `tolokaforge_langfuse/CHANGELOG.md` under a `BREAKING CHANGE` section, which is where upgrade guidance belongs; what stayed in the living docs is the current-state statement (a receiver setting left at the tracing top level, or an unknown key inside the namespace, is an error at config load). One more while I was there: ADR-0047 no longer names a number at all, so the next bump touches the code and one living doc instead of three.

**Major, the silent degradation.** You are right that the pin is the only thing holding this today, and a warning is the wrong answer for a property that is the reason the write is permitted at all. `make_otlp_exporter(retry=False)` now raises the new `SingleAttemptUnavailable` rather than returning the stock exporter, and `plugin.build` turns it into a `TracingConfigError` at the point you named, where the family is already resolved. Four tests hold it down, two of them negative so the v3 path is provably untouched: the raise itself, the same SDK still serving a retrying exporter when `retry` is left alone, the v4 run refused at start, and the v3 run still built. The requirement is written into `docs/OBSERVABILITY.md` ("One POST per batch") and into ADR-0048's consequences.

**Major, internal references.** Both `docs/TECHDEL-*` paths are gone from the ADRs' Links. So are all 30 `D-v4-*` / `F*` tags across `otel.py`, `otlp_spans.py`, `media.py`, `gradings.py`, `plugin.py` and the wheel's tests, plus three `step 00` citations and one `write_mode_probe.json` in the same class that your list did not reach. The prose each tag sat on stays; the provenance is the ADRs. ADR-0043's `TECHDEL-591` left alone as out of scope.

**Non-blocking.** The comment above `INGESTION_VERSION_HEADER` moved onto the `SERVER_V3, SERVER_V4` import it was actually describing. The over-length functions in `otel.py` and `projection.py` I have left for the follow-up you suggested, since extracting per-record-kind builders is a refactor of its own and would bury this diff.

Green on the head: 223 wheel tests (+4), the seam's observability suite and the canonical tier locally (2299 passed, 22 skipped), ruff and black clean, and CI's `test-smoke` over the full tier.

Noted on ADR-0047 and ADR-0048 being the remaining gate; both are still `Proposed` and waiting on the engine owner.

PR #1598, opened 2026-09-16
fix(automation): integrate a model only the gateway serves (rebase of #1064 + fixes)

kostiantynvoiku-web:
Rebase of #1064 onto `main` (86de95fa, v0.25.3), plus two fixes. @bberkes-toloka's commit and authorship are preserved; my changes are separate commits.

## Problem

`integrate-model.yml` cannot finish clean for a model OpenRouter does not carry — the live case is Cohere Command A+ (`azure_ai/cohere-command-a-plus-05-2026`, `route=litellm`). Two independent gaps, neither visible until a run is well under way:

1. **The certificate gate closes on a re-onboarding.** A first onboarding synthesises a candidate certificate gated on the provider key the run already sets. A re-onboarding reuses the **curated** certificate, whose `env_key` names the deployment serving the route (`TF_COHERE_A_PLUS_GATEWAY_LIVE`), and no workflow sets it — so every probe skips and the cleanliness gate reads "capability suite did not run". A transport fact wearing a capability mask.
2. **A gateway-only model can never be priced.** `ensure-pricing` resolves by exact id against the OpenRouter catalog, so a gateway-only id never matches, and `COST_USD_POPULATED` is a CORE capability that `reconcile-cert` refuses as `known_unsupported`. The run either burns its fix-loop iterations on it or ships a certificate asserting a capability measured at zero.

## Change

- **`automation cert-env-gate --model-id <slug>`** prints the variable a curated certificate gates on, when it is unset. Nothing for a synthesised candidate, nothing for a satisfied gate, nothing for an unlisted model; fails loud (exit 1) when the registry cannot be imported. The workflow opens it (`<GATE>=1` into `GITHUB_ENV`) only on the gateway route; on OpenRouter an unset gate is a warning, since nothing there can vouch for it.
- **`ensure-pricing --input-usd/--output-usd`**, wired to a new `pricing` dispatch input (`"<input>,<output>"` per 1M tokens), writes an operator-declared price verbatim. It never overwrites an existing entry, never fetches on that path, and refuses a malformed pair. Without a declaration the miss is still reported and left for a human, which is the only honest answer when no source exists.
- **The finalize prompt** stops hardcoding `env_key="OPENROUTER_API_KEY"` and stops telling the agent to fetch a price from an API that does not carry the model; a new certificate on the gateway route gets the derived gate name `TF_<MODEL_ID>_GATEWAY_LIVE`, so a later re-onboarding finds it again.
- **Docs:** a new "Integrating a model only the gateway serves" section, cross-linked from the existing gateway-only limits paragraph rather than duplicating it.

## Fixes on top of #1064

1. **The registry import was stale.** `cert.py` imported `tests.integration.llm.registry`, a path that no longer exists after the models split; it now goes through the public `tolokaforge.testing.certify` seam, the same one `_load_cert` already uses. Rebased as-is, the command would have failed loud on every run.
2. **The "already set" check reads through `SecretManager`**, not `os.environ` — that is what the certify `live_client` fixture does, so `.env` counts, and no credential is read raw (AGENTS.md secrets rule).
3. **`::error::` goes to stderr.** The workflow does `GATE="$(uv run automation cert-env-gate ...)"`, so an error on stdout would have been swallowed *into the gate variable* instead of reaching the log.
4. **A declared price is exactly two numbers.** The original split on `%%,*` / `##*,`, so `0.8,3.2,4.0` silently dropped the middle field and priced the model at `0.8/4.0`. Now a bash regex requires exactly two, and `declared_entry` refuses non-finite, negative, and `0,0` — an all-zero pair is a missing price, not a free model, which is the rule the catalog path already applies.
5. **The gate reads the *effective* route** (`observation/cost/route.txt`, written by the `.env` step) rather than `inputs.route`. A `route=litellm` dispatch whose proxy secrets are unset falls back to OpenRouter with a warning; it must not then open a gate the run cannot honour. Same source for the derived cert gate name. This is a judgment call beyond the letter of #1064 — say the word and it reverts to `inputs.route` in one line each.

## Tests

```
pytest tools/automation/tests                         321 passed
pytest tests/canonical/test_integrate_model_bucket_flow.py   24 passed
```

Nine new tests: four for the gate (unset named, satisfied silent, unlisted silent, unimportable registry loud on stderr with stdout empty) and five for the declared-price refusals. Ruff + black clean on all five touched Python files; the workflow YAML parses and the step order is Ensure pricing → Open gate → Capability probes.

Verified against the real registry and a temp `pricing.json`: an unset gate prints `OPENROUTER_API_KEY` for `openrouter__openai_gpt-5.4` and nothing once set; a declared price is written verbatim, an existing entry is not overwritten, `--check` passes afterwards, and `0,0` exits 1.

## Scope

`tools/automation` (cert, cli, pricing, finalize prompt, tests), `.github/workflows/integrate-model.yml`, `docs/AUTO_INTEGRATION.md`. No engine or models-wheel code. No model-name branches.

Follow-up worth an issue: the `pricing` input exists only on `workflow_dispatch`, so a gateway-only model requested through the Slack poller still needs a manual dispatch.

Companion: the Cohere Command A+ models-wheel PR, whose live capability evidence this unblocks.


kostiantynvoiku-web:
`test-smoke` is red on three canonical tests that are **already red on `main`** — see main's own run [35087132164](https://github.com/Toloka/tolokaforge/actions/runs/35087132164) from 10:50 today, which fails on the same three plus two more the smoke subset does not run:

| test | assertion |
|---|---|
| `test_grading_substrate_parity::test_manifest_covers_every_declared_config_field` | `StateChecksConfig.auto_mask_clock_columns` has no `GradingKey` |
| `test_lost_trial_attribution` | `empty_completion` is now in the termination list |
| `test_run_aggregate_models_snapshot` | same, one extra `TerminationReason` |

They look like canon that #1594 (arena v3 batch 2 — auto-clock-mask, `empty_completion` routed to INFRASTRUCTURE_ABORT) moved without regenerating. Nothing in this PR touches grading, the runner, or the engine at all — the diff is `tools/automation`, `integrate-model.yml` and `docs/AUTO_INTEGRATION.md`. `tools/automation/tests` is 321 passed locally, and `lint`, `hygiene-review`, `wheel-install-gate`, CodeQL and both CodeQL analyses are green here.

cc @CiroGamboa in case the canon regeneration is already in flight.

balazs-toloka (approved):

Issue #934, opened 2026-08-07
feat(llm): new extension slots (assistant_text, params, narrow message_assembly); slot values become {name, params}


CiroGamboa:
## Context

The current six-slot policy registry does not cover three lifecycle stages models empirically need to adapt: assistant-text reshaping (which today blocks the open Cohere integration #929 — the model wraps responses in `<|START_TEXT|>…<|END_TEXT|>` markers with no hook to reach them), generation-param extensibility (every new knob is a `GenerationParams` signature change), and configurable filler injection (the filler string at `client.py`'s empty-assistant path is hard-coded and caused a Gemini regression when tuned once for one model). ADR-0030 § Extension-point surface adds three new slots and reshapes slot values to be configurable.

## Deliverables
- [ ] `AssistantTextPolicy`, `ParamsPolicy`, and narrow `MessageAssemblyPolicy` base classes registered in `_POLICY_REGISTRIES`
- [ ] Cohere-style assistant-text-marker handling is expressible as a `AssistantTextPolicy` subclass (unblocking #929)
- [ ] Generation-param knobs flow through `ParamsPolicy` without editing `GenerationParams.__init__`; `_VALID_PARAMS_KEYS` inspection goes away
- [ ] Preset slot values accept both bare `name` (legacy) and `{name, params}` (new) for one-release compat
- [ ] Overlay validator rejects unknown keys *inside* preset blocks with the allow-list = engine keys ∪ models-declared keys
- [ ] Narrow message-assembly slot lands: the current `inject_empty_assistant_filler` string becomes policy data alongside its flag

## Acceptance Criteria
- The `#929` Cohere blocker can be closed with a new policy subclass and a preset entry — no engine code beyond what this PR ships
- Overlay files with a typoed nested key fail loud with the offending file + closest match; overlay files carrying a legitimate models-declared knob load silently
- Every existing preset continues to load without modification (bare `name` still accepted)
- Gemini's `gemini_drop_placeholder_signature` case becomes ordinary params flow — no more `if reasoning_name == "gemini" and ...` special case in the loader

## Design Decisions
- **Extension points receive enough context to express the adaptation** (ADR requirement 2). `AssistantTextPolicy.parse_assistant_text` gets the full text plus model config, not a pre-digested slice.
- **`{name, params}` is additive.** Bare `name` keeps working across the one-release window so existing overlays don't churn.
[...]

PR #1599, opened 2026-09-16
feat(models): integrate Cohere Command A+ (azure_ai/cohere-command-a-plus-05-2026)

kostiantynvoiku-web:
Integrates **Cohere Command A+** (`azure_ai/cohere-command-a-plus-05-2026`) into the models wheel: preset, certificate, policies, pricing. Mergeable successor of #1124 (the auto-resolved integration on `test/cohere-split-r1`) plus @bberkes-toloka's review fixes from `fix/cohere-integration-review` (c0b24eb5), rebased onto today's `main`, with the canon moved and the one policy that run could not produce.

The model is served only by a LiteLLM gateway's `azure_ai` route (Azure AI Foundry, Preview), not by OpenRouter. Everything model-specific is data or a models-wheel class; **no engine behaviour changes** — the only file touched under `tolokaforge/core/llm/` is a docstring.

## What ships

| Piece | Where | Note |
|---|---|---|
| Preset `cohere_command_a_plus_05_2026` (dated route) | `data/model_presets.yaml` | `cohere_recursive` / `scalar_array_dict_map` / `openai_summary_replay` |
| Preset `cohere_command_a_plus` (family) | same | `strict` / `array_dict_map` / `openai` |
| Certificate `openrouter__azure_ai_cohere-command-a-plus-05-2026` | `certificates/registry.py` | 18 required / 5 known_unsupported (23/23 declared); gated on `TF_COHERE_A_PLUS_GATEWAY_LIVE` |
| `CohereRecursiveSchema` | `policies/cohere.py` → `schema_sanitizer.cohere_recursive` | cycle-tolerant `$ref` inlining, Gemini's relaxations reset |
| `CohereMarkerAssistantText` | `policies/cohere.py` → `assistant_text_policy.cohere_markers` | strips `<\|START_TEXT\|>…<\|END_TEXT\|>`; no-op on clean text |
| Pricing row | `data/pricing.json` | $0.80 / $3.20 per 1M |

Both presets also carry `params.param_value_rules.tool_choice.auto: drop` and `params.reasoning_via_extra_body: true`.

## Why these choices

- **`tool_choice` is dropped as data, not worked around.** Cohere's Chat API has only `REQUIRED`/`NONE`; omitting the parameter is its documented "the model is free to choose", i.e. what `auto` means. The `azure_ai` route refuses the parameter before the model is reached (`UnsupportedParamsError`), so every tool probe fails at once without the rule — verified by live A/B on 2026-08-01 and again by three Arena tau smokes on 2026-09-11 (`toloka-partners/tolokaforge-tasks` runs 34571890048 / 34572376077 / 34575599112, 0/3 trials each). The rule is `param_value_rules` (#1110), consulted in `_build_kwargs`; no engine code.
- **`reasoning_via_extra_body: true` is pinned on the preset.** The model was measured under the `providers.openrouter` overlay, i.e. with reasoning in `extra_body.reasoning`. The Arena binds this agent as `provider: openai` — the documented binding for an OpenAI-shaped gateway route, and the only way to scope `LLM_PROXY_PROVIDERS` so the simulator and judge are **not** proxied — which carries no such overlay and would switch the wire to `reasoning_effort`, a parameter this route is just as likely to refuse. Pinning keeps the measured wire format independent of the provider name.
- **The marker stripper ships here**, as `assistant_text_policy.py` promised ("a shipped Cohere policy ships alongside the Cohere preset when the family lands"). The gateway serving the route strips the markers itself since 2026-09-08 (toloka-python #6719), so the class is a no-op in normal operation and is bound as defence in depth: it fails safe if that hook is rolled back or the model is reached another way. Contract pinned in `test_policies_cohere_markers.py`.
- **`recursive_ref_tool_call` is a declared ceiling.** The root-key rename from the auto-resolve run is gone (review finding 1: its gate degraded to "the emitted key ends with the declared parameter name", which would also rename `line_items`→`items` and credit a model that emitted the wrong parameter). The quirk is intermittent per tool, so it is recorded honestly rather than repaired.

## The #1124 review findings

1. Root-key repair over-broad → removed, ceiling declared (c0b24eb5).
2. Cert gated on `OPENROUTER_API_KEY` for a model OpenRouter does not serve → `TF_COHERE_A_PLUS_GATEWAY_LIVE`, now documented in `.env.example`.
3. Handoff manifest misreported the baseline preset → pipeline-side, not in this PR.
4. The marker hypothesis was untestable then (no slot, and `basic_completion` passes on wrapped text) → the slot exists since #934/#1024; the policy ships here with its own tests.
5. Pricing derived rather than declared → kept at the Azure Global Standard list price, which the derivation reproduced exactly (zero-residual fit over 1,317 gateway-billed calls); #1064 adds the declared-price input for the pipeline.
6. **CI never ran on #1124 and nine canonical guards were red** → this branch is pushed with a real token, so CI runs. The guards are moved here: `test_certify_excluded_capabilities` (Cohere first in the implicit-prompt-caching and RE2 ratchet lists), `test_policy_registry_merge` (+2 registrations), `test_policies_reachable` (+2 classes), and `certify_suite_collection/collection.json` regenerated. The flip-byte identity test was deleted on `main` by #1160 and is dropped from the branch (it was the only rebase conflict); `PER_MODEL_SUBCLASSES` is derived on `main` and needs no edit.

One canary flipped exactly as its own comment predicted: `test_registry_ships_passthrough_as_only_entry` said it was "the canary that flips when a Cohere-marker subclass ships". It now pins the honest invariant instead — the **engine** ships only `passthrough`, and every other entry in the merged registry must come from a wheel entry point.

## Verification

Local, on a Windows host with this branch's wheel installed editable:

```
tolokaforge_models/tests/unit                                81 passed
tests/unit/llm + tests/unit/testing_certify                 890 passed
tests/canonical/test_policy_registry_merge.py
tests/canonical/test_certify_excluded_capabilities.py
tests/canonical/test_capability_registry.py                  30 passed
tests/canonical/test_certify_suite_collection.py              1 passed
tests/canonical/test_models_wheel_replay.py                   2 passed
ruff check / ruff format --check / black --check              clean
```

`collection.json` grew by 39 nodeids, all for this model, none removed. `snapshots/models_wheel_replay/metric.json` is **unchanged** from `main` (see the merge note below). The rest of `tests/canonical` cannot run on this host — Docker is unavailable and several modules import `termios`/`pty` — so CI is the gate for those.

Preset resolution smoke via `build_capabilities`: `azure_ai/cohere-command-a-plus-05-2026` under both `openai` and `openrouter` → `CohereRecursiveSchema` / `CohereMarkerAssistantText` / `ScalarArrayDictMapResponse` / `OpenAISummaryReplayReasoningCodec`, `tool_choice auto → DROP`, `reasoning_via_extra_body=True`; `azure_ai/cohere-command-a-plus` → `StrictSchema` / `ArrayDictMapResponse` / `OpenAIReasoningCodec`, same rule and pin.

**Live capability suite — not yet attached.** It needs #1598 (the rebase of #1064) on `main` first: that is what opens the curated certificate's env gate on the gateway route, without which every probe skips and observe reads as "capability suite did not run". Once #1598 lands I will dispatch `integrate-model.yml` on this PR with `route=litellm` and paste the table here. The 2026-08-13 observe/resolve run on #1124 over the same gateway reconciled 23 probed capabilities with 0 violations; the re-run will reflect the demoted `recursive_ref_tool_call`.

## Merge and release

- **Please keep `integrate:` out of the squash subject** (e.g. `feat(models): integrate Cohere Command A+ …`). `test_models_wheel_replay` walks `^integrate: ` commits reachable from HEAD, so an `integrate:` squash would move `metric.json` and red `main` until a follow-up regeneration — and it would also fire `release-models-on-integrate.yml` automatically. The inherited auto-resolve commit was reworded on this branch for the same reason. Release deliberately instead: `release-models.yml` with `bump: minor` → `models-v1.6.0`.
- Two inherited commits carry the author identity `t <t@t>` from the original branch; left alone to preserve provenance.
- Bucket: a human PR that touches `tests/canonical/**` classifies as Bucket B, which is expected for a certificate addition that moves canon.
- The Arena consumes `tolokaforge-models` transitively from the engine's git rev, so it will pick this up at its next engine repin; until then it runs the same preset as an out-of-tree overlay (`toloka-partners/tolokaforge-tasks` #360).

Depends on #1598. Closes the engine half of TECHDEL-357.


kostiantynvoiku-web:
`test-smoke` is red on five tests that are **already red on `main`** — main's own run [35087132164](https://github.com/Toloka/tolokaforge/actions/runs/35087132164) from 10:50 today fails on exactly the same five, and the sets match node-for-node:

```
tests/canonical/test_grading_substrate_parity.py::test_manifest_covers_every_declared_config_field
tests/canonical/test_lost_trial_attribution.py::test_a_lost_trial_is_counted_against_the_run_and_scored_not_at_all
tests/canonical/test_run_aggregate_models_snapshot.py::test_per_task_metrics_round_trip_with_every_outcome_class
tests/unit/test_config_validator.py::TestUnmappedAgentModelReportsInfoNotError::test_unmapped_agent_model_is_not_reported_unable
tests/unit/test_metrics.py::TestInfrastructureAbortsLeaveTheDenominator::test_two_of_four_rate_limited
```

They read as canon that #1594 (arena v3 batch 2 — `auto_mask_clock_columns`, `empty_completion` routed to INFRASTRUCTURE_ABORT) moved without regenerating; none of them touches models, presets, certificates, policies or pricing.

Every guard this PR actually moves is green in CI and locally: `hygiene-review`, `lint`, `wheel-install-gate`, CodeQL, and locally `test_policy_registry_merge`, `test_certify_excluded_capabilities`, `test_capability_registry`, `test_certify_suite_collection`, `test_models_wheel_replay` (30 + 1 + 2 passed), `tolokaforge_models/tests/unit` (81 passed) and `tests/unit/llm` + `tests/unit/testing_certify` (890 passed).

cc @CiroGamboa.

balazs-toloka:
Not a blocker for this PR, and the gate is right for the mechanism as it stands today: `env_key`
is a single variable name that `fixtures.py` resolves through the SecretManager, so it can only
answer "is this name set", never "does this deployment serve this route".

Worth a follow-up though, because the cost is one switch per gateway-only model and that list only
grows. Could you look at whether the flag can be derived instead of declared?

Since gateway route resolution landed, the engine can answer the real question at runtime:
`fetch_gateway_catalog` plus `resolve_gateway_route` tell you whether *this* gateway serves *this*
route. That is strictly stronger than a hand-set variable. It cannot be set wrongly, it needs no
maintenance when a route is added or removed, and it closes exactly the hole the flag is defending
against: a contributor pointing `LLM_PROXY_BASE_URL` at their own proxy would open the gate today,
every probe would hit model-not-found, and one ratchet accepts that 400 as proof its quirk is
intact, so the run reports green having tested nothing. A catalog check skips instead.

Cost looks small: one request per process on hosts that actually have a gateway configured, and
hosts without one return early from `resolve_proxy_config` before any I/O.

One thing the flag buys that a catalog check would not, worth deciding deliberately: it is also an
opt-out. Someone with the right gateway who does not want to spend on live calls simply leaves it
unset today.

balazs-toloka (approved):

PR #1616, opened 2026-09-18
feat(coding-harness): meter opencode, grok-build and gemini-cli, and refuse a zero that was never measured

CiroGamboa:
Follow-ups to #1593, all found by running the seven shipped harnesses against real Arena tasks after that PR's work landed. Three commits, each revertable alone.

#1593 made harness trials report their cost. Running every harness showed two of them still reported nothing, one reported a number that was wrong by 4.6×, and one could not run at all. This closes those.

## 1. `opencode` reports its cost, and a wrong price flags the trial it priced

**`opencode` reported nothing.** Its stream carries a `step_finish` per step with that step's `tokens` and `cost` — 22 of them on the trial that surfaced it — and no parser had been written for it. `tokens.input` is the non-cached remainder (a recorded step: `input=1`, `cache.write=387`, `cache.read=15623`, `output=304`, `total=16315`), so the parser folds the cache counters back into the inclusive prompt total, as the `claude-code` parser does. A reader passing `input` through would price a cache-heavy trial as though it had almost no prompt.

**`cost_cache_rate_fallback` never fired on a harness trial.** It was set only on the engine's own request path — but a harness trial makes no engine request, so the one field whose job is to say "this cost is overstated" stayed `False` on a live trial that overstated by **4.6×** ($0.884 against opencode's own $0.222; 5.4× on a later run). The preflight half of that signal worked all along; the after-the-fact half was wired to a path harness trials never take. Now set where both harness pricing paths already resolve the model.

Neither had any test coverage, which is why both shipped.

## 2. `grok-build` is metered, and the rule that blocked it is narrowed to its hazard

`grok-build` prints no usage — its stream is `text` events closing on a stop reason — and could not use the proxy either, because a spec declaring `config_files` was refused outright alongside `request_middleware`.

The refusal was wider than the hazard it guarded. A CLI reading its endpoint from an on-disk file never touches the env var again, so the middleware's bash-time rewrite cannot reach it — true. But the preamble boots the proxy from the declared upstream *one step before* that rewrite, which is the ordering `_middleware_preamble` already documents as deliberate, and which `test_if_the_validator_ever_loosens_the_preamble_order_is_middleware_first` was written in advance to protect.

So the config file can simply name the proxy. `base_url` now renders as the proxy's address whenever a spec declares middleware, and what stays refused is a template that hard-codes an endpoint of its own — the case that really would route around the proxy, silently.

## 3. `gemini-cli` runs on OpenRouter, and is metered

`gemini-cli` speaks Google's native `generateContent` and has no OpenAI-compatible auth mode — 0.55.1 knows `LOGIN_WITH_GOOGLE`, `USE_GEMINI`, `USE_VERTEX_AI`, `COMPUTE_ADC`, `GATEWAY` and `LEGACY_CLOUD_SHELL`, verified against the bundle — so OpenRouter cannot serve it directly, and the shipped LiteLLM overlay targets a *passthrough* that forwards to Google with the gateway's own Google credential.

LiteLLM's translation endpoint answers the same request from whatever backend its `model_list` maps the model to. A LiteLLM the operator runs therefore turns an OpenRouter key into a Gemini-speaking endpoint, which is what makes this harness runnable on an OpenRouter-only host at all. Ships as a sibling overlay, `gemini_openrouter_overlay.yaml`; the passthrough one is untouched.

One trap is documented with it: the CLI silently falls back to a flash model name of its own, and a gateway registering only the configured slug answers `Invalid model name passed in model=gemini-3.5-flash` — the trial then does no work and still returns a score.

**The proxy now reads Google's `usageMetadata` alongside OpenAI's `usage`**, normalised onto one basis: `promptTokenCount` already includes the cached part, as OpenAI's `prompt_tokens` does, while `candidatesTokenCount` **excludes** thinking tokens where `completion_tokens` includes them — so thinking is added back, or a consumer summing completions silently under-counts the part of the answer the model charged for and did not show.

### And the defect that surfaced

A LiteLLM gateway translating `generateContent` reports real counts on the unary path and **zeros on the streamed one**, which is the path the CLI takes. The tap faithfully recorded zeros and the trial reported `cost_usd: 0.0` for a task's worth of work — a run that did the job, recorded as free. Records that sum to nothing are now treated as **no measurement** rather than as no spend, which is what `_apply_harness_wire_usage`'s own docstring already said it wanted.

## Verification

Live trials on real Arena tasks, not fixtures:

| harness | task | score | cost | source |
|---|---|---|---|---|
| `opencode` | `git-recovery-challenge` | 1.00 | `$0.884` → flagged `cost_cache_rate_fallback: true` | stdout (`opencode/json`) |
| `grok-build` | `git-recovery-challenge` | 1.00 | `$0.076572`, 72,489 prompt / 54,144 cache-read | `middleware_proxy` |
| `gemini-cli` | `git-recovery-challenge` | **1.00** | unmeasured, correctly — gateway reports zeros on the streamed path | `middleware_proxy` |

`gemini-cli`'s 1.00 is against the **0.58** the same task returns when the agent never starts — which is what it scored before this change, on three tasks, while making zero model calls.

The proxy's container integration test now runs **both wire shapes** and asserts identical counts from each: a record's meaning must not depend on which CLI produced it. `2 passed`.

`10,870 passed, 22 skipped` across `tests/unit`, `tests/canonical` and the harness package, against current `main` — **no failures, and none to explain**: the five that were red when #1593 landed have since been fixed on `main`.

### One test was wrong before it was right

The regression test for the all-zero guard initially passed *identically with the guard deleted* — a defaulted `usage_log_container_path` meant it never reached the code it named. It was corrected, and then verified by disabling the guard and confirming the failure. Worth stating because the defects in this PR are the kind a test claims to cover and does not.


CiroGamboa:
Review fixes pushed in `d6899aa7`. All seven findings addressed; the first was real and the worst of them.

**1 — `step_finish` without a cost crashed the parser.** Reproduced: `_as_float` returns `None` where `_as_int` returns `0`, so `cost += None` raised `TypeError`. `parse_harness_stdout` runs unguarded on the raw stream after the CLI has already done and paid for its work, so this turned an unmetered trial into a **lost** one. Fixed, with a test.

**2 — cost defaulted to `0.0` and was always reported.** A stream whose steps carry no cost produced `cost_usd: 0.0` — the "$0.00 on a trial that ran" reading that commit 3 of this PR exists to remove, reintroduced on the stdout path. Cost now sums only the steps that reported one and stays `None` when none did.

**3 — the cache-fold assumed an Anthropic-shaped `input`.** Correct, and the sharpest finding: the basis follows the *provider*, not the CLI, and the docs support routing opencode at non-Anthropic vendors. Rather than assume, the parser now reads the basis off the step's own `total` — if `input + output == total` the input is already inclusive and the cache counters are not folded. Tested both shapes.

**4 — the config-template rule was whole-template.** Right on both halves: a template asking for `{{ base_url }}` on one line could hard-code an endpoint on another and pass (a live shape for grok-build's four model blocks), and `{{base_url}}` / `{{ base_url | trim }}` were refused despite rendering the same value. Now per line, skipping comments, with the spellings accepted.

**5 — Gemini thinking tokens dropped without a candidates count.** Fixed; also fixed the related guard, which treated a cached-only block as no usage.

**6 — a `generateContent` response as a JSON array read as nothing.** That is what `streamGenerateContent` answers without `alt=sse`. Handled.

**7 — the all-zero guard inspected two of four counters.** Now all four, matching what its own comment claimed.

**On the falsification point**, since it is the theme of this work: each of the five new guard tests was run with its fix disabled and observed to fail, then restored and observed to pass. Output on request.

`10,882 passed, 22 skipped` against current `main`; proxy container test `2 passed` across both wire shapes. `ruff` clean.

CiroGamboa:
Added the two provider gates in `fefacf0b`, following the question of what happens when OpenRouter or LiteLLM is down.

**How it works today, for the record:** `FallbackLLMClient` gives an ordered model-fallback chain — but it wraps `LLMClient.generate`, and a harness trial calls that **never**. The CLI owns its own connection; the engine only reads the model's *name* off the config to price it. So for all seven harnesses there is no failover, no retry, and until now no detection. The gemini-cli run proved it: every call 403'd and three Arena tasks still scored 0.42–0.58.

**Deliberately not auto-switching providers.** A trial that starts on OpenRouter and finishes on LiteLLM produces a cost and a score belonging to neither, which for a run whose entire output is a comparison is worse than failing. So: decide before, detect during.

**Before the run** — the orchestrator probes the endpoint the harness will actually use, taken from the adapter's *resolved* provider envelope with the credential attached, and refuses. Verified against both real endpoints: OpenRouter `REACHABLE`, the team LiteLLM gateway `REFUSED: HTTP 403`. A 404 passes (the probe asks whether an authorised caller reaches the host, not whether one route exists); 401/403/5xx and transport failures do not. `TOLOKAFORGE_SKIP_PROVIDER_PREFLIGHT=1` opts out for air-gapped runs, and an endpoint the probe cannot judge is left alone — silence means *not checked*, never *checked and fine*.

**During the run** — the proxy now writes one record per provider response instead of only per usage-bearing one. A refusal carries no counts, so token sums are unchanged, but a trial whose every request was refused is finally legible. Such a trial finalises `ERROR`/`API_ERROR`, which routes it to a synthesized grade rather than the task's reward file. One served request leaves it alone (retries and rate limits are not a dead run), and no records at all condemns nothing.

Two existing proxy tests asserted the old "only record usage-bearing responses" contract and were updated rather than worked around; the container test now asserts one record per response with exactly one carrying counts.

`10,898 passed, 22 skipped`; proxy container test `2 passed` on both wire shapes. Each new guard test was run with its guard disabled and observed to fail.

CiroGamboa:
Superseded by #1618, which contains these commits unchanged plus the pricing refresh and the measurement guards. Consolidated so the whole surface reviews as one.

PR #1617, opened 2026-09-18
chore(models): refresh the pricing table from OpenRouter

CiroGamboa:
The table declares its own source and its own age — `"source_url": "https://openrouter.ai/api/v1/models"`, `"updated_at": "2026-09-02"` — and nothing in the engine reads either. `pricing-updater` is in no CI workflow, so refreshing is entirely manual, and it had drifted 16 days.

## What that cost

An Arena sweep across the seven shipped harnesses priced four of them. Two were billed at rates OpenRouter no longer charges:

| model | table said | OpenRouter charges | effect |
|---|---|---|---|
| `openai/gpt-5.6-sol` | $5.00 / M input | **$2.00** | codex overstated **2.5×** |
| `moonshotai/kimi-k3` | $3.00 / M input | **$2.10** | kimi-code overstated **1.4×** |
| `anthropic/claude-sonnet-4.6` | $3.00 | $3.00 | correct |
| `qwen/qwen3-coder-plus` | $0.65 | $0.65 | correct |

The two that were right are why the arithmetic was never in doubt: `claude-code`'s computed cost reproduced the CLI's own reported bill to fifteen significant figures, four times. The method was sound throughout; the inputs had aged.

It also changed the answer. Ranked by cost, the sweep read `qwen < kimi < claude < codex`. It is actually `kimi < qwen < codex < claude` — two inversions, and `codex` moves from most expensive to third.

Also refreshed: `x-ai/grok-4.5` cache-read $0.50 → $0.30, and `google/gemini-3.6-flash` halved on both rates.

## What this is, and is not

Merge mode, so the hand-curated cache rates that `pricing-updater`'s field-level merge exists to preserve survive the refresh. 423 → 565 models; the growth is new entries, not rewrites.

**This fixes today and nothing more.** Nothing stops the same drift recurring next month — the engine still cannot tell a fresh table from a stale one. That gate is a separate change and is not in this PR.

## Verification

Re-pricing the sweep's **recorded token counts** against the refreshed table reproduces the corrected figures exactly, without re-running a trial:

```
harness   as reported  re-priced now
claude         2.0548         2.0548
codex          3.7037         1.4594
kimi           1.3380         0.9804
qwen           1.0308         1.0308
```

`2,216 passed` across `tests/canonical` — including the models-fingerprint snapshot, which the pricing table feeds. `57 passed` across the pricing unit tests, `36 passed` in `tools/pricing-updater`.

PR #1618, opened 2026-09-18
feat(coding-harness): meter every shipped harness, refuse a dead provider, and stop a wrong price passing as a number

CiroGamboa:
Supersedes #1616 and #1617, whose commits are included here.

Three strands, one surface. Everything below was found by running the seven shipped coding harnesses against real Arena tasks and reading what came back — never by the test suite, which was green throughout.

---

## Part 1 — Three harnesses could not report what they spent

**`opencode` reported nothing.** Its stream carries a `step_finish` per step with that step's `tokens` and `cost` — 22 of them on the trial that surfaced it — and no parser had been written. Whether `tokens.input` already includes the cached prompt depends on the **provider** opencode routed to, not on opencode, and the docs support routing it at either; so the basis is read off the step's own `total` rather than assumed, and both shapes are tested.

**`grok-build` could not be metered, and the rule stopping it was wider than its hazard.** A spec declaring `config_files` was refused outright alongside `request_middleware`, because a CLI reading its endpoint from a file never sees the middleware's bash-time rewrite. True — but the preamble boots the proxy from the declared upstream *one step before* that rewrite, which `_middleware_preamble` already documents as deliberate and which `test_if_the_validator_ever_loosens_the_preamble_order_is_middleware_first` was written in advance to protect. So the config file can simply name the proxy. What stays refused is any *line* hard-coding an endpoint of its own — per line, because `grok-build` declares one model block per model.

**`gemini-cli` could not run at all.** It speaks Google's native `generateContent` and has no OpenAI-compatible auth mode (0.55.1 knows `LOGIN_WITH_GOOGLE`, `USE_GEMINI`, `USE_VERTEX_AI`, `COMPUTE_ADC`, `GATEWAY`, `LEGACY_CLOUD_SHELL` — verified against the bundle), so OpenRouter cannot serve it and the shipped overlay targets a passthrough needing a Google credential. LiteLLM's *translation* endpoint answers the same request from whatever backend its `model_list` maps to, so a LiteLLM the operator runs turns an OpenRouter key into a Gemini-speaking endpoint. Ships as a sibling overlay; the passthrough one is untouched. The proxy now reads `usageMetadata` alongside `usage`, normalised onto one basis — `candidatesTokenCount` **excludes** thinking tokens where `completion_tokens` includes them, so thinking is added back.

## Part 2 — A dead provider looked like a weak agent

A coding-harness CLI owns its own connection. `FallbackLLMClient` wraps `LLMClient.generate`, and **a harness trial calls it never** — so no retry and no model fallback is in the path. When the provider refused everything, the CLI still wrote a transcript and exited, the tool call returned success, and the trial was scored against an untouched repository: **three Arena tasks scored 0.42–0.58 while the gateway 403'd every call.**

Two gates, because the two failures differ. **Before the run**, the endpoint the harness will actually use is probed — the adapter's resolved envelope, credential attached, since an endpoint answering only authorised callers is the case worth catching — and the run is refused. Verified against both real endpoints: OpenRouter `REACHABLE`, the team gateway `REFUSED: HTTP 403`. A 404 passes; the question is whether an authorised caller reaches the host, not whether one route exists.

**During the run**, the proxy records one line per provider response instead of only per usage-bearing one. A refusal carries no counts, so token sums are unchanged, but a trial where nothing succeeded is finally legible, and is finalised `ERROR`/`API_ERROR` — a synthesized grade rather than the task's reward file. One served request leaves it alone; no records at all condemns nothing.

**Deliberately not auto-switching providers.** A trial starting on OpenRouter and finishing on LiteLLM yields a cost and a score belonging to neither, which for a run whose whole output is a comparison is worse than failing.

## Part 3 — The numbers themselves were wrong, and nothing could have said so

The shipped pricing table is a copy. It names its source and the moment it was fetched, refreshing it is a manual command in **no CI workflow**, and nothing read either field. A copy 16 days behind billed `openai/gpt-5.6-sol` at `$5.00`/M input against an actual `$2.00`, and `moonshotai/kimi-k3` at `$3.00` against `$2.10` — overstating a harness sweep 2.5× and 1.4×, and **inverting its cost ranking**.

Every existing pricing signal passed while that was happening, because each asks about a *trial* and staleness is a question about the *table*:

| signal | asks | why it passed |
|---|---|---|
| cache-rate preflight | is this row complete? | it was |
| `cost_cache_rate_fallback` | were cache tokens billed at the input rate? | they weren't |
| vendor cross-check | does our arithmetic match the CLI's figure? | it did, for the correctly-priced models |

So the table is refreshed **and** four guards added:

1. **A run refuses prices it cannot vouch for** — the rates it will bill are compared against the table's own `source_url`, and a disagreement beyond 1% refuses, naming each rate and ratio. Age alone warns. Failing to *ask* is never an answer: an unreachable source leaves the run alone. `TOLOKAFORGE_SKIP_PRICING_FRESHNESS=1` opts out.
2. **Every trial records the rates it was priced at** and the key that decided the lookup — normalisation strips `openrouter/` and infers a namespace, so the row billed is routinely not the one the config names. Correcting the sweep meant rebuilding rates by hand; now a corrected table re-prices any recorded run from its own bundle.
3. **The vendor cross-check is checked.** `harness_reported_cost_usd` was documented as "the cross-check" from the day it was added and compared to ours nowhere.
4. **The nightly harness matrix can fail.** It ran all seven harnesses against real packs and **asserted nothing** — a trial reporting no cost, or scoring against a repository its agent never touched, left the job green. Every defect here would have passed it.

Also fixed: `cost_cache_rate_fallback` was set only on the engine's request path, never on either harness path, so the one field saying "this cost is overstated" stayed `False` on a trial overstating by 4.6×. And an all-zero wire measurement is now treated as **no measurement** rather than as no spend, after a LiteLLM gateway reported real counts on the unary path and zeros on the streamed one — the path `gemini-cli` takes.

---

## Verification

Live trials on real Arena tasks throughout, because every defect here was invisible to the suite:

| harness | task | score | cost |
|---|---|---|---|
| `opencode` | `git-recovery-challenge` | 1.00 | `$0.884`, flagged `cost_cache_rate_fallback: true` |
| `grok-build` | `git-recovery-challenge` | 1.00 | `$0.076572` — was null |
| `gemini-cli` | `git-recovery-challenge` | **1.00** | unmeasured, correctly — gateway zeroes the streamed path |

`gemini-cli`'s 1.00 is against the **0.58** the same task returns when the agent never starts, which is what it scored before, on three tasks, while making zero model calls.

**The freshness guard, both directions:** on the stale table it refuses, naming all three drifted rates; with the refresh in this PR it logs `Pricing table agrees with its source`.

**The CI gate, against the real recorded bundles:** healthy `claude-code` run `0 problems`; dead-agent `gemini-cli` run 3 problems; unmetered `grok-build` run 3 problems; overstated `opencode` run 1 problem.

**Re-pricing the sweep from recorded tokens** reproduces the corrected figures without re-running: `codex` `$3.7037 → $1.4594`, `kimi-code` `$1.3380 → $0.9804`, `claude-code` and `qwen-code` unchanged.

`10,917 passed, 22 skipped` across `tests/unit`, `tests/canonical` and the harness package, against current `main`. Proxy container test `2 passed` on both wire shapes. `ruff` clean.

**Every guard was run with its guard disabled and observed to fail** — the discipline this stream exists to establish, adopted after a regression test of mine passed identically with its fix deleted, and after a review found my `opencode` parser crashed on a step with no cost, losing the trial after the CLI had already been paid for.

## Still open, deliberately

- **`avg_cost_usd` is diluted** — a filtered numerator over an unfiltered denominator, the arithmetic already fixed for scores and not for cost. Aggregation, not measurement.
- **`grok-build`'s cost remains overstated** by a missing cache-write rate the source does not publish; the per-trial flag now says so.


CiroGamboa:
Review fixes pushed in `a5019ffd` — all fifteen findings addressed. **Three would have broken CI outright**, which is a good argument for the review having run at all.

**The three that would have broken things**

1. **The CI gate would have gone permanently red.** I made `cost_cache_rate_fallback` a failure; four of the seven matrix models resolve to rows the source publishes no cache-write rate for (`anthropic/claude-sonnet-4-6` carries neither), so the flag is their *normal state*. A gate that always fails teaches everyone to ignore it. Annotated now. Genuinely unmeasured trials still fail, with `gemini-cli` in a named exemption set carrying its reason inline — the `UNSHIELDED_HARNESSES` pattern: a documented gap does not fail, a silent regression still does.
2. **The provider preflight probed the container's URL from the host.** `host.docker.internal` does not resolve on a Linux host at all, so a perfectly healthy container-network gateway would have hard-refused the run. Only an authoritative answer condemns now — 401/403/5xx — and a transport failure means *not checked*. That still catches the live failure, which was a 403.
3. **The preflight never ran on the shipped configuration.** It read `models.agent.harness`, while this repo's own matrix workflow and the terminal-bench recipes select the harness through `harness_adapter.params.agent_harness`, and the lift runs the other way. My tests only exercised the two helpers, never the orchestrator method — so they could not have caught it.

**Two paths still let a dead provider through**, which is the failure these guards exist for: a served `/models` GET (two harnesses allowlist one) masked a trial whose every completion was refused, and a `URLError` returned *before* the tap ran, so a connection-refused provider left no record at all.

**One was the same defect I keep fixing, reintroduced by me.** `_parse_opencode_json` counted a step before checking for its tokens block, so a stream whose steps carried none reported zeros — priced at $0.00, suppressing the wire fallback that might still have measured it. "Not measured" as "measured as zero", on the one path that had no guard.

**Also:** the freshness check fetched the live model list on *every* `load_tasks()`, making the unit suite non-hermetic and four tests 4× slower — it now reaches the network only when the table is actually old, and the suite sets the opt-out so hermeticity stops depending on when someone last refreshed a file. `cost_cache_rate_fallback` fired on a missing rate regardless of whether that token class was observed. `pricing_basis`'s insertion orphaned `harness_reported_cost_usd`'s docstring onto the wrong field. My opencode rationale was inverted — folding an inclusive `input` **overstates**, it does not understate. New fields, both opt-out variables and the inert `path_filter` comment are now documented.

`10,925 passed, 22 skipped`; proxy container test `2 passed` on both wire shapes. The CI gate re-verified against the real recorded bundles: healthy `claude-code` passes clean, `gemini-cli` passes with three notes under its exemption, `opencode` passes with an overstated-cost note, and a genuinely unmetered bundle still fails.

PR #1622, opened 2026-09-21
fix(observability): enforce single-attempt OTLP exports and reject redirects

balazs-toloka:
The v4 exporter must make one OTLP POST attempt per batch. Bypassing the OpenTelemetry SDK's outer retry loop was insufficient: its `_export` method re-posts after a lost connection, HTTP redirects can resend the body, and a session adapter can retry independently.

The engine-free `tolokaforge_langfuse.otlp_transport` module now sends the request directly using the SDK's endpoint, session, timeout, compression and certificates, with redirects disabled and adapter retries set to zero. Missing SDK internals fail at startup. Only 2xx responses count as successful exports: 301/302/303/307/308 responses fail without following the redirect, so undelivered spans are reported in the receipt. The v3 family retains the stock exporter.

The tests use real `requests.Response` objects. This matters because `Response.ok` is true for redirects; a fabricated `ok=False` response previously concealed the false-success bug.

ADR-0048, the operator documentation, the wheel changelog and related code descriptions now describe write-once as a producer policy. Measurements on Langfuse 4.38.0 in `events_only` show that re-sending the same observation id converges to one observation, last write wins, including when content or environment changes. Single-attempt delivery avoids unnecessary requests and unintended overwrites; it does not guarantee delivery or prevent permanent duplicate rows. The existing preview, root and score layout remains in place.

Validation:
- 287 Langfuse and observability tests passed.
- All five redirect regression cases failed before the 2xx-only check and pass after it.
- A local HTTP server confirmed one request per batch for success, 307, 308, 503 and a lost response; redirects now return FAILURE. Gzip and deflate payloads decode correctly.
- Ruff lint/format, Black, whitespace checks and all 11 import-boundary contracts passed.

The earlier `test-smoke` failure in `test_manifest_covers_every_declared_config_field` (`StateChecksConfig.auto_normalize_nullables`) also reproduces on base commit `3c2023b5`; it is unrelated to this change. The earlier `hygiene-review` job failed before producing a review.

The offline connector still needs to adopt the shared transport when its plugin-wheel pin advances.


CiroGamboa (changes requested):
## Review — post an OTLP batch exactly once (write-once receiver)

The transport code is correct, well-tested, and improves fail-fast + secret-log hygiene. But the diff ships — and in two places *strengthens* — a safety premise this PR's own body has verified false, across three source-of-truth docs. Doc-only changes; the code needs none.

### Blocker — three docs assert a rationale the PR body has disproven

The body's 2026-09-21 correction states it plainly: on Langfuse 4.38.0 `events_only` a re-sent observation id is an **update (last-write-wins), not a second undeletable row** — "this PR does not prevent an undeletable duplicate, because there is none to prevent." Yet:

- `docs/adr/0048-...:110-114` — "a duplicate that cannot be deleted … The single post is the safety argument itself."
- `docs/OBSERVABILITY.md:237-239` — "with no way to delete either copy … This is the whole safety argument for writing to such a receiver."
- `tolokaforge_langfuse/CHANGELOG.md:37-38` — this diff changed "posts each batch once" → "**exactly once**, because a batch the receiver already wrote … would be a duplicate it cannot delete."

Merging ships documentation the author knows to be wrong into the ADR, the operator-facing doc, and the published wheel CHANGELOG (Core Rule 8). A future reader takes the "cannot be deleted" argument at face value.

**Fix:** keep the *true* justification the body already gives — one POST is cheaper, and a silent last-write-wins overwrite is not something a producer should do by accident — and rewrite around that. The "three repeats" technical content (retry loop / `_export` re-post / redirect + adapter) is accurate and worth keeping; only the *why* changes. Suggested:

> One POST per batch. A re-sent observation id on this family is an update (last-write-wins), not a new row, so a repeat is not a lost record — but it silently overwrites what the first post wrote, which a producer should not do by accident, and the extra request buys nothing. The v4 family therefore posts each batch once, which takes more than turning the exporter's retry loop off: the SDK's `_export` re-posts on a lost connection, `requests` follows a 307/308 by re-sending the body, and a session's adapter can retry on its own. The exporter makes the request itself with redirects refused and no adapter attempts; an SDK whose exporter cannot be asked for this fails the run at start rather than falling back to a retrying one.

If the premise correction must be a separate change by the engine owner, then at minimum drop the strengthening edits here (leave the prior wording rather than making the docs *more* wrong) and file a tracked follow-up for the ADR premise + F9.

### Major — past-tense implementation archaeology in the ADR

`docs/adr/0048-...:118` — "This was found by review on 2026-09-21, after the first implementation had skipped only the outer loop." That narrates a prior code defect, not the decision (Core Rule 8 / §5). **Fix:** delete the sentence; keep the current-state description ("the exporter issues the `session.post` itself; its test counts posts at the HTTP layer").

### Minor — internal reference in the PR body

The body cites `docs/TECHDEL-497-langfuse-integration/…` (evidence path + `BACKFILL_THROUGHPUT_PLAN.md`). Internal path in a public PR body — worth scrubbing (not in the diff, non-gating).

### Clean / improved (verified)

The code is sound irrespective of the premise: `SingleAttemptSpanExporter` owns `session.post` (kills `_export`'s `ConnectionError` re-post), `allow_redirects=False`, `HTTPAdapter(max_retries=0)`; fails **closed** via `SingleAttemptUnavailable` + the `REQUIRED_EXPORTER_ATTRIBUTES` check rather than degrading to the retrying exporter; the failure log is narrowed to `type(exc).__name__`; tests moved from stubbing `_export` to counting posts at the HTTP layer (`test_a_lost_connection_is_not_a_second_post`, `test_a_redirect_is_not_followed`); the engine-free invariant is preserved and enforced (`test_engine_free_modules.py:22` + a build-and-use probe). One POST per batch is the right behaviour — for the true reason, not the disproven one.

**Net:** the code can merge as-is; the docs must not ship a knowingly-false safety argument. The premise correction is explicitly deferred to the engine owner in the body — please resolve it here or as a tracked follow-up before this lands.

